PR #15, opened 2024-05-28
Debian build steps

snadrus:
It depends on a base64-encoded key shared privately to sign the DEB. 
It will use the internal version. 
The output is 2 deb files that are arch-specific to the build environment, have the inbuilt version, are signed, and cover each of opencl and cuda. 

@magik6k I think this is a super-set of the changes you needed to make to get it building. Lets use this instead.


apt/make_debs.go
```diff
@@ -70,14 +111,20 @@ func part2(base, product, extra string) {
 		OrPanic(os.MkdirAll(base, 0755))
 		OrPanic(copyFile("curio", path.Join(base, "curio")))
 		OrPanic(copyFile("sptool", path.Join(base, "sptool")))
+		base = path.Join(dir, "etc", "systemd", "system")
```
LexLuthr, line 114: This part is used for user defined custom services. We should use the system default /lib/systemd/system/

build/version.go
```diff
@@ -25,7 +39,12 @@ func BuildTypeString() string {
 	}
 }
 
-const BuildVersion = "1.22.0"
+// Intent: Major.Network.EvenReleases
+var BuildVersionArray = [3]int{1, 22, 0}
```
LexLuthr, line 43: What's the advantage of this array?

LexLuthr (commented):
Looks good with a suggestion and a question.

build/version.go
snadrus: elsewhere (make_debs.go) interpret the version. Having a standard of it being numbers makes that interpretation simpler.

LexLuthr (approved):

PR #20, opened 2024-05-29
UI: Display active task history, retry sector button

magik6k:
This PR adds a table showing recent task history on the sector page (for tasks in the pipeline table, so really useful for debugging what caused the sector to halt progress) + Adds a "Resume" button on sectors which have tasks that have been retried too many times

The retry button removes failed taskIDs from the pipeline table so that the poller can start a new task in its place.

Tested on my setup, works as intended and looks coherent with the rest of the UI; Unfortunately I retried my failed sector before making this PR, so not much to share in the form of a screenshot.

### DIFF
--- a/harmony/harmonydb/sql/20231217-sdr-pipeline.sql
+++ b/harmony/harmonydb/sql/20231217-sdr-pipeline.sql
@@ -136,3 +136,7 @@ create table sectors_allocated_numbers (
     sp_id bigint not null primary key,
     allocated jsonb not null
 );
+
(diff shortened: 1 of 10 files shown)


magik6k:
Also added total 24h task counts to the main page
2024-05-30-095140_1371x988_scrot

magik6k:
The 24h counts table will have clickable task names to drill down into history and be able to see what has failed

LexLuthr (approved):

PR #31, opened 2024-06-03
Fix wdpost skip / race issues

magik6k:
This PR:
* Fixes a race in WdPoSt generation code
* Fixes skipped sector handling in post verification
* Corrects handling of failing-to-verify WdPosts

### DIFF
--- a/cmd/curio/proving.go
+++ b/cmd/curio/proving.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"github.com/filecoin-project/curio/cmd/curio/tasks"
 	"github.com/filecoin-project/curio/deps"
+	"github.com/yugabyte/pgx/v5"
 	"os"
 	"time"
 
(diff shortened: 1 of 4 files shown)


cmd/curio/proving.go
```diff
@@ -121,6 +126,40 @@ var wdPostTaskCmd = &cli.Command{
 				break
 			}
 			fmt.Print(".")
```
snadrus, line 128: Log?

tasks/window/compute_do.go
```diff
@@ -190,32 +191,49 @@ func (t *WdPostTask) DoPartition(ctx context.Context, ts *types.TipSet, maddr ad
 				return nil, xerrors.Errorf("post generation randomness was different from random beacon")
 			}
 
+			for _, skippedNum := range computeSkipped {
```
snadrus, line 194: an array that we just keep overwriting this one value from? Should this be a sum?
More comments please.

snadrus (approved):

cmd/curio/proving.go
magik6k: This isn't a Println, just a print appending to the line

PR #32, opened 2024-06-04
Paramfetch with aria2 support

magik6k:
This is a copy of https://github.com/filecoin-project/go-paramfetch/blob/master/paramfetch.go with:
* Heavily opinionated support for aria2
* Parallel fetch support

In the future we can think of adding support for p2p fetches from the cluster, but for now this makes my paramfetch operations 7-10x faster

### DIFF
--- a/lib/fastparamfetch/paramfetch.go
+++ b/lib/fastparamfetch/paramfetch.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"io"
 	"net/http"
 	"net/url"
@@ -45,6 +46,11 @@ type fetch struct {
 	fetchLk sync.Mutex
 
 	errs []error
+
+	fsLockRelease func()
+	fsLockOnce    sync.Once
+	lockFail      bool // true if we failed to acquire the lock at least once, meaning that is was claimed by another process
+	lockErr       error
 }
 
 func getParamDir() string {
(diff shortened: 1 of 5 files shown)


LexLuthr (commented):
We should update our dependencies to include aria2c here on Linux and MacOS systems.

magik6k:
Yes, but not that it is technically optional in this PR, so nothing breaks if you don't have it, just the initial paramfetch is not as fast

snadrus:
Please add:
  , aria2
to 
   File: apt/DEBIAN/control 
   Line: starts with "Depends:"

I've verified that name in the past 4 Ubuntu versions in the Universe repo.

LexLuthr (approved):
Tested. Looks good.

lib/fastparamfetch/paramfetch.go
```diff
@@ -0,0 +1,319 @@
...
+	return err
+}
+
+func fetchWithAria2c(ctx context.Context, out, url string) error {
+	aria2cPath, err := exec.LookPath("aria2c")
+	if err != nil {
+		return xerrors.New("aria2c not found in PATH")
+	}
+
+	cmd := exec.CommandContext(ctx, aria2cPath, "--continue", "-x5", "-o", out, url)
```
LexLuthr, line 310:
```suggestion
	cmd := exec.CommandContext(ctx, aria2cPath, "--continue", "-x16", "-s16", "-o", out, url)
```

LexLuthr (approved):
Just 1 one small change to speed up download

PR #33, opened 2024-06-04
fix: happy path itests

LexLuthr:


itests/curio_test.go
```diff
@@ -1,355 +1,425 @@
...
+		Hosts:    []string{envElse("CURIO_HARMONYDB_HOSTS", "127.0.0.1")},
+		Database: "yugabyte",
+		Username: "yugabyte",
+		Password: "yugabyte",
+		Port:     "5433",
+	}
+	db, err := harmonydb.NewFromConfigWithITestID(dbConfig, sharedITestID)
+	require.NoError(t, err)
+
+	defer db.ITestDeleteAll()
```
magik6k, line 73: Would be better to pass `testing.T` (or `testing.T.Cleanup` / `func(onTestEnd func())`) into the test db constructor and register a t.Cleanup hook there.

magik6k (approved):
One tiny nit, but other than that this is great.

magik6k (approved):

PR #39, opened 2024-06-06
fix: update dependencies

LexLuthr:
Update FFI version to NV22
Update Lotus to v1.27.0
Bring in "filler" lib to Curio as Lotus v1.27.0 doesn't have it.
Bring in "storage/paths"
Create local "build" package
Fix devnet images


harmony/harmonydb/metrics.go
```diff
@@ -0,0 +1,77 @@
...
+			Aggregation: view.LastValue(),
+			TagKeys:     []tag.Key{dbTag},
+		},
+		&view.View{
+			Measure:     DBMeasures.Errors,
+			Aggregation: view.Sum(),
+			TagKeys:     []tag.Key{dbTag},
+		},
+	)
+	err := prometheus.Register(DBMeasures.Waits)
```
magik6k, line 68: You can't have duplicate measures registered in prometheus, and those will conflict with the lotus ones, which is why I've changed curio harmonydb to point at the one in lotus

Really I still believe that we may want to put harmonydb + dbindex impl into a separate repo so that there's only one harmonydb.

Having two will keep being a point of annoyance.

LexLuthr: Undoing the metrics. We should create our own metric library after GA with new measure and more suitable for Curio metrics.

PR #42, opened 2024-06-07
fix: Resolve shadowed error in SaveConfigToLayerMigrateSectors

strahe:
This PR fixes a shadowed error in the `SaveConfigToLayerMigrateSectors` function.


magik6k:
I'd probably assign to a variable with a different name for clarity

strahe:
```
func someFunction() (err error) {
    defer func() {
       //  err = nil  // shadowed
        err := gotErr() 
        err1 := gotErr()
    }()
    // ... some code that might set err ...
    return
}
```
Using different variable names indeed makes it clearer, but the effect is the same.

LexLuthr:
Fixed in https://github.com/filecoin-project/curio/pull/39

PR #48, opened 2024-06-10
Fix WinningPoSt with remote sectors

magik6k:
Another regression from ffiselect, apparently made me loose 32 FIL on mainnet.

Not having this resulted in errors like
```
error: failed to compute winning post proof: get winning sector:6822,vanila failed: fsstat: path not found
```


snadrus (approved):

snadrus:
I think this fixes where people said stuff had to be local.

PR #53, opened 2024-06-12
fix: filler calculation

LexLuthr:


magik6k (approved):
Oh this makes sense. Also makes sense that in kind of passed tests where we probably used full-sized pieces

PR #58, opened 2024-06-13
feat: synthetic PoRep

LexLuthr:


LexLuthr:
@magik6k Should we provide Max task value in config for Synth tasks or an internal 50 would be good enough?

cmd/curio/tasks/tasks.go
```diff
@@ -186,8 +186,9 @@ func StartTasks(ctx context.Context, dependencies *deps.Deps) (*harmonytask.Task
 		if cfg.Subsystems.EnableSealSDRTrees {
 			treeDTask := seal.NewTreeDTask(sp, db, slr, cfg.Subsystems.SealSDRTreesMaxTasks)
 			treeRCTask := seal.NewTreeRCTask(sp, db, slr, cfg.Subsystems.SealSDRTreesMaxTasks)
+			treeSynth := seal.NewSyntheticProofTask(sp, db, slr)
```
magik6k, line 189: It's not a tree, more generating proofs
```suggestion
			genSynth := seal.NewSyntheticProofTask(sp, db, slr)
```

lib/ffi/sdr_funcs.go
```diff
@@ -663,3 +670,47 @@ func (sb *SealCalls) TreeD(ctx context.Context, sector storiface.SectorRef, unse
...
+
+	err = ffi.GenerateSynthProofs(sector.ProofType, sealed, unsealed, fspaths.Cache, fspaths.Sealed, sector.ID.Number, sector.ID.Miner, randomness, pieces)
+	if err != nil {
+		return xerrors.Errorf("generating synthetic proof: %w", err)
+	}
+
+	if err = ffi.ClearCache(uint64(ssize), fspaths.Cache); err != nil {
+		// Note: non-fatal error.
+		log.Warn("failed to GenerateSynthProofs(): ", err)
+		log.Warnf("num:%d tkt:%v, sealedCID:%v, unsealedCID:%v", sector.ID.Number, randomness, sealed, unsealed)
```
magik6k, line 694: Honestly this should be fatal. Removing stuff should never fail quietly, this can lead to us pulling in garbage into long-term storage.

All Seal calls should cope with edge-case inputs but at the same time should do their best to leave the data in clean state.

lib/ffi/sdr_funcs.go
```diff
@@ -663,3 +670,47 @@ func (sb *SealCalls) TreeD(ctx context.Context, sector storiface.SectorRef, unse
...
+	}
+
+	var sd [32]byte
+	_, _ = rand.Read(sd[:])
```
magik6k, line 698: You definitely want to generate this randomness inside the for loop, otherwise you're reading the same bits of data each time

lib/ffi/sdr_funcs.go
```diff
@@ -663,3 +670,47 @@ func (sb *SealCalls) TreeD(ctx context.Context, sector storiface.SectorRef, unse
...
+
+	// Generate 3 random commits
+	for i := 0; i < 3; i++ {
```
magik6k, line 701: I'd put this `3` in a const at the top of the file so that it's easier to tune

tasks/seal/task_synth_proofs.go
```diff
@@ -0,0 +1,239 @@
...
+			Number: abi.SectorNumber(sectorParams.SectorNumber),
+		},
+		ProofType: sectorParams.RegSealProof,
+	}
+
+	var pieces []struct {
+		PieceIndex  int64  `db:"piece_index"`
+		PieceCID    string `db:"piece_cid"`
+		PieceSize   int64  `db:"piece_size"`
+		DataRawSize *int64 `db:"data_raw_size"`
```
magik6k, line 99: You definitely don't want raw data size here, just piece size

magik6k (commented):
Few comments, looks great overall!

tasks/seal/task_synth_proofs.go
LexLuthr: We need that to calculate offset and padding for filler pieces.

LexLuthr: I am not sure sure if `abi.PaddedPieceSize(PieceSize).Unpadded == DataRawSize` is true.

magik6k: No, DataRawSize is the .car size, without the null-padding making it power-of-two-after-fr32-padding.

You just want to use PieceSize

LexLuthr: Resolved with new dealData function usage

tasks/seal/task_sdr.go
```diff
@@ -84,73 +80,11 @@ func (s *SDRTask) Do(taskID harmonytask.TaskID, stillOwned func() bool) (done bo
...
-		PieceCID    string `db:"piece_cid"`
-		PieceSize   int64  `db:"piece_size"`
-		DataRawSize *int64 `db:"data_raw_size"`
-	}
-
-	err = s.db.Select(ctx, &pieces, `
-		SELECT piece_index, piece_cid, piece_size, data_raw_size
-		FROM sectors_sdr_initial_pieces
-		WHERE sp_id = $1 AND sector_number = $2 ORDER BY piece_index ASC`, sectorParams.SpID, sectorParams.SectorNumber)
+	dealData, err := dealdata.DealDataSDRPoRep(ctx, s.db, s.sc, sectorParams.SpID, sectorParams.SectorNumber, sectorParams.RegSealProof)
```
magik6k, line 83: DealData will create readers for the pieces, we don't need that, just want the CommD.

Probably easiest addressed by adding a `onlyCommD` bool param to the `dealdata.DealDataSDRPoRep` func

lib/ffi/sdr_funcs.go
```diff
@@ -669,3 +678,41 @@ func (sb *SealCalls) TreeD(ctx context.Context, sector storiface.SectorRef, unse
...
+
+	// Generate 3 random commits
+	for i := 0; i < C1CheckNumber; i++ {
+		var sd [32]byte
+		_, _ = rand.Read(sd[:])
+		_, err = ffi.SealCommitPhase1(sector.ProofType, sealed, unsealed, fspaths.Cache, fspaths.Sealed, sector.ID.Number, sector.ID.Miner, randomness, sd[:], pieces)
+		if err != nil {
+			return xerrors.Errorf("checking PreCommit for synthetic proofs for num:%d tkt:%v seed:%v sealedCID:%v, unsealedCID:%v failed: %w", sector.ID.Number, randomness, sd[:], sealed, unsealed, err)
+		}
+	}
```
magik6k, line 711: Would be nice to independently also put those check-rounds in TreeRC

lib/ffi/sdr_funcs.go
```diff
@@ -669,3 +678,41 @@ func (sb *SealCalls) TreeD(ctx context.Context, sector storiface.SectorRef, unse
...
+
+	err = ffi.GenerateSynthProofs(sector.ProofType, sealed, unsealed, fspaths.Cache, fspaths.Sealed, sector.ID.Number, sector.ID.Miner, randomness, pieces)
+	if err != nil {
+		return xerrors.Errorf("generating synthetic proof: %w", err)
+	}
+
+	if err = ffi.ClearCache(uint64(ssize), fspaths.Cache); err != nil {
+		return xerrors.Errorf("failed to clear cache for synthetic proof of sector %d of miner %d", sector.ID.Miner, sector.ID.Number)
+	}
```
magik6k, line 701: We want to call ClearCache after successful C1 rounds, having layers means that we can first try recomputing Trees

magik6k (commented):
CI seems sad + a few comments, but this looks really good already

tasks/seal/task_sdr.go
LexLuthr: Adding this. We it would also reduce the footPrint if the function for sync task. It doesn't require data readers either.

lib/ffi/sdr_funcs.go
LexLuthr: This would increase the TreeRC time but few seconds if we do this for all sectors. Is that acceptable?

LexLuthr: Question, does having cache files around would force C1 to use cache files instead of synthetics ones?

magik6k: > This would increase the TreeRC time but few seconds if we do this for all sectors. Is that acceptable?

Yes, that's fine

LexLuthr: So, this doesn't work for synthetic proof sectors. Synthetic cache files are expected by C1 in this case.

LexLuthr: I have pushed a new commit with required changes.

tasks/seal/task_synth_proofs.go
```diff
@@ -0,0 +1,203 @@
...
+func (s *SyntheticProofTask) TypeDetails() harmonytask.TaskTypeDetails {
+	ssize := abi.SectorSize(32 << 30)
+	ram := uint64(8 << 30)
+	if IsDevnet {
+		ssize = abi.SectorSize(2 << 20)
+		ram = uint64(1 << 30)
+	}
+
+	res := harmonytask.TaskTypeDetails{
+		Max:  50,
```
magik6k, line 157: Would add this max to the config, also default lower, probably 16

LexLuthr: Added to config with new default value as other don't have one either.

magik6k (approved):

PR #59, opened 2024-06-15
forest Test

snadrus:
Create a CircleCI test to alert whenever Curio begins using a chain API that is not one we have informed the Forest team about. 

In creating this, 3 new API usages were discovered and shared via email. 
SPTool & Boost are not included here.
"enforcement" applies to the use of the API from /deps/. 

Direct uses (like SPTool & Boost do) are not tracked. Please use /deps/deps.go's chain API. It offers the entire lotus API, but errors in testing. 

Got a testing error? Just add the API to /api/api_chain.go's ChainSubsetForForest andn share it with the Forest team.


LexLuthr (commented):
I think we are using v1API in couple of more places in web. One would be for chain sync status and there is one more in simpleinfo under hapi.

LexLuthr:
I am not sure if this is the final or first draft as the approach here ties us to a build. We probably don't want that. User's shouldn't need to build specific to nodes. Tests should also be independent in similar manner.

snadrus:
> I am not sure if this is the final or first draft as the approach here ties us to a build. We probably don't want that. User's shouldn't need to build specific to nodes. Tests should also be independent in similar manner.

Actually, we can target the same full API for either Lotus or Forest as long as we know that the subset we use is Forest-safe. This code adds a unit test to ensure such a thing is true.

api/proxy_gen.go
magik6k, line 1: I don't see the generator for this in the Makefile, probably should add.

api/api_chain_all.go
magik6k, line 1: Honestly I'd remove this and use `ChainSubsetForForest` everywhere

api/api_chain.go
```diff
@@ -0,0 +1,166 @@
...
+	apitypes "github.com/filecoin-project/lotus/api/types"
+	"github.com/filecoin-project/lotus/chain/actors/builtin/miner"
+	"github.com/filecoin-project/lotus/chain/types"
+	"github.com/filecoin-project/lotus/journal/alerting"
+)
+
+// ChainChainSubsetForForest is a subset of the Filecoin API that is supported by Forest.
+type ChainSubsetForForest interface {
+	Common
+	Net
```
magik6k, line 35: Do we use anything from Net?

magik6k, line 35: I'd bring the methods we use into top-level

api/api_chain.go
```diff
@@ -0,0 +1,166 @@
...
+
+	"github.com/filecoin-project/lotus/api"
+	apitypes "github.com/filecoin-project/lotus/api/types"
+	"github.com/filecoin-project/lotus/chain/actors/builtin/miner"
+	"github.com/filecoin-project/lotus/chain/types"
+	"github.com/filecoin-project/lotus/journal/alerting"
+)
+
+// ChainChainSubsetForForest is a subset of the Filecoin API that is supported by Forest.
+type ChainSubsetForForest interface {
```
magik6k, line 33: Imo we can just call this e.g. `CurioChainRPC` and use everywhere.

magik6k: We should also add a static check somewhere below asserting that lotus api.FullNode satisfies this interface

```
var _ CurioChainRPC = api.FullNode(nil)
```

deps/apiinfo.go
```diff
@@ -75,20 +79,96 @@ func getFullNodeAPIV1Curio(ctx *cli.Context, ainfoCfg []string, opts ...cliutil.
...
+// newChainNodeRPCV1 creates a new http jsonrpc client.
+func newChainNodeRPCV1(ctx context.Context, addr string, requestHeader http.Header, opts ...jsonrpc.Option) (api.Chain, jsonrpc.ClientCloser, error) {
+	var res api.ChainStruct
+	closer, err := jsonrpc.NewMergeClient(ctx, addr, "Filecoin",
+		api.GetInternalStructs(&res), requestHeader, append([]jsonrpc.Option{jsonrpc.WithErrors(RPCErrors)}, opts...)...)
+
+	return &res, closer, err
+}
+
+func FullNodeProxy[T api.Chain](ins []T, outstr *api.ChainStruct) {
```
magik6k, line 118: Oh I really like bringing this here, should allow us to eventually port smarts from https://github.com/filecoin-project/lotus/pull/11470 here + add some more Curio-specific ones.

magik6k (commented):
Might be simpler to always directly use the Curio subset interface, the aliases shouldn't really be needed.

Also would be neat to have some test which e.g. tries to comment out the methods in the interface one at a time, and if it manages to compile curio with a method commented out it would complain loudly that a method appears to be unused

api/api_chain.go
snadrus: no, rm

snadrus: haha, we use nothing. good catch.

api/proxy_gen.go
snadrus: ChainSubsetForForest (renamed) is actually built by the same generator calls since it doesn't need a build-flag to exist: just a build-flag to activate.

api/api_chain_all.go
snadrus: I think we all work better when we have the full API available to us in the IDE. Please resolve this.

deps/apiinfo.go
snadrus: 👍

LexLuthr (approved):

PR #63, opened 2024-06-19
no more lotus build

snadrus:
Dropping all deps that get to lotus/build
waiting on https://github.com/filecoin-project/lotus/pull/12112


magik6k (approved):
2 small nitpicks, other than that looks good, just:
* Needs `main` merge-in
* Needs lotus dep update if not on latest lotus master
* After merge would be a good idea to re-test guided-setup

PR #66, opened 2024-06-22
The network scheduler's optimization first-pass.

snadrus:


harmony/harmonytask/task_type_handler.go
```diff
@@ -150,7 +260,8 @@ canAcceptAgain:
 				}
 			}
 			ids = tryAgain
-			goto top
+			b, _ := h.considerWork(from, ids)
```
LexLuthr, line 263: Should we consider work again? Given that BidLogic is single threaded, shouldn't we just dump the win and let someone else take it?

harmony/harmonytask/task_type_handler.go
```diff
@@ -63,45 +66,151 @@ const (
...
+		WHERE owner_id IS NULL AND name=$1
+		ORDER BY creation_time ASC
+		LIMIT $2`, h.TaskTypeDetails.Name, numBidders)
+	return unownedTasks, err
+}
+
+// BidLogic is called to attempt to start work on a task-id of this task type.
+// It presumes single-threaded calling, so there should not be a multi-threaded re-entry.
+// The only caller should be the one work poller thread.
+func (h *taskTypeHandler) BidLogic(from string, ids []TaskID, sch *SchedulingInfo) (workAccepted, liveBids bool) {
```
LexLuthr, line 132: A godoc might be good for BidLogic.

harmony/harmonytask/task_type_handler.go
```diff
@@ -63,45 +66,151 @@ const (
...
+	if err != nil {
+		log.Error("Unable to read harmony_machine_details ", err)
+		return nil, err
+	}
+	var unownedTasks []TaskID
+	err = h.TaskEngine.db.Select(h.TaskEngine.ctx, &unownedTasks, `SELECT id
+		FROM harmony_task
+		WHERE owner_id IS NULL AND name=$1
+		ORDER BY creation_time ASC
+		LIMIT $2`, h.TaskTypeDetails.Name, numBidders)
```
LexLuthr, line 125: I don't understand why we are limiting to the number of bidders here. Wouldn't this break everyone should bid on all possible tasks?

harmony/harmonydb/sql/20240621-harmony-mch-task-list.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE harmony_task ADD COLUMN created_time TIMESTAMP NOT NULL DEFAULT current_timestamp;
+CREATE TABLE harmony_task_bid (
+    id SERIAL PRIMARY KEY NOT NULL,
```
LexLuthr, line 3: Should we create a constraint that this ID will only exists if it exists in harmony_task table?

LexLuthr (commented):
Bidding logic needs documentation. Some questions around retry.

harmony/harmonytask/task_type_handler.go
snadrus: I'm handling the case when 10000 of something drops in to the queue. 
Do we spend minutes in bidding just to rebid? 
    That could break the healthy uptake in jobs and the bidding timers. 
Do we cache our bids?
  That limits the usefulness of bidding as it cannot include capacity.
So the remaining option is to universally limit the max number of items we care to bid on. 
  - A limit of 1 could be too low for big networks. 
  - A limit of "everyone's capacity for this task" is closer to right, but requires (slow) per-cycle pre-negotiation.
  - A limit of "1x bidders" will keep work flowing, but may take multiple bid rounds to fill bigger machines. 
A future adjustment could include more than 1x per machine if the "handful of behemoths" clusters suffer.

harmony/harmonytask/task_type_handler.go
snadrus: We's already spent the effort to:
- decide this is the most important task we can work on 
- decide that we have resources to work on it 
- read a list from SQL
so it seems unfortunate to leave one on the table, and further, it's a priority inversion to abandon this work for other, lower-priority work.

harmony/harmonytask/task_type_handler.go
snadrus: I don't understand

snadrus: I made it private. I will describe it in godoc

snadrus:
I will build out the first user of this new API in a sub-branch of this one so that it becomes a clean example of how to do other ones.

snadrus:
TO FIX: There is a priority inversion because bidded-on jobs aren't considered "used resources" so lower-priority jobs get scheduled instead. 
Fix possibilities:
A) Virtual resource depletion: only schedule lower tasks as-if we won all bids.
B) Stop on first bid-made: simpler, but could go a long time without scheduling anything.
C) Have a cycle like:  (1) Accept wins (2) schedule sub-level jobs (3) Bid new jobs. 
      This one is probably the best option.

PR #78, opened 2024-07-02
Revert "FFISelect improvements"

magik6k:
Reverts filecoin-project/curio#70

I don't think we even want to not spawn gpu subprocesses, really we probably want to spawn some non-gpu tasks with ffiselect as well.


snadrus:
I figure this will make the common case faster, but certainly it's nice to have cancel-able tasks.

snadrus (approved):

PR #79, opened 2024-07-02
Mongo UI fixes

snadrus:
- Recent Wins Needs Refreshing: every 10m
- Sync Status indicates No when "Out of Sync" 
- `curio test window-post task` does not handle multiple minerIDs. 
- There is an issue with displaying or managing Cluster Task Miner IDs when dealing with multiple actors.

TODO:
- Create Log File on Initialization
- Make Redeclare Faster and Not Run on a Single Thread, and mute it from the logs.


web/static/win-stats.mjs
```diff
@@ -14,6 +14,7 @@ class WinStats extends LitElement {
 
     async loadData() {
         this.data = await RPCCall('WinStats');
+        setTimeout(() => this.loadData(), 10 * 60 * 1000); // 10 minutes
```
magik6k, line 17: 10 min in a little high, you don't win that often, but I'd still lower this to maybe 2 minutes

web/static/cluster-tasks.mjs
```diff
@@ -0,0 +1,52 @@
...
+                        <td>${!this.OwnerID ? html`<a href="/hapi/node/{{.OwnerID}}">{{.Owner}}</a>{{end}}</td>`:''}
+                    </tr>
+                    `)}
+                </tbody>
+            </table>
+        `;
+    }
+}
+
+customElements.define('storage-gc-stats', StorageGCStats);
```
magik6k, line 52: This is definitely wrong

magik6k (changes requested):

LexLuthr (approved):

PR #82, opened 2024-07-10
Feat/harmony cron

snadrus:
Cron offers At() which schedules a task to be made that associates itself with an arbitrary table. 
This is an unprivileged task: any task can do this.
Usage:
  Fill your custom table with the custom values that matter to you.  (returning rowID)
  It should have an 'id' and 'task_id' column

  deps.At(
      time.Now().Add(3 * time.Hour), 
      "FooTask", 
      "tableFoo", 
       rowID)

2 fragile points: 
   1. it adds to harmony_task which beforehand only harmonytask did. 
   2. It required a PLSQL func to avoid the sqlInjection system complaining about an arbitrary table.


snadrus:
Question for reviewer: 
   We currently take this low-effort work on all machines. Should we change CanAccept() to be pickier, or is it just not worth the hassle?

snadrus:
This needs a DB-based unit test. We should hold until we have that framework up.

LexLuthr:
Closing this as we decided not to pursue this based on complexity and limited use case from users.

Issue #84, opened 2024-07-10
UI config layer improvements


LexLuthr:
We need to remove 24h tasks column from machine(node) table and add 2 new columns

1. Layers used by node
2. Tasks supported by the node

We should rename "config layer available" to "Source layer" in Actor summary.
<img width="1152" alt="Screenshot 2024-07-10 at 6 22 02 PM" src="https://github.com/filecoin-project/curio/assets/88259624/ddeaf57b-9846-4178-b5dc-64dbda2500e6">

magik6k:
Much agree

PR #85, opened 2024-07-10
Fix/UI 2

snadrus:
https://github.com/filecoin-project/curio/issues/84 

Fix:
 - remove 24h tasks column from machine(node) table and add 2 new columns
      Layers used by node
     Tasks supported by the node
- We should rename "config layer available" to "Source layer" in Actor summary.


LexLuthr:
@snadrus Can we have some screenshots for this? It is hard to understand what rendering will look like?

cmd/curio/rpc/rpc.go
```diff
@@ -244,7 +245,7 @@ func (p *CurioAPI) LogSetLevel(ctx context.Context, subsystem, level string) err
 	return logging.SetLogLevel(subsystem, level)
 }
 
-func ListenAndServe(ctx context.Context, dependencies *deps.Deps, shutdownChan chan struct{}) error {
+func ListenAndServe(ctx context.Context, dependencies *deps.Deps, activeTasks []harmonytask.TaskInterface, shutdownChan chan struct{}) error {
```
LexLuthr, line 248: Can we avoid doing all this if we just add these details to the machine details table?

snadrus: No, this is about knowing which tasks we can get the SP_ID details for and which are "whole system" tasks. This is not a database matter but an introspection of which enabled tasks support this interface.

LexLuthr: Makes sense.

LexLuthr (approved):

PR #105, opened 2024-07-18
feat: add slack webhook alertmanager plugin

LexLuthr:


magik6k (approved):
Really cool

beck-8:
Why add this? Is it duplicate of altermanager?
I think it is unnecessary to spend too much time on this useless/unnecessary alarm. Because a complete alarm system depends on metics and external systems.

Instead, I recommend using this type of tool and removing all other additional plugins.
https://github.com/caronc/apprise-api

PR #119, opened 2024-07-23
WIP Ensure the cli's test commands work | waiting PR 121

snadrus:
This tests the `curio test window-post *` set.


cmd/curio/test-cli.go
```diff
@@ -119,7 +119,7 @@ var wdPostTaskCmd = &cli.Command{
 
 		for len(taskIDs) > 0 {
 			time.Sleep(time.Second)
-			err = deps.DB.QueryRow(ctx, `SELECT task_id, result FROM harmony_test WHERE task_id IN $1`, taskIDs).Scan(&taskID, &result)
+			err = deps.DB.QueryRow(ctx, `SELECT task_id, result FROM harmony_test WHERE task_id IN ($1)`, taskIDs).Scan(&taskID, &result)
```
LexLuthr, line 122:
```suggestion
			err = deps.DB.QueryRow(ctx, `SELECT task_id, result FROM harmony_test WHERE task_id = ANY($1)`, taskIDs).Scan(&taskID, &result)
```

itests_with_chain/cli_post_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"golang.org/x/exp/rand"
+)
+
+type CliThing struct {
+	*exec.Cmd
+	*bytes.Buffer
+}
+
+func CliEnv() func(name string, args ...string) *CliThing {
```
LexLuthr, line 19: CURIO_HARMONYDB_HOSTS needs to be set to "yugabyte" to talk to the docker container. Explains the error in failed test.

LexLuthr (changes requested):
I understand what we are testing here but I don't see why we are doing Forest setup.

snadrus:
> I understand what we are testing here but I don't see why we are doing Forest setup.

I need to get a full Curio CLI running, so it either needs a chain server. What is the best way to do that?

LexLuthr:
Closing this as it requires complete refactor based on current code.

PR #120, opened 2024-07-24
dramatically simpler user task system

snadrus:
Here's a sketch of a dramatically simpler system than what we had before. 
There was as little change as possible to the task engine: I just needed a way to peaceably return the task to the queue without a failure getting logged. 

It's not a perfect system (see downsides), but it's so much simpler to reason about that I think we can verify that it would work. 

Notice the interface passthru.

### DIFF
--- a/deps/config/types.go
+++ b/deps/config/types.go
@@ -241,6 +241,23 @@ type CurioSubsystemsConfig struct {
 
 	// The address that should listen for Web GUI requests.
 	GuiAddress string
+
+	// UserScheduleURL are the URLs for the user schedule optimization service. Please preceed each with
+	// the name of the task that it should be used for, followed by a comma and the URL. For example:
+	// "SealSDR,http://localhost:8080/schedule"
+	// This http endpoint gets a POST request with the following JSON body:
+	// {
+	//   "task_id": "task_id",
+	//   "task_type": "task_type",
+	//   "workers": ["worker1", "worker2"]
+	// }
+	// And looks for a 200 response with the following JSON body:
+	// {
+	//   "worker": "worker1"
+	//   "timeout": 60
+	// }
+	// Timeout in seconds until it will be rescheduled.
+	UserScheduleURL []string
 }
 type CurioFees struct {
 	DefaultMaxFee      types.FIL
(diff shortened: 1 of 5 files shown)


harmony/taskhelp/usertaskmgt/usertaskmgt.go
```diff
@@ -0,0 +1,152 @@
...
+	if err != nil {
+		return false, xerrors.Errorf("could not get owner: %w", err)
+	}
+	if owner == t.hostAndPort {
+		return t.TaskInterface.Do(id, stillMe)
+	}
+	var workerList []string
+	err = t.db.Select(context.Background(), &workerList, `SELECT host_and_port 
+	FROM harmony_machines m JOIN harmony_machine_details d ON d.machine_id=m.id 
+	WHERE tasks LIKE $1`, "%,"+t.name+",%")
```
magik6k, line 115: Only thing I worry about here is that the worker list technically includes workers which won't be able to take the work, but really that's probably fine.

deps/config/types.go
```diff
@@ -241,6 +241,23 @@ type CurioSubsystemsConfig struct {
 
 	// The address that should listen for Web GUI requests.
 	GuiAddress string
+
+	// UserScheduleURL are the URLs for the user schedule optimization service. Please preceed each with
+	// the name of the task that it should be used for, followed by a comma and the URL. For example:
+	// "SealSDR,http://localhost:8080/schedule"
```
magik6k, line 247: Nit:
```suggestion
	// "SDR,http://localhost:8080/schedule"
```

magik6k (commented):
Seeded to stare at this for a bit to figure out the flows, but in general looks good.

Would be great to test this somehow, I don't think any of us will use this feature any time soon, so having that test is quite important.

Would also be nice to implement some small Go server implementing some simple policy like round-robin serving this API - mostly as a base for people to take and modify to fit their needs

harmony/taskhelp/usertaskmgt/usertaskmgt.go
snadrus: We have an expiration of the "appointment" to handle the extremes of that situation.

snadrus:
Also the Do() error cases need intelligent handling so as not to fail tasks on server down.  
User configuration will be needed to determine the best course of action at server down.

PR #124, opened 2024-07-26
no string-to-array

snadrus:
Stop using broken YB function string_to_array()

### DIFF
--- a/tasks/window/compute_task.go
+++ b/tasks/window/compute_task.go
@@ -6,7 +6,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"sort"
-	"strings"
 	"time"
 
 	logging "github.com/ipfs/go-log/v2"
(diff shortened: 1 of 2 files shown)


tasks/window/compute_task.go
```diff
@@ -289,14 +288,14 @@ func (t *WdPostTask) CanAccept(ids []harmonytask.TaskID, te *harmonytask.TaskEng
...
+		`SELECT 
 			task_id,
 			sp_id,
 			proving_period_start,
 			deadline_index,
 			partition_index
-	from wdpost_partition_tasks 
-	where task_id IN (SELECT unnest(string_to_array($1, ','))::bigint)`, strings.Join(lo.Map(ids, entToStr[harmonytask.TaskID]), ","))
+	FROM wdpost_partition_tasks 
+	WHERE task_id IN ($1)`, lo.Map(ids, func(t harmonytask.TaskID, _ int) int { return int(t) }))
```
LexLuthr, line 298:
```suggestion
	WHERE task_id = ANY($1)`, lo.Map(ids, func(t harmonytask.TaskID, _ int) int { return int(t) }))
```

LexLuthr (changes requested):
Postgres syntax is incorrect for lookup in array

tasks/window/compute_task.go
```diff
@@ -289,14 +288,14 @@ func (t *WdPostTask) CanAccept(ids []harmonytask.TaskID, te *harmonytask.TaskEng
...
+		`SELECT 
 			task_id,
 			sp_id,
 			proving_period_start,
 			deadline_index,
 			partition_index
-	from wdpost_partition_tasks 
-	where task_id IN (SELECT unnest(string_to_array($1, ','))::bigint)`, strings.Join(lo.Map(ids, entToStr[harmonytask.TaskID]), ","))
+	FROM wdpost_partition_tasks 
+	WHERE task_id IN ANY($1)`, lo.Map(ids, func(t harmonytask.TaskID, _ int) int { return int(t) }))
```
LexLuthr, line 298:
```suggestion
	WHERE task_id IN ANY($1)`, lo.Map(ids, func(t harmonytask.TaskID, _ int) int64 { return int64(t) }))
```

LexLuthr (changes requested):
task_id is a bigint type in DB so we need to match that in Golang

snadrus:
That's not a problem at all:
https://www.postgresql.org/docs/current/functions-comparison.html

LexLuthr (approved):

PR #126, opened 2024-07-29
chore: release v1.23.0

LexLuthr:
- [ ] Update docs

### DIFF
--- a/build/version.go
+++ b/build/version.go
@@ -41,7 +41,7 @@ func BuildTypeString() string {
 }
 
 // Intent: Major.Network.EvenReleases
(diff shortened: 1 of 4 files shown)


build/version.go
```diff
@@ -41,7 +41,7 @@ func BuildTypeString() string {
 }
 
 // Intent: Major.Network.EvenReleases
```
snadrus, line 43: This intent comment is a bit off: 
'EvenReleases' should say something like 'Updates'

LexLuthr: Changed it to "Patch" as that was what we decided a while back.

magik6k (approved):

Issue #104, opened 2024-07-18
Snap: Review / Improve expiration handling


magik6k:
https://github.com/filecoin-project/curio/pull/52#discussion_r1680114012

At the same time we could optimize the sector picker to never select sectors in their proving window

PR #137, opened 2024-08-09
Fix Sector selection in snap ingester

magik6k:
Should address all concerns from https://github.com/filecoin-project/curio/issues/104

Definitely fixes sector selection logic.

### DIFF
--- a/market/deal_ingest_snap.go
+++ b/market/deal_ingest_snap.go
@@ -236,6 +236,11 @@ func (p *PieceIngesterSnap) AllocatePieceToSector(ctx context.Context, maddr add
 		piece.PublishCid = nil
 	}
 
+	head, err := p.api.ChainHead(ctx)
+	if err != nil {
+		return api.SectorOffset{}, xerrors.Errorf("getting chain head: %w", err)
+	}
+
 	var maxExpiration int64
 	vd.isVerified = piece.PieceActivationManifest.VerifiedAllocationKey != nil
 	if vd.isVerified {
@@ -253,7 +258,7 @@ func (p *PieceIngesterSnap) AllocatePieceToSector(ctx context.Context, maddr add
 		vd.tmin = alloc.TermMin
 		vd.tmax = alloc.TermMax
 
-		maxExpiration = int64(piece.DealSchedule.EndEpoch + alloc.TermMax)
+		maxExpiration = int64(head.Height() + alloc.TermMax)
 	}
 	propJson, err = json.Marshal(piece.PieceActivationManifest)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


magik6k:
The last commit implements the fix also done in https://github.com/filecoin-project/lotus/pull/12342; We'll also need that for PoRep side, opened an issue for it - https://github.com/filecoin-project/curio/issues/138

market/deal_ingest_snap.go
```diff
@@ -253,7 +258,7 @@ func (p *PieceIngesterSnap) AllocatePieceToSector(ctx context.Context, maddr add
 		vd.tmin = alloc.TermMin
 		vd.tmax = alloc.TermMax
 
-		maxExpiration = int64(piece.DealSchedule.EndEpoch + alloc.TermMax)
+		maxExpiration = int64(head.Height() + alloc.TermMax)
```
LexLuthr, line 261: made add 89(instead of 90) days in epoch as well to allow for a bigger pool?

market/deal_ingest_snap.go
```diff
@@ -309,16 +314,47 @@ func (p *PieceIngesterSnap) AllocatePieceToSector(ctx context.Context, maddr add
...
+		}
+
+		sectorLifeTime := si.Expiration - head.Height()
+		if sectorLifeTime < 0 {
+			return false, xerrors.Errorf("sector lifetime is negative!?")
+		}
+		if piece.DealSchedule.EndEpoch > si.Expiration {
+			return false, xerrors.Errorf("sector expiration is too soon: %d < %d", si.Expiration, piece.DealSchedule.EndEpoch)
+		}
+		if maxExpiration != 0 && si.Expiration > abi.ChainEpoch(maxExpiration) {
```
LexLuthr, line 344: This is too hard. We have 90 days grace period. We should not reject those sectors.

market/deal_ingest_snap.go
```diff
@@ -368,14 +409,23 @@ func (p *PieceIngesterSnap) allocateToExisting(ctx context.Context, piece lpiece
...
 			if sec.currentSize+psize <= abi.PaddedPieceSize(p.sectorSize) {
 				if vd.isVerified {
-					sectorLifeTime := sec.latestEndEpoch - sec.earliestStartEpoch
+					si, err := p.api.StateSectorGetInfo(ctx, p.miner, sec.number, types.EmptyTSK)
+					if err != nil {
+						log.Errorw("getting sector info", "error", err, "sector", sec.number, "miner", p.miner)
+						continue
+					}
+
+					sectorLifeTime := si.Expiration - head.Height()
```
LexLuthr, line 418: Isn't this redundant if we check this when we open a new sector?

tasks/seal/task_submit_commit.go
```diff
@@ -428,41 +428,44 @@ type AllocNodeApi interface {
...
-	if precomitInfo.Info.Expiration < ts.Height()+alloc.TermMin {
-		return xerrors.Errorf("sector expiration %d is before than allocation TermMin %d for piece %s", precomitInfo.Info.Expiration, ts.Height()+alloc.TermMin, piece.CID.String())
+	if expiration < ts.Height()+alloc.TermMin {
+		tooLittleBy := ts.Height() + alloc.TermMin - expiration
+
+		return true, xerrors.Errorf("sector expiration %d is before than allocation TermMin %d for piece %s (should be at least %d epochs more)", expiration, ts.Height()+alloc.TermMin, piece.CID.String(), tooLittleBy)
 	}
-	if precomitInfo.Info.Expiration > ts.Height()+alloc.TermMax {
-		return xerrors.Errorf("sector expiration %d is later than allocation TermMax %d for piece %s", precomitInfo.Info.Expiration, ts.Height()+alloc.TermMax, piece.CID.String())
+	if expiration > ts.Height()+alloc.TermMax {
```
LexLuthr, line 460: 90 days grace period check here as well.

LexLuthr (changes requested):
We missed an important point. If a sector is extended using `sptool` then our DB would not know about it. Thus our sector selection is never going to be correct. We need a IAmBored job which updated the sector expiration every few hours in the DB. Only for sector_meta table. 
Or we can check expiration of each sector online and then do the selection.
IAmBored seems a cleaner approach.

magik6k:
> We need a IAmBored job which updated the sector expiration every few hours in the DB

We have that: https://github.com/filecoin-project/curio/blob/main/tasks/metadata/task_sector_expirations.go

market/deal_ingest_snap.go
magik6k: TermMax already contains that grace period.

f05 allocations set TermMin to deal duration and TermMax to deal duration+90 days

market/deal_ingest_snap.go
magik6k: See above. This is exactly as loose as it can be. maxExpiration is the highest allowable sector expiration based on termmax

market/deal_ingest_snap.go
magik6k: Technically, maybe, but time is moving forward, so it doesn't hurt to double check just in case

tasks/seal/task_submit_commit.go
magik6k: TermMax is that grace period

LexLuthr:
> > We need a IAmBored job which updated the sector expiration every few hours in the DB
> 
> We have that: https://github.com/filecoin-project/curio/blob/main/tasks/metadata/task_sector_expirations.go

I don't know how I missed that. 🤦

LexLuthr (approved):
90 days grace period is a f05 market deal thing and not part of power/miner actor.

PR #145, opened 2024-08-12
webui: Fix chain connectivity status display

magik6k:
The ok status was this overflowing alert widget for quite some time, looked really ugly, so made it better.

2024-08-12-232617_820x153_scrot


web/static/chain-connectivity.mjs
```diff
@@ -24,15 +24,6 @@ window.customElements.define('chain-connectivity', class MyElement extends LitEl
...
         }
-        .success {
-            color: green;
-        }
-        .warning {
-            color: yellow;
-        }
-        .error {
-            color: red;
-        }
```
magik6k, line 35: those are in main.css with better colors

LexLuthr (approved):

PR #150, opened 2024-08-14
Document easy way for Xeon processors to build

snadrus:


documentation/en/installation.md
```diff
@@ -159,6 +159,11 @@ git checkout <release version>
        export CGO_CFLAGS="-D__BLST_PORTABLE__"
        ```
 
+    3. Also for old processors, if Curio cannot start because the processor generation is too old, add the variable:\
```
LexLuthr, line 162:
```suggestion
    3. Also for old processors, if Curio cannot start because the processor generation is too old, add the variable and rebuild the binaries:\
```

LexLuthr (approved):
Looks good with one nit

magik6k (commented):
Can we auto-detect maybe?

Something like
```
ifeq ($(shell grep -q avx512f /proc/cpuinfo && echo yes),yes)
    export GOAMD64 := v3
else
    export GOAMD64 := v1
endif
```

LexLuthr:
Auto detect might be a better way instead of docs.

PR #152, opened 2024-08-15
fix: invalid sp in the sector list page

strahe:
image

image

### DIFF
--- a/web/api/sector/sector.go
+++ b/web/api/sector/sector.go
@@ -56,17 +56,19 @@ func Routes(r *mux.Router, deps *deps.Deps) {
 
 func (c *cfg) terminateSectors(w http.ResponseWriter, r *http.Request) {
 	var in []struct {
-		MinerID uint64
-		Sector  uint64
+		MinerAddress string
+		Sector       uint64
 	}
 	apihelper.OrHTTPFail(w, json.NewDecoder(r.Body).Decode(&in))
 	toDel := make(map[minerDetail][]sec)
 	for _, s := range in {
-		maddr, err := address.NewIDAddress(s.MinerID)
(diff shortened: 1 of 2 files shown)


LexLuthr:
@strahe Make sure to get screenshots of this table. Especially the right most part. Since we are replacing a column, I don't think render should be affected but I would like to be sure. Also, this change will break `terminateSectors` method as the input for that method will change. You need do the following:
1. Update the index.html to use the updated column name "MinerAddress" as part of terminate &  delete
2. Update terminateSectors to set address string as input and handle from there

strahe:
image

image

web/api/sector/sector.go
```diff
@@ -96,7 +98,7 @@ func (c *cfg) getSectors(w http.ResponseWriter, r *http.Request) {
 	// TODO get sector info from chain and from database, then fold them together
 	// and return the result.
 	type sector struct {
-		MinerID        int64 `db:"miner_id"`
+		MinerID        int64 `db:"miner_id" json:"-"`
```
LexLuthr, line 101: Why do we need this json key?

strahe: Because the frontend part no longer requires MinerID, but the Go part still does, it is hidden.

LexLuthr: Okay. We should have been okay even if we sent this key to frontend.

strahe: Should I remove this json tag?

LexLuthr: Let's remove it for now. We might end up needing that ID later.

LexLuthr (approved):

PR #153, opened 2024-08-15
fix: do not use miner config for harmonyDB

LexLuthr:


LexLuthr:
Tested it in devnet and it works.

PR #156, opened 2024-08-19
fix: snap sector status

### LINKED ISSUE #155: Fix sector state lmrpc return for Snap deal
Right now the method SectorState is aligned with PoRep pipeline. A snap deal results in us sending "Proving" state too early. This results is Boost trying to index the deal where unsealed copy doesn't exists yet.

LexLuthr:
Fixes #155

### DIFF
--- a/market/fakelm/lmimpl.go
+++ b/market/fakelm/lmimpl.go
@@ -76,15 +76,19 @@ func (l *LMRPCProvider) SectorsStatus(ctx context.Context, sid abi.SectorNumber,
 	// TODO: Add snap, Add open_sector_pieces
 
 	var ssip []struct {
-		PieceCID *string `db:"piece_cid"`
-		DealID   *int64  `db:"f05_deal_id"`
-		DDOPAM   *string `db:"ddo_pam"`
-		Complete bool    `db:"after_commit_msg_success"`
-		Failed   bool    `db:"failed"`
-		SDR      bool    `db:"after_sdr"`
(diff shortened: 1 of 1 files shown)


market/fakelm/lmimpl.go
```diff
@@ -76,15 +76,19 @@ func (l *LMRPCProvider) SectorsStatus(ctx context.Context, sid abi.SectorNumber,
...
+		Complete        bool    `db:"after_commit_msg_success"`
+		Failed          bool    `db:"failed"`
+		SDR             bool    `db:"after_sdr"`
+		PoRep           bool    `db:"after_porep"`
+		Tree            bool    `db:"after_tree_r"`
+		IsSnap          bool    `db:"is_snap"`
+		Encode          bool    `db:"after_encode"`
+		SnapProve       bool    `db:"after_prove"`
+		SnapCommit      bool    `db:"after_prove_msg_success"`
+		SnapMoveStorage bool    `db:"after_move_storage"`
```
magik6k, line 91: Don't those appear as null when the sector is not in the snap pipeline?

magik6k (commented):
If this works then seems fine, but definitely needs to be tested manually at least

market/fakelm/lmimpl.go
LexLuthr: Thus COALESCE with FALSE so it can either be true or false.

LexLuthr:
Closed in favor of #165

PR #161, opened 2024-08-20
test harmonydb

snadrus:
moved from Lotus.


LexLuthr:
Metrics issue will not resolve even if you remove dependency on harmonyDB. Metrics are registered using a init(). Only way is to remove them from lotus side first and then depend on that version and recreate on our side. 
I would suggest rebasing on main once. Magik consolidated the DB creation for test into a single call.

LexLuthr (approved):

PR #165, opened 2024-08-22
lmrpc: Avoid table scans in SectorsStatus

magik6k:
Afaict the query behavior is exactly the same, it just uses 2 more CTE to guide postgres to use the correct indexes


magik6k:
Now this also contains fixes from https://github.com/filecoin-project/curio/pull/156

PR #167, opened 2024-08-22
sptool: sectors extend flags in days/deltas

magik6k:
* Make --from / --to flags (specifying source sector expiration range) accept delta-from-now (`+` prefix), and duration in days (`d` suffix)
* Make --extension flag accept duration in days (`d` suffix)
* Add a --no-cc flag which skips CC sector extensions


cmd/sptool/sector.go
```diff
@@ -349,9 +363,585 @@ var sectorsListCmd = &cli.Command{
...
+
+				sectors = append(sectors, abi.SectorNumber(id))
+			}
+		} else if cctx.IsSet("sector-file") {
+			sectors, err = getSectorsFromFile(cctx.String("sector-file"))
+			if err != nil {
+				return err
+			}
+		} else {
+			from := currEpoch + 120
```
snadrus, line 574: what are these constants?

snadrus (approved):

cmd/sptool/sector.go
magik6k: Those just match the default flags

PR #178, opened 2024-08-27
lmrpc: Maintain local piecerefs

magik6k:
This should fix all known issues with PiecePark
* When pieces-refs are created but fail to park, piece refs will now get removed correctly
  * This also fixes the issue when the PiecePark task can fail with EOF.. due to stale URLs in stale piece_refs
* Don't assume that data is only read from the first piece
* Correctly remove piece_refs which don't end up added to a sector
* Cleanup orphaned piece_refs on startup
  * This should cleanup all mess accumulated by existing nodes
* Fix an edge-case where a piece_ref could be added to a piece which is currently getting removed


magik6k:
NOTE: This PR won't land in `main`, but will be open until we land Curio-native markets.

Curio markets won't require this fix, and it is a little conflicting

magik6k:
No longer needed with new markets

PR #182, opened 2024-08-29
feat: index provider

LexLuthr:
This PR breaks IPNI advertisement and publishing into separate modules. If IPNI is enabled:
1. ipni_task: creates IPNI task for all deals which are not marked as complete. All deals are processed on storage server side and ads are stored in DB.
2. ipni-provider: This module publishes ads for each provider every 10 minutes. It is also responsible for providing data to IPNI during sync. New provider should be created during main market HTTP server startup and sub-router in the package can be attached to the main mux of the HTTP server.


tasks/indexing/task_ipni.go
```diff
@@ -0,0 +1,452 @@
...
+}
+
+func (I *IPNITask) Adder(taskFunc harmonytask.AddTaskFunc) {}
+
+func (I *IPNITask) GetSpid(db *harmonydb.DB, taskID int64) string {
+	var spid string
+	err := db.QueryRow(context.Background(), `SELECT sp_id FROM ipni_task WHERE task_id = $1`, taskID).Scan(&spid)
+	if err != nil {
+		log.Errorf("getting spid: %s", err)
+		return ""
```
gammazero, line 448: Is it OK to return an empty value, or should an error be returned?

tasks/indexing/task_ipni.go
```diff
@@ -0,0 +1,452 @@
...
+func (I *IPNITask) Do(taskID harmonytask.TaskID, stillOwned func() bool) (done bool, err error) {
+	ctx := context.Background()
+
+	var tasks []struct {
+		SPID   int64                   `db:"sp_id"`
+		Sector abi.SectorNumber        `db:"sector"`
+		Proof  abi.RegisteredSealProof `db:"reg_seal_proof"`
+		Offset int64                   `db:"sector_offset"`
+		CTXID  []byte                  `db:"context_id"`
+		RM     bool                    `db:"is_rm"`
```
gammazero, line 68: Naming suggestions since these are not abbreviations:
```suggestion
		CtxID  []byte                  `db:"context_id"`
		Rm     bool                    `db:"is_rm"`
```

tasks/indexing/task_ipni.go
```diff
@@ -0,0 +1,454 @@
...
+											sector, 
+											reg_seal_proof,
+											sector_offset,
+											context_id,
+											is_rm,
+											provider	
+										FROM 
+											ipni_task
+										WHERE 
+											task_id = $1;`, taskID)
```
gammazero, line 83: Since only one row is returned, it seems more efficient to use `db.QueryRow` and then `db.StructScan`

lib/ipni/ipni-provider/ipni-provider.go
```diff
@@ -0,0 +1,577 @@
...
+		Addresses  string
+		Signature  []byte
+		Entries    string
+		ContextID  []byte
+		IsRm       bool
+	}
+
+	err := p.db.Select(ctx, &ads, `SELECT context_id,
+       is_rm, previous, provider, addresses, 
+       signature, entries FROM ipni WHERE ad_cid = $1 AND provider = $2`, ad.String(), provider)
```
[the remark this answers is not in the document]
LexLuthr: This was my fault. Go fmt doesn't understand SQL parts. We run fmt as part of `make gen`. I have adjusted it to be more readable like the rest.

lib/ipni/ipni-provider/ipni-provider.go
```diff
@@ -0,0 +1,577 @@
...
+
+	topic := "/indexer/ingest/" + string(nn)
+
+	if deps.Cfg.Market.StorageMarketConfig.IPNI.TopicName != "" {
+		topic = deps.Cfg.Market.StorageMarketConfig.IPNI.TopicName
+	}
+
+	var announceURLs []*url.URL
+
+	for _, us := range deps.Cfg.Market.StorageMarketConfig.IPNI.DirectAnnounceURLs {
```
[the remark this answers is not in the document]
LexLuthr: Good catch. Memory management.

lib/ipni/chunker/chunker.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/ipni/go-libipni/ingest/schema"
+	"github.com/multiformats/go-multihash"
+
+	"github.com/filecoin-project/curio/lib/ipni/ipniculib"
+)
+
+var log = logging.Logger("chunker")
+
+const EntriesChunkSize = 16384
+const EntriesCacheCapacity = 4096
```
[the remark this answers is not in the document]
LexLuthr: Moved one to config. That was the plan. Made the other one internal const.

lib/ipni/ipni-provider/ipni-provider.go
```diff
@@ -0,0 +1,577 @@
...
+	switch req {
+	case "head":
+		sh, err := p.getHead(r.Context(), providerID)
+		if err != nil {
+			if errors.Is(err, ErrNotFound) {
+				http.Error(w, "", http.StatusNoContent)
+				return
+			}
+			log.Errorf("failed to get signed head for peer %s: %w", providerID, err)
+			http.Error(w, err.Error(), http.StatusInternalServerError)
```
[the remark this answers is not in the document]
LexLuthr: That makes sense. I will remove the errors.

tasks/indexing/task_ipni.go
LexLuthr: We expect only 1 row but it can return more. Thus the condition check below this. The DB is a custom wrapper of pgx. It kind of allows to neatly query everything into an array.

tasks/indexing/task_ipni.go
LexLuthr: Yes. That one is used by UI for showing which task is running on which node. Empty will simply results in one task missing from the list instead of an error for the whole page.

deps/config/types.go
```diff
@@ -614,3 +627,31 @@ type Libp2pConfig struct {
...
+type IPNIConfig struct {
+	// Enable set whether to enable indexing announcement to the network and expose endpoints that
+	// allow indexer nodes to process announcements. Enabled by default.
+	Enable bool
+
+	// TopicName sets the topic name on which the changes to the advertised content are announced.
+	// If not explicitly specified, the topic name is automatically inferred from the network name
+	// in following format: '/indexer/ingest/<network-name>'
+	// Defaults to empty, which implies the topic name is inferred from network name.
+	TopicName string
```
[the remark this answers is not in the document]
LexLuthr: "topic" only exists because `NewSignedHead` required it. Now that I look at it, I think we don't need topic. Can you please confirm?

deps/config/types.go
```diff
@@ -614,3 +627,31 @@ type Libp2pConfig struct {
 	// Format: multiaddress
 	NoAnnounceAddresses []string
 }
+
+type IPNIConfig struct {
+	// Enable set whether to enable indexing announcement to the network and expose endpoints that
+	// allow indexer nodes to process announcements. Enabled by default.
+	Enable bool
```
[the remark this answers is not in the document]
LexLuthr: `Disable` would make sense. This will completely disable everything. HTTP server is not part of this PR where this module will be called. When it is disabled, it will simply not create a provider. Thus no publishing. I should disable indexing jobs as well in case of disable. Don't make sense to create ads if we are not publishing.

harmony/harmonydb/sql/20240823-ipni.sql
```diff
@@ -0,0 +1,165 @@
+-- Table for storing IPNI ads
+CREATE TABLE ipni (
+    order_number BIGSERIAL PRIMARY KEY, -- Unique increasing order number
+    ad_cid TEXT NOT NULL,
+    context_id TEXT NOT NULL, -- abi.PieceInfo in Curio
```
magik6k, line 5: BYTEA?

lib/ipni/ipni-provider/ipni-provider.go
```diff
@@ -0,0 +1,577 @@
...
+	// announceURLs enables sending direct announcements via HTTP. This is
+	// the list of indexer URLs to send direct HTTP announce messages to.
+	announceURLs        []*url.URL
+	httpServerAddresses []multiaddr.Multiaddr
+}
+
+func NewProvider(api ipniAPI, deps *deps.Deps) (*Provider, error) {
+	c, err := lru.New[ipld.Link, datamodel.Node](chunker.EntriesCacheCapacity)
+	if err != nil {
+		return nil, xerrors.Errorf("creating new cache: %w", err)
```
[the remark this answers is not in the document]
magik6k: No, std errors are missing one extremely important feature in xerrors, which is the reason why xerrors was un-deprecated a while ago - xerrors.Errorf("%w") outputs errors with stack frames, which you can then print as a really nice error trace with `%+v`. Stdlib errors don't have that due to tiny added overhead, but we do want that feature. Those traces have saved me countless hours of confusion and many roundtrips to users to get better error messages

harmony/harmonydb/sql/20240823-ipni.sql
LexLuthr: Yes. I will update that.

harmony/harmonydb/sql/20240823-ipni.sql
```diff
@@ -0,0 +1,165 @@
...
+CREATE UNIQUE INDEX ipni_ad_cid ON ipni(ad_cid);
+
+-- This index will speed up lookups based on the ad_cid, which is frequently used to identify specific ads
+CREATE UNIQUE INDEX ipni_context_id ON ipni(context_id, ad_cid, is_rm);
+
+-- Since the get_ad_chain function relies on both provider and ad_cid to find the order_number, this index will optimize that query:
+CREATE INDEX ipni_provider_ad_cid ON ipni(provider, ad_cid);
+
+
+CREATE TABLE ipni_head (
```
magik6k, line 33: TODO (me): Maybe this doesn't need to exist and can just be max(ipni.order_number)

market/ipni/chunker/chunker.go
```diff
@@ -0,0 +1,156 @@
...
+	offset uint64
+}
+
+// CarMultihashIterator constructs a new MultihashIterator from a CAR index.
+//
+// This iterator supplies multihashes in deterministic order of their
+// corresponding CAR offset. The order is maintained consistently regardless of
+// the underlying IterableIndex implementation. Returns error if duplicate
+// offsets detected.
+func CarMultihashIterator(idx index.IterableIndex) (*SliceMhIterator, error) {
```
magik6k, line 125: TODO (me): Kill this with fire

Makefile
magik6k, line 1: TODO (me): Seems like this snuck back into here again

magik6k (approved):
Will work on this on top of the main branch

bajtos:
FYI: I am working on a FRC to document the Retrieval Checking Requirements, see https://github.com/filecoin-project/FIPs/pull/1089

PR #183, opened 2024-08-29
Move storiface into Curio

magik6k:
This PR moves storiface (and legacy partialfile) from lotus into curio to untie it from lotus and get us all the flexibility needed with storage types.

Needed for unseal


lib/storiface/filetype.go
```diff
@@ -0,0 +1,460 @@
...
+
+	"golang.org/x/xerrors"
+
+	"github.com/filecoin-project/go-state-types/abi"
+)
+
+// FTUnsealed represents an unsealed sector file.
+// FTSealed represents a sealed sector file.
+// FTCache represents a cache sector file.
+// FTUpdate represents an update sector file.
```
snadrus, line 14: what's an update?

lib/storiface/storage.go
```diff
@@ -0,0 +1,247 @@
...
+type StorageConfig struct {
+	StoragePaths []LocalPath
+}
+
+type LocalPath struct {
+	Path string
+}
+
+// LocalStorageMeta [path]/sectorstore.json
+type LocalStorageMeta struct {
```
snadrus, line 193: Is this a subset of StorageInfo ?
It may be nice to avoid dup by including the simpler type in the advanced type,.

snadrus (approved):

lib/storiface/filetype.go
LexLuthr: Snap sealed file

lib/storiface/storage.go
```diff
@@ -0,0 +1,247 @@
...
+
+type ReplicaUpdateProof []byte
+type ReplicaVanillaProofs [][]byte
+
+type ReplicaUpdateOut struct {
+	NewSealed   cid.Cid
+	NewUnsealed cid.Cid
+}
+
+type Sealer interface {
```
LexLuthr, line 76: Do we need this interface anymore?

tasks/window/fault_tracker.go
LexLuthr, line 1: Why do we need this?

LexLuthr (approved):
Looks good. Maybe some cleanup of things not used by Curio?

lib/storiface/filetype.go
magik6k: Updated replica (Snap sealed file)

lib/storiface/storage.go
magik6k: This one in the on-disk path config, the second one contains runtime information about the path

lib/storiface/storage.go
magik6k: Oh we don't

PR #186, opened 2024-09-02
feat: add service-setup subcommand

LexLuthr:
This PR modifies the following:
1. The guided-setup no longer create a curio.env file. It simply adds the DB related variables to .bashrc/.zshrc files. This allows users to run curio commands easily after guided-setup. In the end, it also prompts user to run `service-setup` command.

2. The new service-setup command creates the service files and then takes user input for DB and other vars to creates the /etc/curio.env file.

### DIFF
--- a/cmd/curio/guidedsetup/guidedsetup.go
+++ b/cmd/curio/guidedsetup/guidedsetup.go
@@ -6,6 +6,7 @@
 package guidedsetup
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"crypto/rand"
@@ -17,7 +18,9 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"os/user"
 	"path"
+	"path/filepath"
 	"strings"
 	"syscall"
 
@@ -26,7 +29,6 @@ import (
 	"github.com/manifoldco/promptui"
 	"github.com/mitchellh/go-homedir"
 	"github.com/samber/lo"
-	"github.com/snadrus/must"
 	"github.com/urfave/cli/v2"
 	"golang.org/x/text/language"
 	"golang.org/x/text/message"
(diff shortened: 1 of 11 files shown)


cmd/curio/guidedsetup/guidedsetup.go
```diff
@@ -244,41 +246,127 @@ func complete(d *MigrationData) {
...
-	places := append([]string{"/tmp/curio.env",
-		must.One(os.Getwd()) + "/curio.env",
-		must.One(os.UserHomeDir()) + "/curio.env"}, EnvFiles...)
-saveConfigFile:
-	_, where, err := (&promptui.Select{
-		Label:     d.T("Where should we save your database config file?"),
-		Items:     places,
-		Templates: d.selectTemplates,
-	}).Run()
+	// Determine the current user's shell
```
magik6k, line 250: I would definitely first ask if the user wants to put curio DB credentials in the shell rc file.

e.g. I have my dotfiles managed in git, and would be rather surprised to find new entries there without allowing/expecting them to be written there.

cmd/curio/guidedsetup/guidedsetup.go
```diff
@@ -244,41 +246,127 @@ func complete(d *MigrationData) {
...
+
+	// Append missing export lines
+	for key, added := range existingVars {
+		if !added {
+			lines = append(lines, exportLines[key])
+		}
+	}
+
+	// Reopen the file in write mode to overwrite with updated content
+	file, err = os.OpenFile(rcFilePath, os.O_WRONLY|os.O_TRUNC, 0644)
```
magik6k, line 339: We should close the previous instance of `file` first

Never open critical system files with O_TRUNC, that's a recipe for disaster.

I'd open in append mode, append a newline, then our entries.

LexLuthr: That was a disaster waiting to happen. Forgot to replace the o_trunc with o_append

LexLuthr:
I would say we don't really need this after so many users onboarding without any issue.

PR #189, opened 2024-09-03
feat: Update the UI for the config list page

anhoffa:
**List of changes:**
- updated the UI for the config list page

new:
curio-new-config-list

old:
curio-old-config-list

- added a `StyledLitElement` that allows the custom components to inherit global styles. Styling can be applied in two ways:
```js
class LitComponent extends StyledLitElement {
  static get styles() {
    return [
      StyledLitElement.styles,
      css`...`
    ]
  };
  ...
}
```
or
```js
class LitComponent extends StyledLitElement {
  ...
}

LitComponent.styles = css`...`
```
- changed url from `/config/` to `/pages/config/list/`
- in the layer list, changed the information about the node addresses to the number of nodes (per @magik6k's request)


snadrus:
Concern 1: Knowing which box is using a config and needs a restart is an important piece of this thing.

magik6k:
> Concern 1: Knowing which box is using a config and needs a restart is an important piece of this thing.

I don't think this is something we can fit reasonably on this page, it can be added to the layer/editor page later imo. Putting node lists here makes this page unreadable in any non-tiny cluster

magik6k:
What would help a lot more, but definitely isn't scope for this PR would be some kind of config versions page, showing nodes which run outdated configuration, which layers changed, and maybe have options to schedule restarts

(nodes could also pick up that their binaries got updated and also suggest restarting on that page)

strahe:
> What would help a lot more, but definitely isn't scope for this PR would be some kind of config versions page, showing nodes which run outdated configuration, which layers changed, and maybe have options to schedule restarts
> 
> (nodes could also pick up that their binaries got updated and also suggest restarting on that page)

This sounds very cool.

web/static/ux/css-reset.js
```diff
@@ -0,0 +1,21 @@
...
+        padding: 0;
+    }
+
+    ul, ol {
+        list-style: none;
+        padding: 0;
+    }
+
+    a {
+        text-decoration: none;
```
snadrus, line 19: I could do without obfuscating all links.

anhoffa: It should work the same as the one previously defined in [main.css](https://github.com/filecoin-project/curio/blob/3e37077a14f42f8b075576f37ed668cee31e6df3/web/static/ux/main.css#L76), though I agree that some default link styling would be nice, I was thinking about adjusting it in the next revamps.

magik6k (approved):

PR #194, opened 2024-09-04
feat: create http server

LexLuthr:
This PR add a HTTPS server and a loadbalancer to Curio market.

Pending:
- [ ] Config defaults (will add once server is finalised after the review)


documentation/en/configuration/default-curio-configuration.md
```diff
@@ -552,4 +552,79 @@ description: The default curio configuration
...
+
+  # ReadHeaderTimeout is amount of time allowed to read request headers
+  #
+  # type: time.Duration
+  #ReadHeaderTimeout = "0s"
+
+  # EnableCORS indicates whether Cross-Origin Resource Sharing (CORS) is enabled or not.
+  #
+  # type: bool
+  #EnableCORS = false
```
magik6k, line 600: Probably should default to true?

documentation/en/configuration/default-curio-configuration.md
```diff
@@ -552,4 +552,79 @@ description: The default curio configuration
...
+[HTTP]
+  # DomainName specifies the domain name that the server uses to serve HTTP requests.
+  #
+  # type: string
+  #DomainName = ""
+
+  # CertCacheDir path to the cache directory for storing SSL certificates needed for HTTPS.
+  #
+  # type: string
+  #CertCacheDir = ""
```
magik6k, line 565: I worry that letsencrypt won't be particularly happy when one tries to run the cluster with multiple http nodes, might be trying to put this in the DB somehow

documentation/en/configuration/default-curio-configuration.md
```diff
@@ -552,4 +552,79 @@ description: The default curio configuration
...
+  # EnableCORS indicates whether Cross-Origin Resource Sharing (CORS) is enabled or not.
+  #
+  # type: bool
+  #EnableCORS = false
+
+  # EnableLoadBalancer indicates whether load balancing between backend servers is enabled. It should only
+  # be enabled on one node per domain name.
+  #
+  # type: bool
+  #EnableLoadBalancer = false
```
magik6k, line 606: Honestly this I'd leave to real loadbalancer software, the Curio http server setup should be extremely simple, boiling down to all public http endpoints on all nodes behaving the same, then either pointing LB or dns at all of the endpoints

deps/config/types.go
```diff
@@ -548,3 +563,37 @@ type ApisConfig struct {
...
+	DomainName string
+
+	// ListenAddr is the address that the server listens for HTTP requests.
+	ListenAddr string
+
+	// ReadTimeout is the maximum duration for reading the entire or next request, including body, from the client.
+	ReadTimeout time.Duration
+
+	// WriteTimeout is the maximum duration before timing out writes of the response to the client.
+	WriteTimeout time.Duration
```
magik6k, line 579: Are we later able to set this higher for responses from e.g. the /piece endpoint?

magik6k (approved):

deps/config/types.go
LexLuthr: Yes. I will try to find better write defaults once this is merged. More practical ones.

PR #202, opened 2024-09-18
zh

snadrus:
Make Curio more approachable and friendly for speakers of Simplified Mandarin (Chinese). 

Docs:
zh category added
Labels have EACH of English and SImplified Mandarin. This allows:
 - English documentation subpage links with en --> zh changed should work. 
 - When portions of pages change, it will be easy to find that portion in the translated pages for updating. 

Coming soon:
- Docs diagrams
- CLI usage & descriptions
- Verifying this auto-translation is sane.


snadrus:
@LexLuthr Do you have SVGs of the diagram images? 
GPT's PNG --> SVG capabilities were a good start, but sub-par. It's direct translation was concept-art.

strahe:
Awesome, this would be very helpful.

LexLuthr (approved):
I would suggest merging this PR as is. Even if merged, this won't create the new docs in Gitbook. I still need to attach that part to the Gitbook side and start the sync for "zh" version.
We can fix remaining things in a follow up PR and then publish the "zh" version.

snadrus:
I've got a biggish addition to this pr that fixes the cli translations. It's a TODO. We also could use the diagrams translated.

snadrus:
OK I will merge. The holes it has are nothing compared to it missing entirely.

PR #205, opened 2024-09-19
fix: add detailed resource log

LexLuthr:


harmony/harmonytask/harmonytask.go
magik6k, line 1: Is this only done on startup or on every scheduling loop? If every sched loop we probably want this to be debug

LexLuthr: Yes. I will send these to debug. Branch was mostly created to try to pin the incorrect utilisation seen on a couple of miners. Ideally, I would prefer to avoid shipping these once we find the bug.

LexLuthr: Done. Moved to a separate logger with Debug. This should allow us to get us these logs only when required.

magik6k (approved):

PR #206, opened 2024-09-20
feat: Simpler IPNI

magik6k:
* Change IPNI entry creation to be dependent on number of blocks in a piece
  * Below 500k entries per piece (64kb block in 32G) read from DB
  * Above 500k re-create entries by doing range car-reads
  * No caches involved anymore, entries get recreated on the fly, with low latency
* Switched entries from dag-json to dag-cbor, apparently IPNI shouldn't care much, and in the future this will make it easy to switch to the much, much faster cbor-gen for entry creation.
  * Ads remain as dag-json

### DIFF
--- a/market/indexstore/indexstore.go
+++ b/market/indexstore/indexstore.go
@@ -52,19 +52,6 @@ type Record struct {
 	OffsetSize `json:"offsetsize"`
 }
 
-// Probability of a collision in two 20 byte hashes (birthday problem):
-// 2^(20*8/2) = 1.4 x 10^24
-const multihashLimitBytes = 20
-
-// trimMultihash trims the multihash to the last multihashLimitBytes bytes
-func trimMultihash(mh multihash.Multihash) []byte {
-	var idx int
-	if len(mh) > multihashLimitBytes {
-		idx = len(mh) - multihashLimitBytes
-	}
-	return mh[idx:]
-}
-
 var ErrNotFound = errors.New("not found")
 
 func normalizeMultihashError(m multihash.Multihash, err error) error {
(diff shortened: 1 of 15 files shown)


market/ipni/chunker/initial-chunker.go
```diff
@@ -0,0 +1,267 @@
...
+	prevChunks []carChunkMeta
+}
+
+type carChunkMeta struct {
+	link  ipld.Link
+	start int64
+	nodes int64
+}
+
+func NewInitialChunker() *InitialChunker {
```
LexLuthr, line 49: Can we init this with preallocated size/memory for carPending? it might be more efficient.

LexLuthr (approved):
Looks good!

market/indexstore/indexstore.go
```diff
@@ -295,3 +295,19 @@ func (i *IndexStore) PiecesContainingMultihash(ctx context.Context, m multihash.
...
 }
+
+func (i *IndexStore) GetPieceHashRange(ctx context.Context, piece cid.Cid, start multihash.Multihash, num int64) ([]multihash.Multihash, error) {
+	qry := "SELECT PayloadMultihash FROM PayloadToPiece WHERE PieceCid = ? AND PayloadMultihash >= ? LIMIT ?"
+	iter := i.session.Query(qry, piece.Bytes(), trimMultihash(start), num).WithContext(ctx).Iter()
+
+	var hashes []multihash.Multihash
+	var r []byte
+	for iter.Scan(&r) {
+		m := multihash.Multihash(r)
```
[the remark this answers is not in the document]
magik6k: Oooh that's a... really good question that kinda fundamentally changes an assumption here

market/ipni/chunker/initial-chunker.go
magik6k: Won't change that much, plus I intend to have this eventually rewritten to use cbor-gen everywhere anyways. Right now I'd guess that 99% of overhead will be in ipld-prime serialization/deserialization code.

PR #208, opened 2024-09-20
feat: retrievals

LexLuthr:


market/retrieval/piecehandler.go
```diff
@@ -0,0 +1,99 @@
...
+	w.Header().Set("Cache-Control", "public, max-age=29030400, immutable")
+}
+
+func serveContent(res http.ResponseWriter, req *http.Request, content io.ReadSeeker) {
+	// Note that the last modified time is a constant value because the data
+	// in a piece identified by a cid will never change.
+
+	if req.Method == http.MethodHead {
+		// For an HTTP HEAD request ServeContent doesn't send any data (just headers)
+		http.ServeContent(res, req, "", time.Time{}, nil)
```
magik6k, line 93: Won't this send incorrect content-length? Or does it not send that header if data is nil?

market/retrieval/remoteblockstore/metric.go
```diff
@@ -0,0 +1,123 @@
...
+
+import (
+	"go.opencensus.io/stats"
+	"go.opencensus.io/stats/view"
+
+	"github.com/filecoin-project/lotus/metrics"
+)
+
+// Distribution
+var defaultMillisecondsDistribution = view.Distribution(0.01, 0.05, 0.1, 0.3, 0.6, 0.8, 1, 2, 3, 4, 5, 6, 8, 10, 13, 16, 20, 25, 30, 40, 50, 65, 80, 100, 130, 160, 200, 250, 300, 400, 500, 650, 800, 1000, 2000, 3000, 4000, 5000, 7500, 10000, 20000, 50000, 100000)
```
magik6k, line 11: This is way too much imo, prometheus doesn't like that much arity in data, would maybe do 30% of those

market/retrieval/remoteblockstore/reader.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+		unsealed, err := ro.pp.IsUnsealed(ctx, sr, storiface.UnpaddedByteIndex(dl.Offset), dl.Length)
+		if err != nil {
+			merr = multierror.Append(merr, err)
+			continue
+		}
+
+		if !unsealed {
+			continue
+		}
```
magik6k, line 105:
```suggestion
```

This is redundant and will only slow things down, `pp.ReadPiece->tryReadUnsealedPiece->storage.Reader` already does this check internally

market/retrieval/remoteblockstore/remoteblockstore.go
```diff
@@ -0,0 +1,287 @@
...
+
+var log = logging.Logger("remote-blockstore")
+
+type idxAPI interface {
+	PiecesContainingMultihash(ctx context.Context, m multihash.Multihash) ([]cid.Cid, error)
+	GetOffsetSize(ctx context.Context, pieceCid cid.Cid, hash multihash.Multihash) (*indexstore.OffsetSize, error)
+}
+
+type pieceProviderAPI interface {
+	IsUnsealed(ctx context.Context, sector storiface.SectorRef, offset storiface.UnpaddedByteIndex, size abi.UnpaddedPieceSize) (bool, error)
```
magik6k, line 41: Ideally we don't need this one

market/retrieval/remoteblockstore/remoteblockstore.go
```diff
@@ -0,0 +1,287 @@
...
+			}(reader)
+
+			// Get the offset of the block within the piece (CAR file)
+			offsetSize, err := ro.idxApi.GetOffsetSize(ctx, pieceCid, c.Hash())
+			if err != nil {
+				return nil, fmt.Errorf("getting offset/size for cid %s in piece %s: %w", c, pieceCid, err)
+			}
+
+			// Seek to the section offset
+			readerAt := readerutil.NewReadSeekerFromReaderAt(reader, int64(offsetSize.Offset))
```
magik6k, line 166: Pretty sure this thing is like `io.NewSectionReader()` except duplicate

market/retrieval/remoteblockstore/remoteblockstore.go
```diff
@@ -0,0 +1,287 @@
...
+			// Read the block data
+			bufferSize := 4096
```
magik6k, line 168: Why do we not just read the correct size here?

magik6k (commented):
Looks pretty good, couple comments to address

documentation/en/configuration/default-curio-configuration.md
```diff
@@ -734,7 +740,13 @@ description: The default curio configuration
 
 
 [HTTP]
-  # DomainName specifies the domain name that the server uses to serve HTTP requests.
+  # Enable the HTTP server on the node
+  #
+  # type: bool
+  #Enable = false
```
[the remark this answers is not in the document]
LexLuthr: Removed the market duplicate. We will be using this HTTP server for everything so putting it under market seems unfair. Default to True might not be a good in this case. We don't want to start HTTP server on all nodes. Just select nodes user chooses.

market/retrieval/piecehandler.go
```diff
@@ -0,0 +1,99 @@
...
+		w.WriteHeader(http.StatusBadRequest)
+		stats.Record(ctx, remoteblockstore.HttpPieceByCid400ResponseCount.M(1))
+		return
+	}
+
+	// Get a reader over the piece
+	reader, err := rp.bs.GetSharedPieceReader(ctx, pieceCid)
+	if err != nil {
+		log.Errorf("server error getting content for piece CID %s: %s", pieceCid, err)
+		if strings.Contains(strings.ToLower(err.Error()), "no deals found") {
```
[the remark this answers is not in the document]
LexLuthr: I will convert it into a proper error.

market/retrieval/piecehandler.go
LexLuthr: I have set the content length later. Head request should not send Content-Length header.

market/retrieval/remoteblockstore/reader.go
LexLuthr: Removed this.

market/retrieval/remoteblockstore/remoteblockstore.go
LexLuthr: removed

market/retrieval/remoteblockstore/remoteblockstore.go
LexLuthr: That was a work around for 2k sectors in Boost. Now that we can reproduce the pieceReader panic, I have removed the custom buffer size and went with default.

market/retrieval/remoteblockstore/metric.go
LexLuthr: Dumped most of these.

magik6k (approved):

PR #209, opened 2024-09-20
This fixes timely automatic recovery tasks

JKK1:
This should fix the issues where a task is not recovered in time due to overlaps with other deadlines. (UNTESTED)

See: https://filecoinproject.slack.com/archives/C06LF5YP8S3/p1726747849478179


magik6k:
Hey, thanks for the PR!

I'll try to get to reviewing/testing this soon (probably not next week though), I'm slightly concerned that running recovery so often may strain the storage system a little much in some scenarios, but I'll have a stronger opinion after more careful review.

LexLuthr:
wdPost has changed since this PR was opened. We do half the storage read so it should be okay with current config. If there is still an issue, please update this PR with details.

PR #211, opened 2024-09-25
cunative: Only pull in native deps with a tag

magik6k:
This makes it possible to use curio as a library again.

Context: https://filecoinproject.slack.com/archives/C06GD1SS56Y/p1727193366887699


strahe:
Tested, it works.

LexLuthr (approved):

PR #214, opened 2024-09-26
feat: add ddo table

LexLuthr:


magik6k (approved):
I'd expect to see a PAM in that table, at least eventually; If this is enough for boost migration, then I guess it's fine

PR #215, opened 2024-09-26
fix: ipni issues

LexLuthr:
1. Switch to newer libipni and use headers to distinguish between ad and entry CID request. 
2. Use a cached reader for entry responses. IPNI will ask all the chunks in a row. Getting a reader each time is inefficient.
3. Fix Makefile issue
4. Fix IPNI provider HTTP URL string and remove unused variables


market/ipni/ipni-provider/ipni-provider.go
```diff
@@ -140,17 +159,23 @@ func NewProvider(d *deps.Deps) (*Provider, error) {
 		httpServerAddresses[i] = addr
 	}
 
+	prCache := ttlcache.NewCache()
+	_ = prCache.SetTTL(time.Minute * 10)
+	prCache.SetCacheSizeLimit(MaxCachedReaders)
```
magik6k, line 164: Maybe we should have this wrapped into a single, cached, sector reader utility in Deps? It's one more layer, but this code will likely get duplicated in other retrieval spots

market/ipni/ipni-provider/ipni-provider.go
```diff
@@ -491,59 +489,120 @@ func (p *Provider) handleGet(w http.ResponseWriter, r *http.Request) {
...
-						return
-					}
-					log.Errorf("failed to get entry %s for peer %s: %w", b.String(), providerID, err)
-					http.Error(w, "", http.StatusInternalServerError)
+
+		h := r.Header.Get(ipnisync.CidSchemaHeader)
+
+		switch h {
+		case ipnisync.CidSchemaAdvertisement:
+			ad, err := p.getAd(r.Context(), b, providerID)
```
magik6k, line 497: This switch would be a lot cleaner if p.getAd returned serialized bytes

market/ipni/ipni-provider/reader.go
magik6k, line 1: Isn't this the same cached reader we use for retrievals? (At least I somewhat vaguely remember writing this in boost)

magik6k (approved):
In general looks good, some comments about things which could be done a little more cleanly

market/ipni/ipni-provider/reader.go
LexLuthr: Yes. I will turn it into a dep and use it in IPNI and retrievals.

market/ipni/ipni-provider/ipni-provider.go
LexLuthr: If I don't return the actual ad then I cannot use the func for getHead() as head needs to be a signed advertisement. Either I add more code and process getHead() separately or just return the advertisement like now.

magik6k: Makes sense, I guess then we could just do it as a separate `getAdBytes` function for use here and keep `getAd` for head getting as is.

Either case no strong opinion, can also keep this in the current shape, it works too.

LexLuthr: Add the the bytes function. It makes get() easier.

market/ipni/ipni-provider/ipni-provider.go
```diff
@@ -331,98 +376,59 @@ func (p *Provider) getEntry(block cid.Cid, provider string) ([]byte, error) {
...
-		FROM
-			market_piece_deal mpd
-		INNER JOIN
-			sectors_meta sm
-		ON
-			mpd.sp_id = sm.sp_id
-			AND mpd.sector_num = sm.sector_num
-		WHERE piece_cid = $1`, piece)
+
+	reader, _, err := p.cpr.GetSharedPieceReader(ctx, piece)
```
magik6k, line 382: needs `defer reader.Close()`

magik6k (approved):
Looks good now, just needs the one-like reader close thing and then good to merge

market/ipni/ipni-provider/ipni-provider.go
LexLuthr: Good catch.

LexLuthr: Done

magik6k (approved):

PR #216, opened 2024-09-27
feat: implement retry wait mechanism for task retries

strahe:
The following changes have been made in this PR:
1. Added a `retries` field to the `harmony_task` table to track the number of retries for a task.
2. Added a `RetryWait` field to `TaskTypeDetails` to set the wait time before retrying a task, which can be `nil`.
3. Set `RetryWait` for the `SendMessage` task, while other tasks do not have it set (can be set as needed in the future).

### DIFF
--- a/harmony/harmonydb/sql/20240927-task-retrywait.sql
+++ b/harmony/harmonydb/sql/20240927-task-retrywait.sql
@@ -1,0 +1,1 @@
+ALTER TABLE harmony_task ADD COLUMN retries BIGINT NOT NULL DEFAULT 0;
--- a/harmony/harmonytask/harmonytask.go
+++ b/harmony/harmonytask/harmonytask.go
@@ -7,6 +7,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/samber/lo"
+
 	"go.opencensus.io/stats"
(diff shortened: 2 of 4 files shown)


LexLuthr:
I was thinking about this and maybe we should keep harmony scheduler mostly as is. We can define an additional field for retry in the task interface's TypeDetails(). This way not all tasks have to be retried with backoff. We can let most of the things have default behaviour and introduce backoff on the ones which require it. Then scheduler would skip scheduling for a task with a can accept deny. 
Any other ideas are welcome to keep harmonyTask general purpose.

strahe:
`RetryWait` is nil by default, and most tasks can ignore it, keeping their retry logic unchanged. The newly added retries field in the harmony_task table can be removed, but it makes it difficult to track the number of retries. RetryWait is not flexible enough but can still be used to set a simple delay strategy, such as a fixed 3 seconds.  

However, I still recommend keeping the retries field. This field is not only useful for determining retries but also for other statistical scenarios. It is more efficient than counting from the harmony_task_history table.

LexLuthr:
My bad! You have already gone with the model I thought. I should have reviewed the full PR first.

snadrus (approved):

PR #227, opened 2024-09-30
feat: PDP

magik6k:
Reading:
* https://github.com/storacha/RFC/pull/34
* https://github.com/FILCAT/pdp/blob/main/contracts/src/PDPService.sol

Roadmap:
* [x] Database schema
* [ ] PieceCID version correctness
* [x] HTTP API
  * [x] Basic structure
  * [x] Auth
  * [x] Data ingest
    * [x] POST /piece
    * [x] PUT /piece/upload/..
    * [x] Call notify
      * [x] make optional
    * [x] Store in long-term storage
  * [x] Data retrieval through /piece
    * [ ] Auth
  * [x] PDPTool
    * [x] Creating secret/tokens
    * [x] Ping tool
    * [x] Upload tool
    * [ ] Retrieve tool
    * [x] Piece set manipulation
  * [ ] Piece sets
    * [x] Create
    * [x] Delete
    * [x] Add root (w subroots)
    * [x] Del Root
    * [ ] Get root subroots
  * [x] Proof-set prover
* [ ] UI


pdp/handlers.go
```diff
@@ -0,0 +1,1020 @@
...
+				return false, fmt.Errorf("subroot CID %s not found or does not belong to service %s", cidStr, serviceLabel)
+			}
+		}
+
+		// Now, for each AddRootRequest, validate RootCID and prepare data for ETH transaction
+		for _, addRootReq := range req {
+			// Collect pieceInfos for subroots
+			pieceInfos := make([]abi.PieceInfo, len(addRootReq.Subroots))
+
+			var totalOffset uint64 = 0
```
hannahhoward, line 669: am I correct that reading this code right here provides indirect instructions on how to aggregate all the subroot pieces into a root CID? IOW, I can assume that I can basically want to write aggregation code that will satisfy this code by generating the same RootCID? (which appears to be all the pieces concatenated together). 

Also it would appear non-ffi is a fork of go-commp-utils -- is this public?

pdp/handlers_upload.go
```diff
@@ -0,0 +1,285 @@
...
+	// Verify that the request is authorized using ECDSA JWT
+	serviceID, err := p.verifyJWTToken(r)
+	if err != nil {
+		http.Error(w, "Unauthorized: "+err.Error(), http.StatusUnauthorized)
+		return
+	}
+
+	// Parse request body
+	var req struct {
+		PieceCID string `json:"pieceCid"`
```
hannahhoward, line 40: It looks like you're not requiring piececid v2 here, not in the put handler. Long term, I think we should require this. From the standpoint of the SP, I imagine them to want to filter at the piece level, and I imagine one of the main filter critieria to be size. They should know exactly what size piece they're about to accept before they accept it. 

Unless I'm wrong, a PieceCID v1 doesn't specify that, and absent a size parameter, they could be accepting anything from 32 bytes to 256 MB.

pdp/handlers_upload.go
```diff
@@ -0,0 +1,285 @@
...
+	"github.com/filecoin-project/curio/harmony/harmonydb"
+	"github.com/filecoin-project/curio/lib/dealdata"
+)
+
+var log = logger.Logger("pdp")
+
+// PieceSizeLimit in bytes
+var PieceSizeLimit = abi.PaddedPieceSize(256 << 20).Unpadded()
+
+func (p *PDPService) handlePiecePost(w http.ResponseWriter, r *http.Request) {
```
hannahhoward, line 30: This isn't a now problem, but I kind of forgot to mention that we really do all our uploads as traditional SHA256s of the actually blob of unpadded raw data, not PieceCIDs. We then calculate piece cids later for Filecoin. For demo purposes, we'll have our sidecar take in the data so we don't have to worry about this, and if we want to use the endpoint directly one day, we can discuss how to do this conversion.

lib/pieceprovider/piecepark_reader.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/ipfs/go-cid"
+	"golang.org/x/xerrors"
+
+	"github.com/filecoin-project/go-state-types/abi"
+
+	"github.com/filecoin-project/curio/lib/paths"
+	"github.com/filecoin-project/curio/lib/storiface"
+)
+
+type PieceParkReader struct {
```
hannahhoward, line 16: This is what finds and reads PDP pieces for get requests correct?

hannahhoward (approved):
I've mostly put in comments to check for understanding before I start writing code. But generally LGTM -- looks like it's compliant with the API we agreed on :)

hannahhoward:
am I correct that the retrieval handler already has been modified to support PDP pieces through the parked piece reader?

hannahhoward:
am I correct in believing that the next step on the API is that add root will send a location header to a creation status API for the root, then that will be used to get the root ID which is then possible to pass to the root detail and delete root APIs? Not crazy urgent just want to know.

magik6k (commented):
.

pdp/handlers_upload.go
magik6k: Yes, PieceCIDv2 is still todo, can quickly update if that's a blocker

lib/pieceprovider/piecepark_reader.go
magik6k: Yeah, PiecePark is the general "raw piece" storage in Curio, we eventually intend to serve all retrievals from it (instead of from the fr32-padded "unsealed" sector files), but PDP is the first place which will actually do reads directly from it.

magik6k:
> am I correct that the retrieval handler already has been modified to support PDP pieces through the parked piece reader?

Yes, correct, retrievals work already. No ACL mechanism yet, but it's not going to be hard to implement once we have a spec (you can see that all retrieval requests quite quickly become aware what piece they are reading from + ACLs should be quite cachable)

magik6k:
> am I correct in believing that the next step on the API is that add root will send a location header to a creation status API for the root, then that will be used to get the root ID which is then possible to pass to the root detail and delete root APIs? Not crazy urgent just want to know.

Other than delete all the things should be there. Operations on PDP proofsets (other than create) you should probably watch through eth events. I can also make the message-sending endpoints output the message hash somehow, but you still want to watch for all the events to not miss any contract changes made by the SP

pdp/handlers.go
magik6k: Yes, the subroot aggregation is the same as in filecoin pieces (CommP -> CommD)

The non-ffi thing is the same code as the Rust logic, reimplemented by Riba in Go.

There is one added constraint in PDP - pieces must be sorted by size in descending order - this way no inter-piece padding is needed

pdp/auth.go
```diff
@@ -0,0 +1,89 @@
...
+	"crypto/x509"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/golang-jwt/jwt/v4"
+)
+
+// verifyJWTToken extracts and verifies the JWT token from the request and returns the serviceID.
+func (p *PDPService) verifyJWTToken(r *http.Request) (string, error) {
```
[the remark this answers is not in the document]
magik6k: Yeah, see https://github.com/filecoin-project/curio/blob/54af620b619581074061af0a358baf4a3ebaa1c7/cmd/pdptool/main.go#L201

pdp/handlers.go
hannahhoward: ooh good call out good to know!

hannahhoward:
Upon further review, it's possible we have a design issue here that we need a solution for.

Please let me know if I am understanding the sequence of events correctly when a user uploads a pdp piece:

1. The piece is put in the “Stash”
2. A parked piece record is created
3. Asychronously, the piece moves from the stash to permaneant storage (Task: https://github.com/filecoin-project/curio/blob/feat/pdp/tasks/piece/task_park_piece.go)
4. After that, again asynchronously, the notify task is called, which hits the notify URL. (Task: https://github.com/filecoin-project/curio/blob/feat/pdp/tasks/pdp/notify_task.go)

Why this matters: a basic gaurantee of all hot storage, which storacha tries to offer, is read-on-write. The basic expectation is from the users point of view is that when an upload finishes, it should be retrievable.

One thing that is not clear I'd like to clarify: are steps 3 & 4 likely to run in a couple seconds or much longer?

Fortunately our upload process includes a confirmation step AFTER the upload, and what we can do is essentially block the confirmation until step 4 completes. But that only works if it's a short period of time before the piece is fully retrievable. If not, we really need to find a way to shorten the process.

hannahhoward:
Follow-up: a pull based approach for checking for piece cid -- https://github.com/filecoin-project/curio/pull/304

cmd/pdptool/main.go
```diff
@@ -0,0 +1,1068 @@
...
+		os.Exit(1)
+	}
+}
+
+var authCreateServiceSecretCmd = &cli.Command{
+	Name:  "create-service-secret",
+	Usage: "Generate a new service secret and public key",
+	Action: func(cctx *cli.Context) error {
+		// Generate an ECDSA private key
+		privKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.New(rand.NewSource(time.Now().UnixNano())))
```
magik6k, line 67: Linter probably sees this, but math rand here

tasks/message/watch_eth.go
```diff
@@ -0,0 +1,191 @@
...
+	if confirmedBlockNumber.Sign() < 0 {
+		// Not enough blocks yet
+		return
+	}
+
+	machineID := mw.ht.ResourcesAvailable().MachineID
+
+	// Assign pending transactions with null owner to ourselves
+	{
+		n, err := mw.db.Exec(ctx, `UPDATE message_waits_eth SET waiter_machine_id = $1 WHERE waiter_machine_id IS NULL AND tx_status = 'pending'`, machineID)
```
magik6k, line 77: Could use an index on this table like https://github.com/filecoin-project/curio/pull/392/files

tasks/pdp/task_prove.go
```diff
@@ -0,0 +1,758 @@
...
+				proofStr += ", "
+			}
+			proofStr += "\"0x"
+			proofStr += hex.EncodeToString(proof[:])
+			proofStr += "\""
+		}
+
+		proofStr += "] ] ]"
+
+		log.Infof("PDP Prove Task: proofSetID: %d, taskID: %d, proofs: %s", proofSetID, taskID, proofStr)
```
magik6k, line 228: Should drop the verbosity here

tasks/pdp/task_prove.go
```diff
@@ -0,0 +1,758 @@
...
+		if errors.Is(err, sql.ErrNoRows) {
+			return common.Address{}, errors.New("no sender address with role 'pdp' found")
+		}
+		return common.Address{}, err
+	}
+	address := common.HexToAddress(addressStr)
+	return address, nil
+}
+
+func (p *ProveTask) cleanupDeletedRoots(ctx context.Context, proofSetID int64, pdpVerifier *contract.PDPVerifier) error {
```
magik6k, line 639: There is also the pdp_piecerefs table, and I'm not entirely sure what's the best way to approach this because:
* On upload the piece is only referred in pdp_piecerefs (+the piecepark tables)
* Only after upload is complete and data is verified the PDP service can add the piece to a proofset
* Technically one piece can be in multiple proofsets / roots (as a subroot)
* So probably:
  * Add a flag to pdp_piecerefs which is optionally settable through the delete api which tells curio to drop the piece reference when all pdp_proofset_roots are gone
  * Possibly add an endpoint which lists all pdp_proofset_roots for a given piece (so the service can know why a given piece may be "locked")

magik6k:
Did a deeper review pass with focus on whether this will impact other curio functionality - looks like this should be safe to merge, generally most invasive changes are in piece_park, but default behaviors don't change there, and in retrieval paths, but there PDP is plugged in as a fallback, so it shouldn't impact normal operation.

tasks/pdp/task_prove.go
```diff
@@ -0,0 +1,772 @@
...
+					return false, xerrors.Errorf("failed to select proof sets: %w", err)
+				}
+
+				if len(proofSets) == 0 {
+					// No proof sets to process
+					return false, nil
+				}
+
+				// Determine if there might be more proof sets to process
+				more = len(proofSets) == 2
```
rvagg, line 108:
```suggestion
				more = len(proofSets) > 1
```
?

deps/deps.go
```diff
@@ -254,6 +258,21 @@ func (deps *Deps) PopulateRemainingDeps(ctx context.Context, cctx *cli.Context,
...
+		deps.EthClient = lazy.MakeLazy[*ethclient.Client](func() (*ethclient.Client, error) {
+			// todo: this is a hack, just use the lotus chain api client above
+			switch build.BuildType {
+			case build.BuildCalibnet:
+				return ethclient.Dial("https://api.calibration.node.glif.io/rpc/v1")
+			case build.BuildMainnet:
+				return ethclient.Dial("https://api.node.glif.io/rpc/v1")
+			default:
+				panic("fevm rpc url unknown for this network")
+			}
```
magik6k, line 271: Non-blocking: Still definitely a hack

magik6k (commented):
Generally looks good to merge to main now.

Didn't review in much detail, this PR is just waay to big to meaningfully review - but at least I'm reasonably confident it won't break existing functionality.

deps/deps.go
LexLuthr: We can use the full node for this instead of depending on the glif nodes.

web/static/ux/curio-ux.mjs
```diff
@@ -219,6 +219,12 @@ class CurioUX extends LitElement {
                     <span>IPNI</span>
                 </a>
               </li>
+              <li>
+              <a href="/pages/pdp/" class="nav-link text-white ${active=='/pages/pdp/'? 'active':''}">
+                  <svg class="bi me-2" width="16" height="16"><use xlink:href="#hdd"></use></svg>
```
LexLuthr, line 224:
```suggestion
                  <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-hdd" viewBox="0 0 16 16">
                      <path d="M4.5 11a.5.5 0 1 0 0-1 .5.5 0 0 0 0 1M3 10.5a.5.5 0 1 1-1 0 .5.5 0 0 1 1 0"/>
                      <path d="M16 11a2 2 0 0 1-2 2H2a2 2 0 0 1-2-2V9.51c0-.418.105-.83.305-1.197l2.472-4.531A1.5 1.5 0 0 1 4.094 3h7.812a1.5 1.5 0 0 1 1.317.782l2.472 4.53c.2.368.305.78.305 1.198zM3.655 4.26 1.592 8.043Q1.79 8 2 8h12q.21 0 .408.042L12.345 4.26a.5.5 0 0 0-.439-.26H4.094a.5.5 0 0 0-.44.26zM1 10v1a1 1 0 0 0 1 1h12a1 1 0 0 0 1-1v-1a1 1 0 0 0-1-1H2a1 1 0 0 0-1 1"/>
                   </svg>
```

LexLuthr (commented):
We are also missing a `make` command for `pdptool`

cmd/pdptool/pdp_cleanup.sql
LexLuthr, line 1: What is the use case of this file? I don't see it being used anywhere.

tasks/pdp/task_prove.go
magik6k: Select above has `LIMIT 2`; Doing `> 1` won't hurt though.

LexLuthr (approved):
Approving without review for now.

magik6k:
We test in prod

LETS GOO :rocket:

Issue #243, opened 2024-10-04
Multiple Nodes support in Curio


amughal:
### Checklist

- [X] This is **not** a question or a support request. If you have any Curio related questions, please ask in the [discussion forum](https://github.com/filecoin-project/curio/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/curio/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] This is **not** an enhancement request. If it is, please file a [improvement suggestion](https://github.com/filecoin-project/curio/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Fenhancement&template=enhancement.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/curio/issues) and the [discussion forum](https://github.com/filecoin-project/curio/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/curio/releases), or the most recent RC(release candidate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.
- [X] I did not make any code changes to curio.

### Curio component

- [ ] PoRep Pipeline
- [ ] Snap Deals
- [ ] Market
- [ ] IPNI provider
- [X] Curio UI
- [X] Other

### Curio Version

```text
curio version 1.23.1+mainnet+git_98dff52_2024-09-27T15:17:56-05:00
```

### Describe the Bug

Multiple nodes can be configured in the Curio configuration for Chain messaging. However if one node goes down, Curio does not fall back to the other Node.

### Logging Information

```text
Shown and discussed with Mayank.
This is the internal logic that needs to be fixed.
```

### Repo Steps

1. Run '...'
2. Do '...'
3. See error '...'
...
None

Issue #200, opened 2024-09-11
Chain Node failover not working


johansealstorage:
If I have 2 Lotus Full Chain Nodes and add both to Curio, there are a few issues: 

1. Curio does not always show both Chain Nodes on the Overview screen
2. Curio does not failover to another Chain Node if the currently used Chain Node fails (switched off or loose sync)
3. Curio fails to start up if the last Chain Node it was connected to, is not accessable - even if there are multiple other Chain nodes to try. 

My base config contains the api info for two Chain Nodes.
If I stop the Chain Node that curio is using, curio just shows the failed chain connectivity and never kicks over to using the other Chain Node.
I have tested both Chain Nodes from 2 different boosts, and both are accessible. Also using one at a time in Curio, they are both accessible. 

![image_720](https://github.com/user-attachments/assets/a6e94246-95cf-49ff-af58-cc5dd35c11d4)
![image_480](https://github.com/user-attachments/assets/c49fb929-2c65-43e6-82cb-72f2486ebe8e)

I tried to restart the node without the 10.4.10.201 Full Chain Node and it fails with this:
```
2024-09-10T21:24:10.787Z        WARN    curio/deps      deps/apiinfo.go:51      Not able to establish connection to node with addr: ws://10.4.10.201:9001/rpc/v1, Reason: cannot dial address ws://10.4.10.201:9001/rpc/v1 for dial tcp 10.4.10.201:9001: connect: no route to host: dial tcp 10.4.10.201:9001: connect: no route to host
ERROR: Not able to establish connection to more than a single node

```

Then, I restarted that 10.4.10.201 Full Chain Node and restarted curio node.
Still only 10.4.10.201 showed in the Chain Connectivity.
Then I hit the little "Move Up" arrow in the base layer config to move 10.4.9.201 to Item 1.
Now it shows 2 Chain Nodes:
![image_480](https://github.com/user-attachments/assets/05f580f3-f8d2-4c43-bfc4-8e4620788e9b)

Interesting.
If I move 10.4.10.201 into "Item 1", then curio only shows one Chain Connectivity.
If I move 10.4.9.201 into "Item 1", then curio shows both Chain Nodes.

![image_480](https://github.com/user-attachments/assets/7120276c-a8e0-4edc-a5dc-cf2630fc0aa2)

And even when both Chain Nodes shows in Chain connectivity, as soon as I stop the Chain node that is actively being used, Curio does not switch over to the other Chain Node.  (even after 10 minutes).
[...]

PR #249, opened 2024-10-04
fix: Improved multinode proxy

LexLuthr:
Fixes https://github.com/filecoin-project/curio/issues/243 and https://github.com/filecoin-project/curio/issues/200
This PR bring in the changes from https://github.com/filecoin-project/lotus/pull/11470 along with addressing some comments. 

This PR also removes the forced check to ensure at least 2 chain nodes are reachable. We don't require this check as there is no network segregation or split-brain issues here. Any node will work as long as it is in sync. This check also has some unintended affects of rendering Curio cluster useless if a node goes down out of 2.

### DIFF
--- a/deps/apiinfo.go
+++ b/deps/apiinfo.go
@@ -10,6 +10,7 @@ import (
 	"sync"
 	"time"
 
+	logging "github.com/ipfs/go-log/v2"
 	"github.com/urfave/cli/v2"
 	"golang.org/x/xerrors"
 
@@ -21,6 +22,8 @@ import (
 	"github.com/filecoin-project/lotus/chain/types"
 	cliutil "github.com/filecoin-project/lotus/cli/util"
 )
+
+var clog = logging.Logger("curio/chain")
 
 func GetFullNodeAPIV1Curio(ctx *cli.Context, ainfoCfg []string) (api.Chain, jsonrpc.ClientCloser, error) {
 	if tn, ok := ctx.App.Metadata["testnode-full"]; ok {
@@ -52,7 +55,7 @@ func GetFullNodeAPIV1Curio(ctx *cli.Context, ainfoCfg []string) (api.Chain, json
 	for _, head := range httpHeads {
 		v1api, closer, err := newChainNodeRPCV1(ctx.Context, head.addr, head.header)
 		if err != nil {
-			log.Warnf("Not able to establish connection to node with addr: %s, Reason: %s", head.addr, err.Error())
+			clog.Warnf("Not able to establish connection to node with addr: %s, Reason: %s", head.addr, err.Error())
 			continue
 		}
 		fullNodes = append(fullNodes, v1api)
@@ -156,7 +159,7 @@ func FullNodeProxy[T api.Chain](ins []T, outstr *api.ChainStruct) {
 						unhealthyProviders[i] = true
 						healthyLk.Unlock()
 
-						log.Errorw("rpc check chain head call failed", "fail_type", "rpc_error", "provider", i, "error", err)
+						clog.Debugw("rpc check chain head call failed", "fail_type", "rpc_error", "provider", i, "error", err)
 						return
 					}
 
@@ -170,7 +173,7 @@ func FullNodeProxy[T api.Chain](ins []T, outstr *api.ChainStruct) {
 						// if we're behind the best tipset, mark as unhealthy
 						unhealthyProviders[i] = ch.Height() < bestKnownTipset.Height()-maxBehindBestHealthy
 						if unhealthyProviders[i] {
-							log.Errorw("rpc check chain head call failed", "fail_type", "behind_best", "provider", i, "height", ch.Height(), "best_height", bestKnownTipset.Height())
+							clog.Debugw("rpc check chain head call failed", "fail_type", "behind_best", "provider", i, "height", ch.Height(), "best_height", bestKnownTipset.Height())
 						}
 					}
 					healthyLk.Unlock()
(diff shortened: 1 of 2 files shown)


magik6k (commented):
Would be really good to have a way to expose backing node health in the webui somehow, even just the ones on the node serving the webui

LexLuthr:
> Would be really good to have a way to expose backing node health in the webui somehow, even just the ones on the node serving the webui

That would be a good idea. Let me do that.

deps/apiinfo.go
```diff
@@ -96,54 +98,152 @@ func newChainNodeRPCV1(ctx context.Context, addr string, requestHeader http.Head
 	return &res, closer, err
 }
 
+const initialBackoff = time.Second
+const maxRetryAttempts = 5
+const maxBehinhBestHealthy = 1
```
magik6k, line 103:
```suggestion
const maxBehindBestHealthy = 1
```

deps/apiinfo.go
```diff
@@ -155,3 +255,32 @@ func FullNodeProxy[T api.Chain](ins []T, outstr *api.ChainStruct) {
...
+
+func ErrorIsIn(err error, errorTypes []error) bool {
+	for _, etype := range errorTypes {
+		tmp := reflect.New(reflect.PointerTo(reflect.ValueOf(etype).Elem().Type())).Interface()
+		if errors.As(err, &tmp) {
+			return true
+		}
+	}
+	return false
+}
```
magik6k, line 286: I'm not 100% convinced this actually works correctly, we definitely should test this PR somehow to make sure this / the rest of the stuff here actually works.

Can be as simple as starting with 2 lotus nodes and killing one while some RPC is ongoing (tricky, but possible with some e.g. message send where gas estimation takes a while)

magik6k (approved):
The ErrorIsIn is a little sketchy, though I remember writing / reviewing it very deeply, so it's prooobably fine, but would be amazing to confirm.

deps/apiinfo.go
strahe: I am just a bit curious, why was the Lotus cluster functionality deprecated? Even if using Curio to replace lotus-miner, the Lotus cluster still seems useful.

magik6k: It pulled in massive unaudited raft dependencies, wasn't reliable, made signing endpoints available over libp2p with just simple allowlist filter to peers in the cluster, which was later discovered to not provide enough security because raft was never meant to run in an unsecured network context.

Basically it was a giant unmaintained hack which later on started to also produce a long list of security issues so it just didn't seem like a good idea to keep it around.

johansealstorage:
The multi node switching is working now. 
Testing with 2 nodes: 
1. Switched off one node, curio switches to using the other healthy node.
2. Start up the node that was switched off, and then switched off the second node. Curio continues to work fine by picking up the healthy node. 
3. Out of sync node, curio still operates with the one healthy node.  

Testing with 3 nodes showed the same results as with 2 nodes.  As long as there is at least one healthy node, curio continues to work fine. 


Some log where 2 of the 3 nodes were stopped shows that curio still continued working after successfully switching to the last healthy chain node.

```
  |   | 2024-10-15 11:28:51.341 | 2024-10-15T09:28:51.139Z	WARN	webrpc	webrpc/sync_state.go:140	error creating jsonrpc client: cannot dial address ws://10.4.8.201:9001/rpc/v1 for dial tcp 10.4.8.201:9001: connect: no route to host: dial tcp 10.4.8.201:9001: connect: no route to host |  
  |   | 2024-10-15 11:28:52.093 | 2024-10-15T09:28:52.012Z	INFO	harmonytask	harmonytask/task_type_handler.go:182	Beginning work on Task	{"id": 3213004, "from": "poller", "name": "WinInclCheck"} |  
  |   | 2024-10-15 11:28:52.093 | 2024-10-15T09:28:52.034Z	WARN	webrpc	webrpc/sync_state.go:140	error creating jsonrpc client: cannot dial address ws://10.4.10.201:9001/rpc/v1 for dial tcp 10.4.10.201:9001: connect: no route to host: dial tcp 10.4.10.201:9001: connect: no route to host |  
  |   | 2024-10-15 11:28:52.117 | 2024-10-15T09:28:51.870Z	ERROR	curio/chain	deps/apiinfo.go:277	Failed after 5 attempts, last error: RPCConnectionError |  
  |   | 2024-10-15 11:28:52.117 | 2024-10-15T09:28:51.870Z	ERROR	harmonytask	harmonytask/task_type_handler.go:229	Do() returned error	{"type": "WinInclCheck", "id": "3213004", "error": "RPCConnectionError"} |  
  |   | 2024-10-15 11:28:52.117 | 2024-10-15T09:28:52.013Z	INFO	harmonytask	harmonytask/task_type_handler.go:157	did not accept task	{"task_id": "3213004", "reason": "already Taken", "name": "WinInclCheck"} |  
  |   | 2024-10-15 11:28:52.344 | 2024-10-15T09:28:52.138Z	DEBUG	harmonytask	harmonytask/harmonytask.go:360	skipped scheduling WinInclCheck type tasks on due to Did not accept WinInclCheck task: at max already |  
  |   | 2024-10-15 11:29:01.366 | 2024-10-15T09:29:01.256Z	INFO	harmonytask	harmonytask/task_type_handler.go:157	did not accept task	{"task_id": "3213805", "reason": "already Taken", "name": "WinPost"}
```

johansealstorage:
Tested with the latest UI fix and all Chain nodes are now showing as expected on the UI.  The health of each of then also follows the out of sync and the reachability.  Curio switches nicely between the available ones. 

One situation that might be confusing is if a new API is added to the base layer.  The UI starts tracking its health, but curio in the background will not use that node until curio is restarted.   That might be a bit of an out of sync situation of what the user sees, vs what curio is actually using.   After restart of curio, everything comes into sync again and the UI agrees with what curio uses.

deps/apiinfo.go
```diff
@@ -96,54 +101,152 @@ func newChainNodeRPCV1(ctx context.Context, addr string, requestHeader http.Head
...
+				preferredProvider := new(int)
+				*preferredProvider = nextHealthyProvider(0)
+				if *preferredProvider == preferredAllBad {
+					// select at random, retry will do it's best
+					*preferredProvider = rand.Intn(providerCount)
+				}
 
 				// for calls that need to be performed on the same node
 				// primarily for miner when calling create block and submit block subsequently
 				key := contextKey("retry-node")
```
magik6k, line 231: OOooh I forgot about this, it will make F3 really easy to implement

magik6k (approved):

PR #252, opened 2024-10-04
fix: remove wdPost pre-checks, unused config

LexLuthr:


magik6k (approved):
Seems reasonable, I'm decently confident that this is fine. Will make sure to run on my nodes

PR #254, opened 2024-10-04
fix: libp2p migration and generation

LexLuthr:


cmd/curio/market.go
```diff
@@ -335,3 +339,108 @@ var marketMoveToEscrowCmd = &cli.Command{
...
+		if err != nil {
+			return xerrors.Errorf("failed to generate a new key: %w", err)
+		}
+
+		mid, err := address.IDFromAddress(act)
+		if err != nil {
+			return err
+		}
+
+		_, err = dep.DB.Exec(ctx, `INSERT INTO libp2p (sp_id, priv_key) VALUES ($1, $2) ON CONFLICT(sp_id) DO NOTHING`, mid, pk)
```
magik6k, line 430: Maybe check affected rows, and if zero say that no key was created?

magik6k (approved):

Issue #185, opened 2024-08-31
Show sectors that are queued to be sealed as queued not as SDR


lanzafame:
Currently, sectors that have been staged (by staged I mean the system doesn't currently have enough resources to seal the sectors) to be sealed are shown as being in the SDR task stage.

Ideally, they are shown as Queued. This would fix the timing value associated with the SDR stage as it includes the time the sector was waiting to start SDR as well.

This would be reflected in both the Cluster Tasks list on the Overview page as well as in the PoRep pipeline diagram in the `State Sealing` box.

![Screenshot 2024-08-30 at 13 50 51](https://github.com/user-attachments/assets/ec19c821-de5a-48ed-99ac-e39ad87e09a1)

PR #260, opened 2024-10-07
fix: show task waiting in PoRep pipeline

LexLuthr:
Fixes https://github.com/filecoin-project/curio/issues/185

Table sorted with Running tasks and then pending tasks. Both of them are then ordered by Posted Time.
Screenshot 2024-10-09 at 4 58 14 PM

Different color for waiting tasks
Screenshot 2024-10-09 at 5 21 56 PM

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -257,7 +257,7 @@ build_lotus?=0
 curio_docker_user?=curio
 curio_base_image=$(curio_docker_user)/curio-all-in-one:latest-debug
 ffi_from_source?=0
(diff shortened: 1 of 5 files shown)


LexLuthr:
I think a single query might be better.

Screenshot 2024-10-09 at 7 14 47 PM

magik6k (approved):
Would be nice to also get for Snap (..and also somehow maybe unify some of the pipeline state rendering code)

PR #276, opened 2024-10-14
fix: do not return early from message watcher

LexLuthr:
Early return can choke the cluster if first message results in the error. We should should log the error and try the next one.


magik6k (approved):
Oh gee yes that's very broken

PR #277, opened 2024-10-14
fix: ipni: Correctly handle URL formats

magik6k:
Basically use ipni `maurl` package in all related places to correctly convert to/from normal http(s) URLs


market/ipni/ipni-provider/ipni-provider.go
```diff
@@ -465,132 +482,128 @@ func (p *Provider) reconstructChunkFromDB(ctx context.Context, chunk, piece, fir
 	return b.Bytes(), nil
 }
 
+func (p *Provider) handleGetHead(w http.ResponseWriter, r *http.Request) {
+	log.Infow("Received IPNI request", "path", r.URL.Path)
+
+	providerID := chi.URLParam(r, "providerId")
```
LexLuthr, line 488: ~~I don't think IPNI sends providerId as param in request. If it does, can we reference the source in a comment?~~
Ignore.

PR #280, opened 2024-10-15
Feat/libp2p fixes

magik6k:
* Single libp2p per cluster, runs on one leader
* Curio HTTP servers redirect `/` websocket connections to the leader libp2p endpoint
* IPNI now uses a separate table to manage per-SP PeerIDs, never used in libp2p instances
  * Doesn't matter because Graphsync is not a thing anymore

### DIFF
--- a/cmd/curio/market.go
+++ b/cmd/curio/market.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"bufio"
-	"crypto/rand"
 	"encoding/json"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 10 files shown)


LexLuthr (approved):
Looks good! 
I just have a couple of question:
1. I am assuming that libp2p will not try to punch through the firewall in this case. Correct?
2. Flow would be as below:

- Listen on a random port for all IPs
- Announce that you are running on HTTP announce address with wss
- Update the same on chain
- When a client tries to connect, they actually get connected to our HTTP server which redirect it to libp2p listen address. This libp2p tells clients that it is HTTP announce address and proceeds with the deal.

magik6k:
1. We basically never try to connect outside, and we assume that the http server is already accessible on the outside
2. Yes, there is one libp2p node in the internal network that only really can accept connections from other internal nodes, then the HTTP server acts as a reverse proxy. The announce address we pass into libp2p is only used by the dht (which we don't use anyways) and maybe for autonat checks (which don't even matter and I'm not entirely sure if those are even enabled)

PR #282, opened 2024-10-16
Toolbox/precommit stuck

lanzafame:


cmd/toolbox/main.go
```diff
@@ -0,0 +1,297 @@
...
+	if err != nil {
+		return address.Undef, fmt.Errorf("parsing address: %w", err)
+	}
+	return addr, nil
+}
+
+var precommitStuckCmd = &cli.Command{
+	Name:  "precommit-stuck",
+	Usage: "Perform db operations to fix issues with precommit messages getting stuck",
+	Action: func(cctx *cli.Context) error {
```
LexLuthr, line 93: We should take a user input about which messages to check for. Doing a blanket try on all would be an issue. In fact, we should get a list of all pending ones. Then check if any of them can be found on filfox (already executed), check if FullNode is part the execution tipset. If it is then we show these messages to user and let me choose to fix the problem for specific ones they want. Or maybe have a flag like `--fix-all` then in that case fix them all.

LexLuthr (changes requested):
Let's make some changes to make this into a generic command. It would be very useful for everyone in multiple scenarios.

cmd/toolbox/main.go
```diff
@@ -0,0 +1,297 @@
...
+
+var log = logging.Logger("toolbox")
+
+func main() {
+	local := []*cli.Command{
+		precommitStuckCmd,
+	}
+
+	app := &cli.App{
+		Name:                 "toolbox",
```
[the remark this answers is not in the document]
magik6k: +1

LexLuthr:
Once you are done with changes, don't forget to run `make gen` command to generate/fix everything.

cmd/toolbox/main.go
lanzafame: I agree we should create a pending(stuck)-msg cmd that lists msgs that are stuck. What changes would I need to make to db query to get the different stuck msgs? On the fixing other message type, can you give me an example of what other message/methods may get stuck in the same way as this? Currently, the fixing seems dependent on the precommit msg cid being the thing we are solving around.

LexLuthr:
@lanzafame Hey! Any progress on the remaining changes?

LexLuthr:
https://github.com/filecoin-project/curio/pull/487 supersedes this.

PR #286, opened 2024-10-17
feat: NV24 upgrade

LexLuthr:


api/api_chain.go
```diff
@@ -67,7 +67,7 @@ type CurioChainRPC interface {
 	StateSectorPreCommitInfo(context.Context, address.Address, abi.SectorNumber, types.TipSetKey) (*miner.SectorPreCommitOnChainInfo, error)
 	StateSectorGetInfo(ctx context.Context, maddr address.Address, sectorNumber abi.SectorNumber, tsk types.TipSetKey) (*miner.SectorOnChainInfo, error)
 	StateMinerPreCommitDepositForPower(context.Context, address.Address, miner.SectorPreCommitInfo, types.TipSetKey) (big.Int, error)
-	StateMinerInitialPledgeCollateral(context.Context, address.Address, miner.SectorPreCommitInfo, types.TipSetKey) (big.Int, error)
+	StateMinerInitialPledgeForSector(ctx context.Context, sectorDuration abi.ChainEpoch, sectorSize abi.SectorSize, verifiedSize uint64, tsk types.TipSetKey) (types.BigInt, error)
```
magik6k, line 70: I presume this checks the network version internally based on the tipset?

tasks/seal/task_submit_commit.go
```diff
@@ -248,7 +251,12 @@ func (s *SubmitCommitTask) Do(taskID harmonytask.TaskID, stillOwned func() bool)
 		return false, xerrors.Errorf("could not serialize commit params: %w", err)
 	}
 
-	collateral, err := s.api.StateMinerInitialPledgeCollateral(ctx, maddr, pci.Info, ts.Key())
```
magik6k, line 251: Waait how did that work with filplus deals? It couldn't be correct, right?

magik6k (approved):
Looks good so far, also confirmed that supraseal build still works with the new ffi

tasks/seal/task_submit_commit.go
LexLuthr: I don't think it did. We were sending incorrect collateral AFAIK.

api/api_chain.go
LexLuthr: I don't see a NV check. Why do we need it?

PR #288, opened 2024-10-17
controlled indexing after migration

LexLuthr:
This PR makes the following changes to the market migration.

1. The indexing tasks are no longer created directly.
2. All info is captured in a table `market_mk12_deal_pipeline_migration`
3. A migration function is called every 30 seconds by market poller to migrate data from this table to mk12_pipeline table which in turn creates indexing tasks along with IPNI if required.
4. The function has following logic:
    - Step 1: Check if the migration table has entries
    - Step 2: Count entries ready for indexing in the pipeline table
    - Step 3: Count pending IPNI tasks
    - Calculate how many entries we need to reach 16
    - Calculate how many entries we can move without exceeding 16 pending IPNI tasks
    - Limit by the number of entries available in the migration table
    - Move entries from the migration table to the pipeline table

Relevant changes in Boost will be done.

### DIFF
--- a/harmony/harmonydb/sql/20241017-market-mig-indexing.sql
+++ b/harmony/harmonydb/sql/20241017-market-mig-indexing.sql
@@ -26,7 +26,7 @@ DECLARE
     moved_rows RECORD;
 BEGIN
     -- Step 1: Check if the migration table has entries
-    SELECT COUNT(*) INTO cnt FROM market_mk12_deal_pipeline_migration;
+    SELECT COUNT(*) INTO cnt FROM market_mk12_deal_pipeline_migration LIMIT 16;
     IF cnt = 0 THEN
             RETURN;
     END IF;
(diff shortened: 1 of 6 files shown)


tasks/indexing/task_ipni.go
```diff
@@ -94,6 +97,10 @@ func (I *IPNITask) Do(taskID harmonytask.TaskID, stillOwned func() bool) (done b
 
 	task := tasks[0]
 
+	if task.Complete {
+		return true, nil
+	}
```
magik6k, line 102: I don't quite understand what the purpose of this is. Just being careful around retry?

tasks/indexing/task_ipni.go
```diff
@@ -306,6 +313,7 @@ func (I *IPNITask) TypeDetails() harmonytask.TaskTypeDetails {
 		IAmBored: passcall.Every(5*time.Minute, func(taskFunc harmonytask.AddTaskFunc) error {
 			return I.schedule(context.Background(), taskFunc)
 		}),
+		Max: taskhelp.Max(6),
```
magik6k, line 316: Should be a config, probably with a shared max (like sdr/unseal) with the indexing task (both hammer disk reads quite a lot)

harmony/harmonydb/sql/20241017-market-mig-indexing.sql
```diff
@@ -0,0 +1,88 @@
...
+    -- Counts for existing entries and tasks
+    num_entries_ready_for_indexing INTEGER;
+    num_pending_ipni_tasks INTEGER;
+    num_entries_needed INTEGER;
+    num_entries_to_move INTEGER;
+    cnt INTEGER;
+    moved_rows RECORD;
+BEGIN
+    -- Step 1: Check if the migration table has entries
+    SELECT COUNT(*) INTO cnt FROM market_mk12_deal_pipeline_migration;
```
magik6k, line 29: Probably LIMIT here and in other count selects

magik6k (approved):
This looks basically good, couple small nits, but this already should work

tasks/indexing/task_ipni.go
LexLuthr: Yes. Being extra careful.

tasks/indexing/task_ipni.go
LexLuthr: That makes sense. I will create a common one.

harmony/harmonydb/sql/20241017-market-mig-indexing.sql
LexLuthr: Done with 16 Limit as we don't need more anyway.

PR #289, opened 2024-10-17
Make market deals work

magik6k:
This PR adds a really basic market UI + fixes enough bugs to have deals get into sectors. I wasn't able to get a deal fully into a sector yet because my calibnet miner is a hopeless trashfire, but with some new CC sectors should be able to get this fully tested.

Deal UI
2024-10-17-231028_1783x539_scrot

Storage ask modal
2024-10-17-231037_1822x1768_scrot


LexLuthr (approved):

web/static/pages/market/deal-pipelines.mjs
```diff
@@ -0,0 +1,152 @@
...
+        }
+    }
+
+    getDealStatus(deal) {
+        if (deal.complete) {
+            return 'Complete';
+        } else if (deal.sector) {
+            return 'Sealed';
+        } else if (deal.after_find_deal) {
+            return 'Deal Found';
```
LexLuthr, line 130:
```suggestion
            return 'On Chain';
```

LexLuthr:
I think we should merge this as is. I will do a PR on top of this to change the pipeline to the same structure as PoRep pipeline with more details like sector number , indexing etc. 
I will probably also create a new page to just list the deals and some basic info about them. Most if the API side is already done for this. Just need to paint the UI.

PR #292, opened 2024-10-21
feat: F3

magik6k:


harmony/taskhelp/common.go
```diff
@@ -17,3 +17,14 @@ func SliceIfFound[T any](slice []T, f func(T) bool) ([]T, bool) {
...
+
+// BackgroundTask are tasks that:
+// * Always run in the background
+// * Never finish "successfully"
+func BackgroundTask(name string) string {
+	return "bg:" + name
+}
+
+func IsBackgroundTask(name string) bool {
+	return len(name) > 3 && name[:3] == "bg:"
```
Stebalien, line 29:
```suggestion
	return strings.HasPrefix(name, "bg:")
```

tasks/f3/f3_task.go
```diff
@@ -0,0 +1,285 @@
...
+
+	maddr, err := address.NewIDAddress(uint64(spID))
+	if err != nil {
+		return false, xerrors.Errorf("failed to parse miner address: %w", err)
+	}
+
+	for {
+		if !stillOwned() {
+			return false, nil
+		}
```
Stebalien, line 87:
```suggestion
	for stillOwned() {
```

tasks/f3/f3_task.go
```diff
@@ -0,0 +1,285 @@
...
+
+		var previousTicket []byte
+		err = f.db.QueryRow(ctx, "SELECT previous_ticket FROM f3_tasks WHERE task_id = $1", taskID).Scan(&previousTicket)
+		if err != nil {
+			return false, xerrors.Errorf("failed to get previous ticket: %w", err)
+		}
+
+		// Ensure that calls are made on the same node (the first call will determine the node)
+		ctx := deps.OnSingleNode(ctx)
```
Stebalien, line 96: Don't we want to re-use this between loop iterations? Note: `awaitLeaseExpiry` waits till we're half-way through the lease. The idea is that the next lease will overlap with the previous one to avoid gaps.

However, that only works if we keep hitting the same node. If we don't, `tryGetF3ParticipationTicket` will keep failing and re-trying. Which is fine... I guess. It would just be _nicer_ to try the same node over and over.

tasks/f3/f3_task.go
```diff
@@ -0,0 +1,285 @@
...
+		if !participating {
+			return false, xerrors.Errorf("failed to participate in F3: not participating")
+		}
+
+		_, err = f.db.Exec(ctx, "UPDATE f3_tasks SET previous_ticket = $1 WHERE task_id = $2", ticket, taskID)
+		if err != nil {
+			return false, xerrors.Errorf("failed to update previous ticket: %w", err)
+		}
+
+		bo := &backoff.Backoff{
```
Stebalien, line 116: Move this into `awaitLeaseExpiry`?

tasks/f3/f3_task.go
```diff
@@ -0,0 +1,285 @@
...
+		db:        db,
+		api:       api,
+		leaseTerm: ParticipationLeaseTerm,
+
+		actors: actors,
+	}
+}
+
+func (f *F3Task) Do(taskID harmonytask.TaskID, stillOwned func() bool) (done bool, err error) {
+	ctx := context.Background()
```
Stebalien, line 71: This context seems kind of useless and like it's just complicating things.

tasks/f3/f3_task.go
```diff
@@ -0,0 +1,285 @@
...
+			return false, xerrors.Errorf("failed to update previous ticket: %w", err)
+		}
+
+		bo := &backoff.Backoff{
+			Min:    1 * time.Second,
+			Max:    1 * time.Minute,
+			Factor: 1.5,
+		}
+
+		err = f.awaitLeaseExpiry(stillOwned, ctx, lease, bo)
```
Stebalien, line 122: arg ordering.

tasks/f3/f3_task.go
```diff
@@ -0,0 +1,285 @@
...
+			)
+			return lease, true, nil
+		}
+	}
+	return api.F3ParticipationLease{}, false, ctx.Err()
+}
+
+func (f *F3Task) awaitLeaseExpiry(stillOwned func() bool, ctx context.Context, lease api.F3ParticipationLease, backoff *backoff.Backoff) error {
+	renewLeaseWithin := f.leaseTerm / 2
+	for stillOwned() {
```
Stebalien, line 192: How racy is `stillOwned`? Or, really, what's the timeout on ownership? I.e., if it returns true, am I likely to lose ownership within seconds, milliseconds, minutes, etc?

Stebalien (commented):
Overall seems reasonable. My comments are mostly opinionated nits from someone who really has no say in how you write code, so feel free to ignore them and focus on the questions.

tasks/f3/f3_task.go
magik6k: Minutes, and only in cases when something was wrong with the node running the task - tasks can’t be stolen as long as the nodes owning them are pinging the database. stillOwned simply queries the db to check that the task is still assigned to us. It is by no means perfect, but it’s also there only to handle a rare case of a node loosing connectivity while still running

magik6k: (Nodes are removed from the database with their tasks unassigned after ~5 minutes of not responding)

tasks/f3/f3_task.go
magik6k: This is tricky, I may add another layer here with softer preference between the loops - OnSingleNode will fail calls if the selected backing node becomes dead iirc

tasks/f3/f3_task.go
magik6k: Still need it for all the API calls, and it's easier to just spawn one top-level one.

tasks/f3/f3_task.go
magik6k: Ah, never mind, it's actually going to handle retry correctly

tasks/f3/f3_task.go
```diff
@@ -0,0 +1,281 @@
...
+		Cost: resources.Resources{
+			Cpu: 0,
+			Gpu: 0,
+			Ram: 10 << 20,
+		},
+		MaxFailures: 1,
+	}
+}
+
+func (f *F3Task) Adder(taskFunc harmonytask.AddTaskFunc) {
```
LexLuthr, line 252: Where is this added function called from? Did I miss the scheduling part somewhere?

LexLuthr (commented):
Needs documentation and config to control participation.

tasks/f3/f3_task.go
magik6k: This is the normal adder func called by harmonytask on startup

magik6k:
> Needs documentation and config to control participation.

Not sure about config, we can add a flag matching the lotus one for disabling, but iiuc SPs are never supposed to mine without participating in F3

tasks/f3/f3_task.go
Stebalien: Should be plenty of time.

PR #295, opened 2024-10-23
fix: unsealed sector for synthetic PoRep

LexLuthr:
This PR makes the following changes:
1. Generating Unsealed is now a separate function called GenerateUnsealedSector()
2. This function is called by Synthetic Task if required and generating the unsealed sector before removing the cache.
3. Finalize calls generating unseal only if proof types is not synthetic.
4. Acquiring sector from Storage is still done by the primary function in each task as passed to GenerateUnsealedSector().

- [ ] Test in calibnet

### DIFF
--- a/lib/ffi/sdr_funcs.go
+++ b/lib/ffi/sdr_funcs.go
@@ -19,7 +19,7 @@ import (
 	"github.com/filecoin-project/curio/harmony/harmonytask"
 	"github.com/filecoin-project/curio/lib/ffiselect"
 	"github.com/filecoin-project/curio/lib/proof"
-	storiface "github.com/filecoin-project/curio/lib/storiface"
+	"github.com/filecoin-project/curio/lib/storiface"
 
 	// TODO everywhere here that we call this we should call our proxy instead.
 	ffi "github.com/filecoin-project/filecoin-ffi"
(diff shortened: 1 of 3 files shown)


lib/ffi/sdr_funcs.go
```diff
@@ -780,6 +806,11 @@ func (sb *SealCalls) SyntheticProofs(ctx context.Context, task *harmonytask.Task
 		}
 	}
 
+	// Generate unsealed copy before clearing cache
+	if err = sb.GenerateUnsealedSector(ctx, sector, fspaths, pathIDs, keepUnsealed); err != nil {
+		return xerrors.Errorf("generating unsealed sector: %w", err)
+	}
+
 	if err = ffi.ClearCache(uint64(ssize), fspaths.Cache); err != nil {
```
magik6k, line 814: Are you sure ClearCache will work when tree-d is not present? I imagine it should have no issues, but I'm actually not 100% certain, so we should test this first.

LexLuthr: This was testing on mainnet by 2 SPs.

lib/ffi/sdr_funcs.go
```diff
@@ -525,111 +525,137 @@ func changePathType(path string, newType storiface.SectorFileType) (string, erro
...
-	retryUnsealedMove:
+	return TruncateAndMoveUnsealed(tempUnsealed, sectorPaths.Unsealed, ssize)
+}
 
-		// truncate sealed file to sector size
-		if err := os.Truncate(tempUnsealed, int64(ssize)); err != nil {
-			return xerrors.Errorf("truncating unsealed file to sector size: %w", err)
-		}
+func TruncateAndMoveUnsealed(tempUnsealed, unsealed string, ssize abi.SectorSize) error {
+	// truncate sealed file to sector size
```
magik6k, line 625: Nit
```suggestion
	// truncate unsealed file to sector size
```

magik6k (approved):

phapawit-hir:
Hi there,

it works.

Issue #283, opened 2024-10-16
Curio Configuration Saving needs to warn user to restart curio to apply the changes.


johansealstorage:
### Checklist

- [X] This is **not** a question or a support request. If you have any Curio related questions, please ask in the [discussion forum](https://github.com/filecoin-project/curio/discussions).
- [X] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/curio/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [X] I **have** searched on the [issue tracker](https://github.com/filecoin-project/curio/issues) and the [discussion forum](https://github.com/filecoin-project/curio/discussions), and there is no existing related issue or discussion.
- [X] I am running the [`Latest release`](https://github.com/filecoin-project/curio/releases), or the most recent RC(release candidate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.

### Curio component

- [ ] PoRep Pipeline
- [ ] Snap Deals
- [ ] Market
- [ ] IPNI provider
- [X] Curio UI
- [ ] Other

### Improvement Suggestion

Some configuration layer changes require curio nodes to be restarted in order to pick up the configuration change. 
An example is added or removing ChainApiInfo.  Curio will only start using (or losing) the said API Info once the curio node is restarted.  This is because some configuration changes are only applied on curio startup.
The Curio UI on the other hand will show the config change immediately without a restart.  
This leads to an out of sync situation where the user might see the Chain Connection info on the UI and think that curio is actually using it, but it is not due to the restart requirement of such a config change. 

A possible enhancement would be to warn the user when the Configuration is changed, that the change will require a restart. A message bar or constant reminder might also help to keep reminding the user of the required restart.

The ideal situation would be some kind of way to apply the configuration without a restart by forcing curio to re-read and apply configurations as soon as it gets to a point in its tasks where it is logical to do a configuration refresh.

strahe:
As far as I know, currently, all configurations require a restart to take effect. But as you mentioned, if some configurations could take effect in real-time and others could remind the user that a restart is needed, this feature would be very useful.

LexLuthr:
Let me clarify, no config takes affect before restart. Only UI chain connectivity list is exception here. "UI" is the keyword here. Cluster will not be aware of this change. 

The message would be a nice reminder to SPs like "don't forget to restart the node using this layer".

PR #297, opened 2024-10-23
fix: config save cookie

LexLuthr:
Fixes https://github.com/filecoin-project/curio/issues/283
The configuration save cookie has not been functioning for a while. This PR fixes it. 

Screenshot 2024-10-23 at 7 25 38 PM

### DIFF
--- a/web/static/pages/config/list/config-list.mjs
+++ b/web/static/pages/config/list/config-list.mjs
@@ -68,7 +68,6 @@ class ConfigList extends StyledLitElement {
   }
 
   render() {
-    console.log('Read cookie:', this.message);
     return html`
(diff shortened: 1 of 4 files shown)


web/static/pages/config/list/config-list.mjs
```diff
@@ -68,6 +68,7 @@ class ConfigList extends StyledLitElement {
   }
 
   render() {
+    console.log('Read cookie:', this.message);
```
snadrus, line 71: this wont do anything

snadrus: you probably want getCookieMessage() in a service that this can use & show.

snadrus (changes requested):

web/static/pages/config/list/config-list.mjs
LexLuthr: Leftover from debugging. Removed it.

PR #298, opened 2024-10-24
fix(f3): handle failure to receive manifest from the client

Stebalien:
If we keep going here, we'll panic. See (also includes a regression test):

https://github.com/filecoin-project/lotus/pull/12634


magik6k (approved):
Fortunately that's fine - task handlers catch panics and should correctly retry

PR #302, opened 2024-10-25
f3: Update participate loop to always participate

magik6k:
Ports the fix from https://github.com/filecoin-project/lotus/pull/12640 to curio.

### DIFF
--- a/tasks/f3/f3_task.go
+++ b/tasks/f3/f3_task.go
@@ -6,11 +6,9 @@ import (
 	"time"
 
 	logging "github.com/ipfs/go-log/v2"
-	"github.com/jpillora/backoff"
 	"golang.org/x/xerrors"
 
 	"github.com/filecoin-project/go-address"
-	"github.com/filecoin-project/go-f3/gpbft"
 	"github.com/filecoin-project/go-f3/manifest"
 
 	"github.com/filecoin-project/curio/deps"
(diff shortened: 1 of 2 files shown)


tasks/f3/f3_task.go
```diff
@@ -136,101 +128,97 @@ func (f *F3Task) tryGetF3ParticipationTicket(ctx context.Context, stillOwned fun
...
+			}
+		}
+
+		// Wait until we think we may need to renew the lease
+		waitTime := time.Duration(lease.ValidityTerm-renewLeaseWithin) * manifest.CatchUpAlignment
+		if waitTime <= 0 {
+			waitTime = 100 * time.Millisecond
+		}
+		log.Debugf("F3 participation lease is valid for further %d instances. Re-checking after %s.", lease.ValidityTerm, waitTime)
+		time.Sleep(waitTime)
```
Stebalien, line 198: Note: if you don't want to bother fetching the manifest, you can also just wait a second (or 5) here between polls. We're doing it this way in lotus because waiting a fixed amount of time makes testing difficult.

Stebalien (approved):

tasks/f3/f3_task.go
magik6k: Ah nice, simple good

tasks/f3/f3_task.go
```diff
@@ -136,99 +128,56 @@ func (f *F3Task) tryGetF3ParticipationTicket(ctx context.Context, stillOwned fun
...
-				waitTime = 100 * time.Millisecond
-			}
-			log.Debugf("F3 participation lease is valid for further %d instances. Re-checking after %s.", remainingInstanceLease, waitTime)
-			time.Sleep(waitTime)
 		}
+
+		log.Debugf("F3 participation lease is valid for further %d instances.", lease.ValidityTerm)
+		time.Sleep(time.Second * 5)
 	}
 	return ctx.Err()
```
Stebalien, line 182: Not new but... using `stillOwned()` and the context like this means you'll return `nil` when exiting instead of an error.

Kubuxu:
@magik6k, is there anything we can do to land this? We are quite sure it froze up Calibnet's F3 after people updated to RC3 and restarted their daemons.

tasks/f3/f3_task.go
magik6k: That's actually fine, the f3 task scheduler will see that there is no task running for that SP ID, and will just spawn a new one

LexLuthr (approved):

PR #304, opened 2024-10-27
feat(pdp): add endpoint to poll for completed upload

hannahhoward:
# Goals

This enables a service using PDP to query based on its upload parameters and poll for upload completion. The primary requirement is this should only return a piece cid if there is a retrievable piece for the given parameters.

# Implementation

- Add an provides an endpoint to look up a piece cid for a given set of upload parameters. It will only respond if the piece exists. It will also mark if the piece is retriavable and whether it's ready for PDP aggregation. This largely mimics some initial logic from POST /pdp/piece that determines if it's 201 created (new upload) or 200 (existing upload)
- Add docs for new endpoint
- Correct one comment that appears to refer to the wrong table.

# For discussion

Currently, this will succeed as soon as the piece finishes uploading, and it will seperately note if the piece is retrievable and ready for aggregation. A previous version would 404 if the piece was not retrievable.

### DIFF
--- a/pdp/handlers.go
+++ b/pdp/handlers.go
@@ -104,6 +104,9 @@ func Routes(r *chi.Mux, p *PDPService) {
 	// POST /pdp/piece
 	r.Post(path.Join(PDPRoutePath, "/piece"), p.handlePiecePost)
 
+	// GET /pdp/piece/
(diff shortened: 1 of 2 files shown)


magik6k (approved):
Seems reasonable

PR #306, opened 2024-10-28
fix: rearrange pieces in sector for no padding

LexLuthr:


harmony/harmonydb/sql/20241028-padding.sql
```diff
@@ -0,0 +1,171 @@
+CREATE TABLE deal_index_tracker (
```
magik6k, line 1: Why do we need this instead of looking e.g. at the _initial_pieces tables and the piece_index there?

magik6k (commented):
I'd like to somehow avoid the added state in the DB, any added thing there will make interactions more complex as it's another thing that needs to be kept in sync with everything.

harmony/harmonydb/sql/20241028-padding.sql
LexLuthr: I wasn't a big fan of this approach while I was writing it. I was trying to find some way to avoid state but could not think of a reliable way which did not involve some really long function. Here is the problem:
1. Typically we can easily identify the piece with piece CID and Size.
2. Problem starts when we have the same piece in more than 1 location in the same sector. How do I reliably connect a piece to a deal. We don't even need to index it later but we still need to know about the offset. This was a reason I went with a stateful solution. 

Another approach I can think of would be to add an index_column to market_mk12_deal_pipeline table and populate it during the transfer functions from within the SQL. It seems cheaper than the current solution.

If you have any insight which avoids any of this, I can really use it.

LexLuthr: From Slack:
We are not going to allow same piece more than once in a sector. This removes the stateful tracking to find unique pieces and their offset in sector.

magik6k (approved):
Don't see anything wrong here, we'll need to make sure to test all this quite well though

PR #307, opened 2024-10-28
add back pressure to mk12 deals

LexLuthr:


market/mk12/mk12.go
```diff
@@ -461,3 +499,144 @@ func (m *MK12) processDeal(ctx context.Context, deal *ProviderDealState) (*Provi
...
+			if sectors[i]+size <= targetSize {
+				sectors[i] += size
+				placed = true
+				break
+			}
+		}
+		if !placed {
+			sectors = append(sectors, size)
+		}
+	}
```
snadrus, line 639: Is this a bin-packing algorithm? There are libraries with great guarantees for this NP-Hard problem.

snadrus (approved):

market/mk12/mk12.go
magik6k: In a way it is, but also it's not much of a problem in filecoin-land because the pieces are generally very large, and usually you can fill 100% of the sector with simple greedy strategies

magik6k (approved):

magik6k:
we may want to propagate this backpressure up to the deal accepting code, keeping the theme of maintaining buffers of work at certain size - so reject deals if the sealing pipelines have enough buffered work AND e.g when there are N tasks waiting for data download

market/mk12/mk12.go
LexLuthr: This is not required anymore with a more simpler backpressure mechanism.

PR #310, opened 2024-10-30
fix: deal label storage in DB

LexLuthr:
We save the label as bytes outside of the proposal and then read the same each time we need proposal. Proposal in sector_metadata and pipeline will still be missing this label but we don't really care about that as Label is only useful for market process.


magik6k (approved):
Seems fine. A little bit messy, but that problem is mk12-only, so eventually in far future will be deprecated away.

PR #316, opened 2024-11-05
fix: Improve supraseal batch ref tracking

magik6k:
This rewrites the local batch ref tracker to:
* Track individual sectors
* Sync state with the DB

In the local tracker there are now 3 states for "slots":
* No sector refs - batch can take more work
* work=true - currently ongoing batch seal task
* Sector refs, work=false - waiting for C1 / Finalize - refs removed from the DB will eventually be removed from the local tracker

### DIFF
--- a/cmd/curio/tasks/tasks.go
+++ b/cmd/curio/tasks/tasks.go
@@ -266,15 +266,13 @@ func addSealingTasks(
 	}
 
 	if cfg.Subsystems.EnableBatchSeal {
-		slotMgr = slotmgr.NewSlotMgr()
-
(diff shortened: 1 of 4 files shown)


LexLuthr (commented):
Shouldn't `s.lk.Lock()` be `s.cond.L.Lock()`?

magik6k:
That's the same lock, the cond is created with `sm.cond = sync.NewCond(&sm.lk)`

magik6k:
Will be in https://github.com/filecoin-project/curio/pull/359

PR #334, opened 2024-11-30
Zen/update interface

ZenGround0:
This PR updates to the latest PDP PRs (still awaiting review)
 - Update pdp_verifier interface to include pass through variables 
 - Update challenge window semantics 
   - New columns to the pdp_proof_sets table to track the challenge window
   - Code in handlers to set up these fields and wait on challenge window being open
   - New code gen for an interface expected of the pdp listener that runs eth calls on the listener contract to get static information about challenge window length and dynamic info about the next challenge window opening

A few things I'm really not sure of
- ~Is it better to be adding columns with ALTER TABLE instead of just changing the schema ?~
- Is there a better way to handle the edge case of the first add? In this PR I just make the nextchallengewindow eth call every time 
- Should we include the ProvingPeriod in pdp_proof_sets table?  It's no longer strictly needed but maybe its nice to keep around? 
- Adding the listener Address twice in pdp_proof_sets and the creates table is probably avoidable.  Is there an easy better way? 
- Are we ok with relying on additional assumptions on the listenerAddr?  Are we ok with it from security perspective? 

I haven't tested this out yet but I'm planning to next week.  Since this code / real databases / SQL is all new to me I think its worth getting your opinion on whether I'm doing something really dumb before I bother running this @magik6k.


harmony/harmonydb/sql/20240930-pdp.sql
```diff
@@ -85,6 +86,11 @@ CREATE TABLE pdp_proof_sets (
     create_message_hash TEXT NOT NULL,
     service TEXT NOT NULL REFERENCES pdp_services(service_label) ON DELETE RESTRICT
 );
+ALTER TABLE pdp_proof_sets 
+ADD COLUMN proofset_listener TEXT NOT NULL, -- the listener is assumed to be a PDP Service with proving window informational methods
+ADD COLUMN proving_period BIGINT NOT NULL, -- the proving period for this proofset
+ADD COLUMN challenge_window BIGINT NOT NULL, -- the challenge window for this proofset
+ADD COLUMN prove_at_epoch BIGINT; -- the epoch at which the next challenge window starts and proofs can be submitted
```
magik6k, line 93: I'd just add this to the main CREATE TABLE above

magik6k: Also can those values just be pulled from the contract getters when scheduling? Better to avoid duplicated state in the database when feasible

pdp/handlers.go
```diff
@@ -181,7 +181,7 @@ func (p *PDPService) handleCreateProofSet(w http.ResponseWriter, r *http.Request
 	}
 
 	// Pack the method call data
-	data, err := abiData.Pack("createProofSet", recordKeeperAddr)
+	data, err := abiData.Pack("createProofSet", recordKeeperAddr, []byte{})
```
magik6k, line 184: What's this last param? We may want to allow the caller (which is the service like Storacha) to set it

ZenGround0: Yeah that sounds about right.  The last param is the one we talked about in [this issue](https://github.com/FilOzone/pdp/issues/56).  I didn't want to mess with the HTTP api in this PR but adding extra data field to all of the user facing methods should probably happen.

ZenGround0:
Confirmed the linter issue is just coming from the base branch

harmony/harmonydb/sql/20240930-pdp.sql
```diff
@@ -68,20 +68,31 @@ CREATE TABLE pdp_piece_mh_to_commp (
...
-    -- task invoking nextProvingPeriod, the task should be spawned any time prev_challenge_request_epoch+provingPeriod is in the past
+    -- task invoking nextProvingPeriod, the task should be spawned any time prove_at_epoch+challenge_window is in the past
     challenge_request_task_id BIGINT REFERENCES harmony_task(id) ON DELETE SET NULL,
 
     -- first proofset add or nextProvingPeriod message hash, when the message lands prove_task_id will be spawned and
     -- this value will be set to NULL
     challenge_request_msg_hash TEXT,
 
+    -- the listener is assumed to be a PDP Service with proving window informational methods
+    proofset_listener TEXT NOT NULL,
```
magik6k, line 87: Any way we can just query this from the Verifier contract? Less off-chain state is generally always desirable, especially if it has any remote possibility of being incorrect here.

magik6k (approved):
One small issue and a question.

Other that that looks good.

harmony/harmonydb/sql/20240930-pdp.sql
ZenGround0: Yes this shouldn't be a problem.

pdp/handlers.go
```diff
@@ -806,6 +808,20 @@ func (p *PDPService) handleAddRootToProofSet(w http.ResponseWriter, r *http.Requ
...
 
+	// Determine the next challenge window start by consulting the listener
+	// This is only needed upon the first add
+	// TODO probably we should only call this when we are at the first add?
+	provingSchedule, err := contract.NewIPDPProvingSchedule(common.HexToAddress(proofSetListener), p.ethClient)
+	if err != nil {
+		log.Errorf("failed to create proving schedule binding, check that listener has proving schedule methods: %w", err)
+		return
+	}
+	next_prove_at, err := provingSchedule.NextChallengeWindowStart(nil, big.NewInt(int64(proofSetIDUint64)))
```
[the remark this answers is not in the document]
ZenGround0: Good catch it actually is not, I've added another method that will return the correct one and will integrate it here.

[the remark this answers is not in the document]
ZenGround0: Fixed

PR #347, opened 2024-12-17
Integrate pdp updates -- nextProvingPeriod called after first add

ZenGround0:
This is an attempt to bring curio up to date with the latest changes here: https://github.com/FilOzone/pdp/pull/86.  The main difference is that now pdp add does not register a challenge generation, that must happen explicitly through a call to nextprovingperiod.  

The approach here is 
1. Create a new harmony task just for initializing proof sets with first call to nextProvingPeriod
2. Differentiate these jobs from regular nextProvingPeriod jobs by checking that prove_at_epoch IS NULL
3.  initProvingPeriod jobs also guarded on new proofset schema column boolean flag init_ready being set to true, which happens on the first add


tasks/pdp/task_init_pp.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+
+	// Get the current tipset
+	ts, err := ipp.fil.ChainHead(ctx)
+	if err != nil {
+		return false, xerrors.Errorf("failed to get chain head: %w", err)
+	}
+
+	// Send the transaction
+	reason := "pdp-proving-period"
```
magik6k, line 175: Those should be unique for each send in the code. Will be later used to display message type stats to users in the UI.

```suggestion
	reason := "pdp-proving-init"
```

magik6k (commented):
Reads correctly, so if it works should be fine to merge

ZenGround0:
Ok @magik6k I've finally got this consistently proving.  Please give it another review.

magik6k (approved):

PR #349, opened 2024-12-19
ffiselect: Set /proc/pid/oom_score_adj=900 for non-PoSt tasks (#329)

### LINKED ISSUE #329: ffiselect: Set /proc/../oom_score_adj
We should be setting `/proc/[childpid]/oom_score_adj` to a high value like `900` on tasks which aren't PoSt so that:
* The main process is less likely to be killed
* Other processes on the machine are less likely to be killed (Lotus, yugabyte)
* Things don't break horribly when the scheduler misjudges how much memory we have (i.e. something outside of curio uses a bunch of memory in the middle of a memory-heavy task)

HiberNuts:
## Description

This PR implements OOM (Out Of Memory) score adjustment for non-PoSt tasks as requested in #329. 

### Changes
- Added `set_oom_score_adj` utility function in `src/utils/oom.rs`
- Modified task spawning in `src/ffiselect/mod.rs` to set high OOM scores (900) for non-PoSt tasks
- Added appropriate error handling and logging

### Implementation Details
- Non-PoSt tasks are assigned an OOM score of 900, making them more likely to be killed under memory pressure
- PoSt tasks retain their default OOM score to ensure their critical operations aren't interrupted
- Failed OOM score adjustments are logged but don't interrupt task execution

### Benefits
1. Protects the main process from OOM kills
2. Preserves other critical system processes (Lotus, yugabyte)
3. Provides better handling of memory pressure situations
4. Maintains system stability when memory usage exceeds expectations

### Testing
- Tested on Linux systems with procfs
- Verified OOM score adjustment for both PoSt and non-PoSt tasks
- Confirmed proper error handling when permission issues occur

Fixes #329

### DIFF
--- a/src/ffiselect/mod.rs
+++ b/src/ffiselect/mod.rs
@@ -1,0 +1,18 @@
+use crate::utils::oom::set_oom_score_adj;
+
+pub fn spawn_task(task: Task) -> Result<Child, Error> {
+    let child = Command::new(task.program)
+        .args(&task.args)
+        // ... other command setup ...
+        .spawn()?;
+
+    // Set high OOM score for non-PoSt tasks
+    if !task.is_post_task() {
+        if let Err(e) = set_oom_score_adj(child.id() as i32, 900) {
+            log::warn!("Failed to set OOM score adjustment: {}", e);
+            // Continue execution even if setting OOM score fails
+        }
+    }
+
+    Ok(child)
+} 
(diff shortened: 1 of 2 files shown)


LexLuthr:
Please post the testing process and results here.

magik6k (changes requested):
This code should be in Go, intergrated with the hidden `ffi` command, here: https://github.com/filecoin-project/curio/blob/main/cmd/curio/ffi.go#L24-L31

LexLuthr:
@HiberNuts Can you please make the requested changes so we can merge this PR?

PR #350, opened 2025-01-02
fix: Inhouse supraseal, fix build

magik6k:
This PR imports the whole supraseal repo into curio, also fixes build.

Best to review commit-by-commit.

The imported commit should match https://github.com/magik6k/supra_seal/tree/feat/multi-out-paths (4b5641401318d37906e56e2dbc61b3ec9d0a9257)


extern/supraseal/build.sh
```diff
@@ -0,0 +1,258 @@
...
+     ./configure --with-virtio --with-vhost
+     make -j 10)
+fi
+if [ ! -d "deps/sppark" ]; then
+    git clone --branch v0.1.10 https://github.com/supranational/sppark.git deps/sppark
+fi
+if [ ! -d "deps/blst" ]; then
+    git clone https://github.com/supranational/blst.git deps/blst
+    (cd deps/blst
+    git checkout bef14ca512ea575aff6f661fdad794263938795d
```
snadrus, line 170: this seems brittle. Is there a branch we can follow? Otherwise we may as well in-house it.

snadrus (approved):

extern/supraseal/build.sh
magik6k: Sadly it's not really versioned / doesn't have a stable branch, and following master was what broke the build previously, so really this is best we can do for now.

PR #351, opened 2025-01-06
fix: Batching Hotfixes

magik6k:


tasks/snap/task_submit.go
```diff
@@ -502,7 +502,7 @@ func (s *SubmitTask) CanAccept(ids []harmonytask.TaskID, engine *harmonytask.Tas
 
 func (s *SubmitTask) TypeDetails() harmonytask.TaskTypeDetails {
 	return harmonytask.TaskTypeDetails{
-		Name: "BatchUpdateSubmit",
+		Name: "UpdateBatch",
```
magik6k, line 505: Was 17 chacacters long, so it caused the node to fail to start. Now the naming matches Commit/PrecommitBatch tasks.

LexLuthr (approved):

PR #352, opened 2025-01-07
fix: libp2p handover

LexLuthr:


market/libp2p/libp2p.go
```diff
@@ -253,18 +254,25 @@ func getCfg(ctx context.Context, db *harmonydb.DB, httpConf config.HTTPConfig, m
 		return nil, xerrors.Errorf("marshaling private key: %w", err)
 	}
 
-	var privKey []byte
-	err = db.QueryRow(ctx, `SELECT update_libp2p_node ($1, $2, $3)`, machine, initialPrivBytes, initialPeerID).Scan(&privKey)
-	if err != nil {
-		return nil, xerrors.Errorf("getting private key from DB: %w", err)
-	}
+	for {
```
magik6k, line 257: We really likely just want to run this loop as a poller, wrapping DealProvider
* Check in the DB every minute who owns the libp2p ticket
  * if it was us, and is still us, and we're running DealProvider already do nothing, just keep polling
  * if it was us, and no longer is us, shut down DealProvider
  * if it wasn't us, and now is us, start DealProvider

magik6k (approved):
Probably fine to merge as is now, but we can make it quite a bit better still

magik6k (approved):
Should work

PR #359, opened 2025-01-09
[do not merge] fix: Catch Batch C1 exceptions

magik6k:


magik6k:
Now in https://github.com/filecoin-project/curio/pull/359

PR #367, opened 2025-01-14
Supraseal robustness branch

magik6k:


snadrus (approved):
Should we add something to the docs folder about this?

lib/slotmgr/slotmgr.go
```diff
@@ -2,42 +2,321 @@ package slotmgr
...
+	}
+
+	// load from DB
+	var slotRefs []struct {
+		SpID         int64  `db:"sp_id"`
+		SectorNumber int64  `db:"sector_number"`
+		PipelineSlot uint64 `db:"pipeline_slot"`
+	}
+	err := db.Select(context.Background(), &slotRefs,
+		`SELECT sp_id, sector_number, pipeline_slot FROM batch_sector_refs WHERE machine_host_and_port = $1`,
```
snadrus, line 63: what if it crashed last time? Will db values get stuck in use?

LexLuthr (approved):

lib/slotmgr/slotmgr.go
magik6k: That's meant to be like that. We want to keep storage for a sector marked as in-use persistently, as long as the sector data exists (This is essentially the same as keeping large "cache" files on disk across restarts)

PR #369, opened 2025-01-14
fix: raw_size in deal pipeline

LexLuthr:


magik6k (approved):
I vaguely remember ipni/indexing also cared about raw size, worth checking. Also might be affecting webui RPCs

PR #376, opened 2025-01-21
add pagination to storageGC page

LexLuthr:
# Without filters
Screenshot 2025-01-22 at 1 17 47 PM

# With Filters
Screenshot 2025-01-22 at 1 29 56 PM

Without pagination, we end up with page refusing to load where GC number is large.
image (13)

### DIFF
--- a/deps/apiinfo.go
+++ b/deps/apiinfo.go
@@ -7,6 +7,7 @@ import (
 	"math/rand"
 	"net/http"
 	"reflect"
+	"strings"
 	"sync"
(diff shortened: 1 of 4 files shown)


magik6k (commented):
Pagination means that it's harder to ctrl-f a given thing.

I'd at least add filters like:
* SP ID checkboxes
* File type checkboxes
* Maybe a "find sector number" input

magik6k (approved):

PR #377, opened 2025-01-21
feat: spark contract

LexLuthr:
- [x] Contract
- [x] Test - contract
- [x] Auto update process
- [x] Delete command
- [x] Test in mainnet
- [x] README


market/ipni/ipni-provider/spark.go
bajtos, line 1: Let's avoid referring to Spark, please. We don't want Spark to be the only retrieval checking solution, we want to enable other builders to implement retrieval-checking solutions too. The smart contract mapping miner IDs to IPNI provider IDs should serve everybody, not just Spark.

How about something based on "registry of index provider IDs" or "PeerID lookup".

market/ipni/spark/sol/README.md
```diff
@@ -0,0 +1,302 @@
...
+      string peerID;
+      bytes signedMessage;
+  }
+  ```
+    - `peerID`: The PeerID associated with the miner.
+    - `signedMessage`: A signed JSON message containing miner and PeerID information.
+
+---
+
+### **Events**
```
bajtos, line 41: It's great that you included events! 👏🏻 This allows us to potentially implement a subgraph indexing these events to provide cheaper access than RPC API calls for the current on-chain state.

bajtos (commented):
I reviewed the design and have concerns about the feasibility/usability of the proposed solution. PTAL at my comments below.

market/ipni/spark/sol/README.md
```diff
@@ -0,0 +1,302 @@
...
+    } else {
+      console.log("Signature is invalid.");
+      return false;
+    }
+  } catch (error) {
+    console.error("Error during verification:", error.message);
+    throw error;
+  }
+}
+```
```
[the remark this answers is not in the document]
LexLuthr: Yes, they signed with existing on chain ID.

We don't verify the signature in contract. It is too expensive to do so. We only verify that sender is a control address for the miner address. 

Hmm! For Curio it would be an easy fix. We can verify the signature and if it does not match then update the entry. This would ensure that if peerID changes then same is updated in the contract as well. Although, in real world, users won't be changing peerID. In fact, we don't have any options to update it in a cluster.

market/ipni/spark/sol/README.md
```diff
@@ -0,0 +1,302 @@
...
+
+## **Contract Functions**
+
+### **Ownership Management**
+
+- **Transfer Ownership**
+  ```solidity
+  function transferOwnership(address newOwner) public onlyOwner;
+  ```
+  Allows the owner to transfer contract ownership to a new address.
```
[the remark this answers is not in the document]
LexLuthr: We probably don't need to ever allow admin access. I will remove those parts.

market/ipni/ipni-provider/spark.go
LexLuthr: The PR name and packages are Curio internal. I don't see nay advantage of "registry of index provider IDs" or "PeerID lookup" over "Spark". In fact, "Spark" immediately tells us what the use case is here. The contract itself is **NOT** named "Spark".

market/ipni/spark/sol/README.md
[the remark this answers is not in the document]
LexLuthr: Removed all admin access to the data.

market/ipni/spark/sol/README.md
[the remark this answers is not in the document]
LexLuthr: Fixed the peerID signature mismatch in Curio code. It checks and pushes a fix if required.

market/ipni/ipni-provider/spark.go
```diff
@@ -0,0 +1,230 @@
...
+
+		toAddr, err := to.ToFilecoinAddress()
+		if err != nil {
+			return xerrors.Errorf("failed to convert Eth address to Filecoin address: %w", err)
+		}
+
+		// Parse the contract ABI
+		parsedABI, err := eabi.JSON(strings.NewReader(spark.GetPeerAbi))
+		if err != nil {
+			log.Fatalf("Failed to parse getPeer ABI: %v", err)
```
magik6k, line 73: Return error

market/ipni/ipni-provider/spark.go
```diff
@@ -0,0 +1,230 @@
...
+		}
+
+		// Encode the function call
+		callData, err := parsedABI.Pack("getPeerData", pInfo.SPID)
+		if err != nil {
+			log.Fatalf("Failed to pack function call data: %v", err)
```
magik6k, line 79: Return error, don't do fatal which will crash the node

market/ipni/ipni-provider/spark.go
LexLuthr: Fixed

market/ipni/ipni-provider/spark.go
LexLuthr: Fixed

market/ipni/spark/sol/contracts/MinerPeerIDMapping.sol
```diff
@@ -0,0 +1,114 @@
...
+        bytes memory signedMessage
+    ) public {
+        require(bytes(minerToPeerData[minerID].peerID).length == 0, "Peer data already exists for this MinerID");
+
+        require(isControllingAddress(msg.sender, minerID), "Caller is not the controlling address");
+
+        minerToPeerData[minerID] = PeerData(newPeerID, signedMessage);
+
+        emit PeerDataAdded(minerID, newPeerID);
+    }
```
bajtos, line 42: Let me double-check that I understand this correctly.

1. Only a wallet controlling the `minerID` can modify the peer ID mapping for the `minerID`, thanks to the `isControllingAddress` requirement on L37.
2. Therefore, Spark does not need to validate the signature in `PeerData` to establish trust that we are getting the `PeerID` as provided by the miner (and not any adversary 3rd party).

Is that correct?

Also, IIUC, the `signedMessage` allows us to verify that the entity submitting the MinerId->PeerID has access to the private key for PeerID. This prevents miners from submitting a PeerID of somebody else, e.g. in order to redirect all Spark requests to an SP using that other peer ID. That's a great feature to have in this contract. 

I am not sure if (or how soon) we we need it in Spark, but it's great to have it in the smart contract from the start.

market/ipni/spark/sol/README.md
```diff
@@ -0,0 +1,302 @@
+# **MinerPeerIDMapping Smart Contract**
+
+## **Overview**
+
+The `MinerPeerIDMapping` contract provides a decentralized way to map Filecoin miner IDs (`uint64`) to their PeerIDs along with signed messages for validation. It ensures only authorized entities (owner or controlling addresses) can add, update, or delete mappings.
```
bajtos, line 5: Should this line also mention that the smart contract collects data necessary to verify that the miner is also controlling the PeerID submitted to the mapping table?

market/ipni/spark/sol/README.md
```diff
@@ -0,0 +1,302 @@
...
+
+---
+
+## **Signed Message Format**
+
+The signed message stored in the contract contains JSON-encoded information:
+
+```json
+{
+  "miner": <uint64>,
```
bajtos, line 125: I think it would be a better developer experience if the miner fields were using the `f0` string format since that's the value returned by JSON RPC calls.

market/ipni/spark/sol/README.md
```diff
@@ -0,0 +1,302 @@
...
+### Verify the signature for auth check
+
+```javascript
+import { HttpJsonRpcConnector, LotusClient } from "filecoin.js";
+import * as libp2pCrypto from "@libp2p/crypto"; // Ensure this library is installed
+import * as multihashes from "multihashes"; // For decoding PeerID
+
+/**
+ * Fetch miner's PeerID and verify the signature of a message.
+ * @param {string} minerAddress - Filecoin miner address
```
bajtos, line 171: See, this line expects the minerAddress to be an `f0` string.

IIUC, to fully validate the mapping, we need also need to check that `minerAddress` matches `message.miner` and `minerInfo.PeerID` matches `message.peerID`.

It would be great to include this step in the example code too. Maybe as a new function showing how to fully verify the mapping entry?

market/ipni/spark/sol/README.md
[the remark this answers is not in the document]
bajtos: Perfect! Can you please document this requirement somewhere in the docs so that other implementers like Venus know about it?

bajtos (approved):
Thank you for simplifying the contract!

I have a few comments to consider, see above.

The new version looks good to me 👍🏻  Can you please let me know what is the address of the mainnet deployment we can use in Spark and the FRC?

market/ipni/spark/sol/contracts/MinerPeerIDMapping.sol
LexLuthr: 1. Yes.
2. But I would still recommend verifying the signature to be 100% sure.

market/ipni/spark/sol/README.md
LexLuthr: Signed message line says the same. Any public signature byte can be verified.

market/ipni/spark/sol/README.md
LexLuthr: README will be updated once I have the tested the contract. README was written for the first version.

market/ipni/spark/sol/README.md
bajtos: I see, that makes sense 👍🏻

market/ipni/spark/sol/contracts/MinerPeerIDMapping.sol
bajtos: > 1. Yes.
> 2. But I would still recommend verifying the signature to be 100% sure.

👍🏻

market/ipni/spark/sol/README.md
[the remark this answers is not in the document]
LexLuthr: This is not a library function. Not sure what you want me to document. This is implementation specific. Venus might want to do something else.

market/ipni/spark/sol/README.md
LexLuthr: Matching minerID is unnecessary here. We are already matching that in the contract. It is the function input. 
I can show certain things in example but no way write the exact code folks might need.

You last comment is very confusing. Please write out in detail what exactly are you expecting here.

FYI:
1. We sign message using the on chain peerID.
2. Signing it with IPNI specific peerID is useless. This would not be secure enough. IPNI ads are already signed by this peerID for verification.

So, I am not sure why you think signing with IPNI specific peerID would be better.

bajtos: > Matching minerID is unnecessary here. We are already matching that in the contract.

Agreed.

> You last comment is very confusing. Please write out in detail what exactly are you expecting here.

Let me try :)

I don't see value in signing the message using the on-chain PeerID. The smart contract has already ensured the entity submitting the MinerID->PeerID message controls the miner.

If the message contains IPNI PeerID but is signed using a different PeerID, then a malicious miner can send a message mapping their miner to your retrieval provider, thus redirecting all Spark retrieval requests to somebody else.

That's why I think the message should be signed using the PeerID inside the message. This way, only the entity with access to PeerID's private key can submit such a message.

I would expect the example to show something like the code below (I am writing it from the top of my head, it may not run out of the box):

```js
async function verifyMinerSignature(minerID, peerID, signature) {
  // Step 1: Reconstruct the message that was signed
  const message = JSON.stringify({miner: `f0${minerID}`, peer: peerData.peerID})

  // Step 2: Decode PeerID to obtain the public key
  const decodedPeerID = multihashes.decode(Buffer.from(peerID, "base64"));
  const publicKeyBytes = decodedPeerID.digest;
 
  // Step 3: Generate a Libp2p public key object
  const publicKey = await libp2pCrypto.keys.unmarshalPublicKey(publicKeyBytes);

  // Step 4: Verify the signature
  const isValid = await publicKey.verify(message, signature);

  if (isValid) {
    console.log("Signature is valid.");
    return true;
  } else {
    console.log("Signature is invalid.");
    return false;
  }
}

// example usage

const minerID = 1000
const {peerID, signedMessage} = await contract.getPeerData(minerID);
await verifyMinerSignature(minerID, peerID, signedMessage)
```

market/ipni/spark/sol/README.md
bajtos: I see, you are referring to this line:

https://github.com/filecoin-project/curio/pull/377/files/dce763f30955515840aa032240065638e3a8dcbc#diff-8ee901221d95fab247080ab3104e7497e45f59edc37e7b8d9d14f2811fdc03f4R130

> This message is signed using the private key associated with the on-chain PeerID.

As I am trying to explain in https://github.com/filecoin-project/curio/pull/377#discussion_r1937726379 and https://github.com/filecoin-project/curio/pull/377#discussion_r1954671775, I think this is a mistake - the message should be signed using the private key associated with with the PeerID in the message.

market/ipni/spark/sol/README.md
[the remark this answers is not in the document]
bajtos: >This is not a library function. Not sure what you want me to document. This is implementation specific. Venus might want to do something else.

If we agree to sign the message using the PeerID from StateMinerInfo (which I think we should not), then I would the documentation to remind other implementers (e.g. Venus) to update this mapping after the on-chain PeerID change.

This 👇🏻 

> We can verify the signature and if it does not match then update the entry. This would ensure that if peerID changes then same is updated in the contract as well.

market/ipni/spark/sol/README.md
LexLuthr: Malicious miner cannot sign using other miner's peer ID. That requires private key access. But I don't mind switching the signature. It would make things simpler on my side. I will push a commit with changes. I don't think we need to make any changes to the contract itself. Just docs and code handling the automation.

market/ipni/spark/sol/README.md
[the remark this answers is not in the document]
LexLuthr: Switched to requested peerID.

market/ipni/spark/sol/README.md
LexLuthr: Updated the example based on your comment and switched to using the IPNI specific peerID as requested.

bajtos (approved):
I discovered a bug in the example code in README, I think it's important to fix it. It's not preventing this pull request from being landed, though.

I also have a suggestion to rename `signedMessage` to `signature`, but I understand it may not be important enough at this stage.

market/ipni/spark/sol/README.md
```diff
@@ -167,51 +157,40 @@ import * as libp2pCrypto from "@libp2p/crypto"; // Ensure this library is instal
...
-      console.log("Signature is invalid.");
-      return false;
-    }
-  } catch (error) {
-    console.error("Error during verification:", error.message);
-    throw error;
+
+async function verifyMinerSignature(minerID, peerID, signature) {
+  // Step 1: Reconstruct the message that was signed
+  const message = JSON.stringify({miner: `f0${minerID}`, peer: peerID})
```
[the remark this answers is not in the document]
LexLuthr: My bad, it should be `int64`. Forgot to update the example.

market/ipni/spark/sol/README.md
```diff
@@ -0,0 +1,329 @@
...
+
+---
+
+### **Structs**
+
+- **PeerData**
+  ```solidity
+  struct PeerData {
+      string peerID;
+      bytes signedMessage;
```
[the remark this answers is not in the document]
LexLuthr: Well! That would require changes in contract and a new contract being deployed. Is that okay?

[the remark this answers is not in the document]
LexLuthr: Done.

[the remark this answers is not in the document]
bajtos: > That would require changes in contract and a new contract being deployed. Is that okay?

Deploying a new contract is fine. Spark has not yet been integrated with the contract.

bajtos (approved):

magik6k (approved):

PR #378, opened 2025-01-22
Backport fixes from main

LexLuthr:
This PR skips any fixes which might bring in market code. Thus, we will be missing some features available in main.


LexLuthr:
Curio market is now full developed with a lot of new features. There is no rational use case to backport fixes to Boost adapter anymore.

PR #384, opened 2025-01-27
Delete pdp roots from proofset 

ZenGround0:
The way PDP works users schedule roots for deletion which are then processed during the subsequent next proving period call.

The approach in this PR is 
1. Create a new table `pdp_proofset_root_deletes` for tracking the deletions scheduled on chain
2. The deletion api adds roots to this table
3. The prove task now additionally:
  a. Reads all scheduled deletions
  b. Removes the associated piece refs from parked piece ref table
  c. Cleans up all metadata: pdp_piece_refs, pdp_roots, pdp_proofset_root_deletes

An alternate strategy would be to just directly call the contract to get the ids for deletion.  Its worth considering this, let me know if you prefer it @magik6k


pdp/handlers.go
```diff
@@ -883,6 +883,133 @@ func (p *PDPService) handleAddRootToProofSet(w http.ResponseWriter, r *http.Requ
...
+		}
+
+		return true, nil
+	}, harmonydb.OptionRetry())
+
+	if err != nil {
+		if err.Error() == "root not found" {
+			http.Error(w, "Root not found", http.StatusNotFound)
+			return
+		}
```
magik6k, line 1003: The insert is non-returning, this check doesn't really make sense to me

tasks/pdp/task_prove.go
```diff
@@ -631,6 +636,76 @@ func (p *ProveTask) getSenderAddress(ctx context.Context) (common.Address, error
 	return address, nil
 }
 
+func (p *ProveTask) cleanupDeletedRoots(ctx context.Context, proofSetID int64, pdpVerifier *contract.PDPVerifier) error {
+
+	removals, err := pdpVerifier.GetScheduledRemovals(nil, big.NewInt(proofSetID))
```
magik6k, line 641: Presuming this doesn't need to wait for the message to land?

tasks/pdp/task_prove.go
```diff
@@ -631,6 +636,76 @@ func (p *ProveTask) getSenderAddress(ctx context.Context) (common.Address, error
...
+			}
+
+			// Delete the parked piece ref, this will cascade to the pdp piece ref too
+			_, err = tx.Exec(`
+				DELETE FROM parked_piece_refs
+				WHERE ref_id = $1
+			`, parkedPieceRef)
+			if err != nil {
+				return false, xerrors.Errorf("failed to delete parked piece ref %d: %w", parkedPieceRef, err)
+			}
```
magik6k, line 684: pdp.pdp_pieceref is already a pieceref

parked_piece_refs maps "Piece Refs" to deduplicated "Piece IDs" where there is generally one "PieceID" for each piece-cid (though there can occasionally be duplicates, e.g. when pieces aren't verified to have correct hashes yet)

```suggestion
			// Delete the parked piece ref, this will cascade to the pdp piece ref too
			_, err = tx.Exec(`
				DELETE FROM parked_piece_refs
				WHERE ref_id = $1
			`, pdpPieceRefID)
			if err != nil {
				return false, xerrors.Errorf("failed to delete parked piece ref %d: %w", parkedPieceRef, err)
			}
```

magik6k (commented):
Couple small comments, other that that looks good

tasks/pdp/task_prove.go
ZenGround0: Yeah I think it is ok in this case since we consult chain itself upon proving to check which pieces are ready for removal.

pdp/handlers.go
ZenGround0: My bad this was a relic of the design where I tracked the pending deletes in a new table

magik6k (approved):
Looks correct

PR #387, opened 2025-01-27
fix: Don't rechedule batch sends

magik6k:
This should fix a race condition when there are multiple pollers running and a batch message was sent.

### DIFF
--- a/tasks/seal/poller_commit_msg.go
+++ b/tasks/seal/poller_commit_msg.go
@@ -16,7 +16,6 @@ import (
 	"github.com/filecoin-project/curio/harmony/harmonydb"
 	"github.com/filecoin-project/curio/harmony/harmonytask"
 
-	"github.com/filecoin-project/lotus/chain/actors/builtin/miner"
 	"github.com/filecoin-project/lotus/chain/actors/policy"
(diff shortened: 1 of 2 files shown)


tasks/seal/poller_commit_msg.go
```diff
@@ -154,29 +153,13 @@ func (s *SealPoller) pollStartBatchCommitMsg(ctx context.Context, tasks []pollTa
...
-					return false, xerrors.Errorf("update sectors_sdr_pipeline: %w", err)
-				}
-				if n != 1 {
-					return false, xerrors.Errorf("expected to update 1 row, updated %d", n)
-				}
-
-				return true, nil
-			})
-		}
-	}
```
magik6k, line 171: @LexLuthr I really don't get why this is here, it's not needed, right?

Also at minimum it should return early

LexLuthr (approved):

tasks/seal/poller_commit_msg.go
LexLuthr: That was for aggregation and non aggregate approach I was doing earlier before we moved to always batch but not always aggregate strategy. This can be removed.

PR #395, opened 2025-01-31
fix: finalize ensureOne

LexLuthr:
Fixes the issue where unsealed path is not detected correctly by Finalize ensureOne() function.


phapawit-hir:
Hi there. It works

magik6k (approved):
Ideally we wouldn't be mutating things passed in the arguments, but I guess this fix works, so it's much better than no fix

PR #398, opened 2025-02-01
Fix acme challenge

ZenGround0:
ACME protocol is misconfigured so lets encrypt fails on new cert requests.  See this thread: https://filecoinproject.slack.com/archives/C0717TGU7V2/p1738321145982859


magik6k (approved):
Oooh yeah, that probably broke in a fairly recent merge

PR #403, opened 2025-02-04
feat: gc: Mark snap sector key files for removal

magik6k:
This PR adds additional marking logic to GC Mark task for cleaning up non-needed post-snap sector key files (original CC sector files):

Sectors where:
* `sectors_meta` table indicates a snap sector (orig_unsealed_cid != cur_sealed_cid)
* On-chain sector info is not null (1.5 finality ago)
* On-chain sector info has a non-nil SectorKeyCID indicating the sector is a snap sector
* On disk, the `update` and `update-cache` files are present for the sector, indicating that PoSt logic won't use the sealed file
* Have the sealed file on disk

Will have the sealed file (only) marked for GC

### DIFF
--- a/tasks/gc/storage_gc_mark.go
+++ b/tasks/gc/storage_gc_mark.go
@@ -2,6 +2,7 @@ package gc
 
 import (
 	"context"
+	"github.com/filecoin-project/go-state-types/builtin"
 	"time"
 
 	cbor "github.com/ipfs/go-ipld-cbor"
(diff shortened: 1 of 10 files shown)


tasks/gc/storage_gc_mark.go
```diff
@@ -282,6 +294,139 @@ func (s *StorageGCMark) Do(taskID harmonytask.TaskID, stillOwned func() bool) (d
...
+	if err != nil {
+		return false, xerrors.Errorf("get finality tipset: %w", err)
+	}
+
+	var minerIDs []int64
+	if err = s.db.Select(ctx, &minerIDs, `SELECT DISTINCT sp_id FROM sectors_meta WHERE orig_unsealed_cid != cur_sealed_cid`); err != nil {
+		return false, xerrors.Errorf("distinct miners from snap sectors: %w", err)
+	}
+
+	finalityMinerStates := make(map[abi.ActorID]miner.State)
```
LexLuthr, line 338: Don't we already save the miner state a few lines up? Can we reuse that one?

LexLuthr: Never mind, I see that this one is with specific epoch.

LexLuthr (approved):
Looks good.

Issue #368, opened 2025-01-14
New config defaults don't apply to existing configs


magik6k:
When new config variables are added, the defaults in existing clusters are zero-value, which is quite annoying when new stuff (like markets) is added.

PR #405, opened 2025-02-06
keep base layer updated with new config

LexLuthr:
Fixes https://github.com/filecoin-project/curio/issues/368

### DIFF
--- a/deps/deps.go
+++ b/deps/deps.go
@@ -2,6 +2,7 @@
 package deps
 
 import (
+	"bytes"
 	"context"
 	"crypto/rand"
 	"database/sql"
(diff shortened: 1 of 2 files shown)


deps/deps.go
```diff
@@ -462,6 +467,45 @@ func GetConfig(ctx context.Context, layers []string, db *harmonydb.DB) (*config.
 	return curioConfig, nil
 }
 
+func updateBaseLayer(ctx context.Context, db *harmonydb.DB) error {
```
magik6k, line 470: Should do the upgrade with one db transaction

LexLuthr: Fixed

deps/deps.go
```diff
@@ -462,6 +467,53 @@ func GetConfig(ctx context.Context, layers []string, db *harmonydb.DB) (*config.
...
+		_, err = LoadConfigWithUpgrades(text, cfg)
+		if err != nil {
+			return false, fmt.Errorf("could not read base layer, bad toml %s: %w", text, err)
+		}
+
+		// Convert the update config to string
+		cb, err := config.ConfigUpdate(cfg, config.DefaultCurioConfig(), config.Commented(true), config.DefaultKeepUncommented(), config.NoEnv())
+		if err != nil {
+			return false, xerrors.Errorf("cannot update base config: %w", err)
+		}
```
magik6k, line 494: So here's a moderate issue - if a cluster has some mixed versions of nodes - new version introduced a new config, and it's in base - after an old node which is not aware of that field restarts, the new field would be removed from base. Probably fixable by working on toml.MetaData returned by LoadConfigWithUpgrades

deps/deps.go
```diff
@@ -462,6 +468,131 @@ func GetConfig(ctx context.Context, layers []string, db *harmonydb.DB) (*config.
...
+
+		// Save the updated base with merged comments
+		_, err = tx.Exec("UPDATE harmony_config SET config=$1 WHERE title='base'", finalConfig)
+		if err != nil {
+			return false, xerrors.Errorf("cannot update base config: %w", err)
+		}
+
+		return true, nil
+	})
```
magik6k, line 518: nit: should have harmony.OptionRetry()

magik6k (approved):
Looks good, just one nit

PR #411, opened 2025-02-13
feat: add configurable replication factor for indexstore

strahe:


LexLuthr:
Actually, I would like to know motivation behind this change. Just to clarify, it does not matter to YB even if you make this change. The replication number is controlled by YB master config and not Curio. 
https://docs.yugabyte.com/preview/deploy/checklist/#replication

https://docs.yugabyte.com/stable/api/ycql/ddl_create_keyspace/#semantics


Above makes this PR unnecessary. That's why I put that hardcoded value. I must have removed the comment describing above at some point. Or maybe that was in the PR and not code.

strahe:
In that case, this PR is indeed unnecessary. I will close it.

PR #379, opened 2025-01-22
feat: devnet with market and deal client

LexLuthr:
This PR has following changes:
1. Small weight new mk12 client to make deals in devnet hosted under `sptool toolbox`
2. The client comes with all basic necessities to make deals
3. The Address used by libp2p is ws instead of wss for devnet/testnets
4. A new container called "piece-server" to host remote pieces and also to act as client
5. Updated Go to 1.23
6. Adds IPNI container and tests ad sync for devnet

- [x] Fix devnet API 
- [x] Create market config
- [x] Test changes in devnet
- [x] Add IPNI container
- [x] Test online deal
- [x] Test offline deal
- [x] Test IPNI sync
- [x] Test Retrievals

## IPNI sync working with correct Address in advertisement
Screenshot 2025-02-19 at 1 47 04 PM

## Devnet sealing verified
Screenshot 2025-02-18 at 10 46 50 PM

PR #415, opened 2025-02-13
fix: ipni ad addr value

LexLuthr:
We have been using the "/ipni-provider" for ad.Addresses when we should have been using retrieval address. Luckily, when someone published the new ad with correct address, it should fix the address displayed on cid.contact to correct one.


magik6k (approved):

LexLuthr:
closing in favor of https://github.com/filecoin-project/curio/pull/379

PR #417, opened 2025-02-14
feat: offline verified ddo deals

LexLuthr:
- A new command to add DDO deals to table and pipeline
- A new column in market_mk12_deal_pipeline to identify the DDO deal
- Small changes in mk12 pipeline to accomodate DDO deals
- UI changes for DDO
- Some small fixes to deal pipeline

- [x] Test without commP
- [x] Test with commP
- [x] Test with URL
- [x] Test with PieceLocator
- [x] Verify indexing
- [x] verify Announcing
- [x] Verify no announce is respected
- [x] Verify if !should_index then deal is not announced
- [x] Create a quick DDO deal script
- [x] Verify f05 deal flow is working

## Piece Info showing DDO deal
Screenshot 2025-02-21 at 12 39 54 PM

## DDO deals in deal list page
Screenshot 2025-02-21 at 12 40 01 PM

## Deal pipeline showing DDO deal
Screenshot 2025-02-21 at 12 40 12 PM

## F05 Deals with DDO
Screenshot 2025-02-21 at 4 55 06 PM

## Completed F05 deals
Screenshot 2025-02-21 at 5 30 01 PM

### DIFF
--- a/harmony/harmonytask/singleton_task.go
+++ b/harmony/harmonytask/singleton_task.go
@@ -17,18 +17,22 @@ func SingletonTaskAdder(minInterval time.Duration, task TaskInterface) func(AddT
 		add(func(taskID TaskID, tx *harmonydb.Tx) (shouldCommit bool, err error) {
 			var existingTaskID *int64
 			var lastRunTime time.Time
+			var shouldRun bool
+
+			now := time.Now()
 
 			// Query to check the existing task entry
 			err = tx.QueryRow(`SELECT task_id, last_run_time FROM harmony_task_singletons WHERE task_name = $1`, taskName).Scan(&existingTaskID, &lastRunTime)
-			if err != nil {
-				if !errors.Is(err, pgx.ErrNoRows) {
-					return false, err // return error if query failed and it's not because of missing row
-				}
+			if errors.Is(err, pgx.ErrNoRows) {
+				// No existing record → Task should run
+				shouldRun = true
+			} else if err != nil {
+				return false, err // Return actual error
+			} else {
+				// Should run if no existing ID or last time is 0
+				shouldRun = existingTaskID == nil || lastRunTime.Add(minInterval).Before(now)
 			}
 
-			now := time.Now().UTC()
-			// Determine if the task should run based on the absence of a record or outdated last_run_time
-			shouldRun := err == pgx.ErrNoRows || (existingTaskID == nil && lastRunTime.Add(minInterval).Before(now))
 			if !shouldRun {
 				return false, nil
 			}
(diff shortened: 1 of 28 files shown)


harmony/harmonytask/singleton_task.go
```diff
@@ -17,31 +17,48 @@ func SingletonTaskAdder(minInterval time.Duration, task TaskInterface) func(AddT
...
 			// Conditionally insert or update the task entry
 			n, err := tx.Exec(`
                 INSERT INTO harmony_task_singletons (task_name, task_id, last_run_time)
-                VALUES ($1, $2, $3)
-                ON CONFLICT (task_name) DO UPDATE
-                SET task_id = COALESCE(harmony_task_singletons.task_id, $2),
-                    last_run_time = $3
-                WHERE harmony_task_singletons.task_id IS NULL
-            `, taskName, taskID, now)
+				VALUES ($1, $2, $3)
```
magik6k, line 43: I don't really understand what this change does

market/ipni/ipni-provider/ipni-provider.go
```diff
@@ -466,34 +466,38 @@ func Routes(r *chi.Mux, p *Provider) {
...
-			return append(slice[:i], slice[i+1:]...)
+	var filtered []*url.URL
+
+	for _, u := range slice {
+		if !strings.Contains(u.String(), target) {
+			filtered = append(filtered, u)
 		}
 	}
-	return slice
+	return filtered
```
magik6k, line 476: could use `lo.Filter`

tasks/indexing/task_indexing.go
```diff
@@ -248,6 +255,40 @@ func (i *IndexingTask) recordCompletion(ctx context.Context, task itask, taskID
...
+		return nil, xerrors.Errorf("getting pending indexing tasks: %w", err)
+	}
+
+	for _, t := range ac {
+		if _, ok := acceptables[t.TaskID]; !ok {
+			continue
+		}
+
+		// Accept any task which should not be indexed as
+		// it does not require storage access
```
magik6k, line 286: Technically true, but we really, really want to run indexing on the storage node. Running indexing remotely will use massive amount of bandwidth unnecessarily

magik6k (commented):
Couple of small comments/questions.

Can we test that non-ddo deals still work with the changes here?

tasks/indexing/task_indexing.go
LexLuthr: This logic is only for those task which should not be indexed. The reason is simple, if I said don't keep unsealed and there is no unsealed then indexing task will never be accepted. Thus allow these tasks to be picked up by any node and then do will simply mark them as finished and make the entry about the piece in the table. Since DO() is doing a few things we need, I am not putting the logic to mark complete in taskID creation function like for IPNI job.

harmony/harmonytask/singleton_task.go
LexLuthr: For Singleton, I am still seeing issues with SP where we have a task_id in singleton table but nothing in harmony. Could be for any reason like bad shutdown etc. But this results in no more GC of any kind. Thus the new logic to update the ID if no corresponding entry in harmony_task.

tasks/indexing/task_indexing.go
magik6k: Oooh

Why is `ShouldIndex` not just a part of the `where` clause in the query?

magik6k (approved):

PR #419, opened 2025-02-17
fix: improved PSD selection and findURL in market

LexLuthr:


magik6k (commented):
The only problem I see that needs addressing is that this can assign one task ID to deals from multiple SPs

Probably should create tasks in a loop, and in each task add we should only schedule PSD for up to one SP

magik6k (approved):

PR #421, opened 2025-02-24
fix: improve filter

LexLuthr:


market/mk12/mk12.go
```diff
@@ -780,14 +782,18 @@ FROM joined
 func (m *MK12) applyFilters(ctx context.Context, deal *ProviderDealState) *validationError {
 
 	var clientRules []struct {
+		Name               string   `db:"name"`
 		Wallets            []string `db:"wallets"`
-		PeerIDs            []string `db:"peer_id"`
+		PeerIDs            []string `db:"peer_ids"`
```
magik6k, line 787: How did this work? Did it even work?

LexLuthr: Nope. It seem, I think redid this part a few times along with UI and DB changes. 
Tested end to end everything. It was hard to test this stuff correctly without a proper client. We have that now.

magik6k (approved):

PR #423, opened 2025-02-26
update slack alert format

LexLuthr:


alertmanager/plugin/slack_webhook.go
```diff
@@ -64,19 +65,23 @@ func (s *SlackWebhook) SendAlert(data *AlertPayload) error {
 
 	// Iterate through the map to construct the remaining blocks
 	for key, value := range data.Details {
+		// Split value into sentences by period followed by space
+		sentences := strings.Split(value.(string), ". ")
```
snadrus, line 69: A bit nervous about what we do to translate here.

snadrus (approved):
Have we considered what to do with translations in regards to these emails?

It should not be too hard to use templates for each sentence. And then just have a translated subject. We will need to have a drop down somewhere about which language to use for emails.

alertmanager/plugin/slack_webhook.go
LexLuthr: These are not emails. These are simple slack messages.

LexLuthr: Our language translation does not seem to be getting any traction or feedback so I am not sure if I want to extend it to alerts at the moment. If someone asks for it, I will be happy to get the translation worked out for alerts.

PR #424, opened 2025-02-26
update metrics

LexLuthr:
This PR added following new metrics:

- task_duration_seconds_per_task
- task_wait_duration_seconds_per_task
- task_wait_duration_seconds
- task_failure_duration_seconds
- task_failure_duration_seconds_per_task

These should make it easy to identify gaps in pipeline and scheduling.


LexLuthr:
Closing this based on slack discussion. Histograms to too computationally expensive and we can do most of this from DB's history table.

strahe:
Just leaving a comment: I tried an alternative approach to export some SQL results as metrics, which might be helpful for some people.

https://github.com/web3tea/curio-exporter

PR #428, opened 2025-03-04
fix: deal removal and list

LexLuthr:
This PR fixes the deal removal button. It was broken after DDO deals.

It also fixes the Deal list to correctly say processed instead of Stored and displayed the error
Screenshot 2025-03-04 at 1 41 24 PM


magik6k (approved):
Quite neat

PR #431, opened 2025-03-05
feat: Balance Manager

LexLuthr:
This PR adds a new task which can automatically top off the miner's market balance. This balance is used for deal collateral. 

- [x] Test in devnet
- [x] Test after refactor

## Starting with 0 balance
Screenshot 2025-03-05 at 10 42 56 PM

## 10 FIL added from worker wallet
Screenshot 2025-03-05 at 10 49 28 PM

## 50 FIL Added from collateral wallet when Threshold is changed to 15 FIL
Screenshot 2025-03-05 at 10 52 14 PM

### DIFF
--- a/deps/config/types.go
+++ b/deps/config/types.go
@@ -105,6 +105,8 @@ func DefaultCurioConfig() *CurioConfig {
 					PublishMsgPeriod:          5 * time.Minute,
 					MaxDealsPerPublishMsg:     8,
 					MaxPublishDealFee:         types.MustParseFIL("0.5 FIL"),
+					CollateralAddThreshold:    types.MustParseFIL("5 FIL"),
+					CollateralAddAmount:       types.MustParseFIL("5 FIL"),
 					ExpectedPoRepSealDuration: 8 * time.Hour,
 					ExpectedSnapSealDuration:  2 * time.Hour,
 					CIDGravityTokens:          []string{},
@@ -366,6 +368,9 @@ type CurioSubsystemsConfig struct {
 	// The maximum amount of indexing and IPNI tasks that can run simultaneously. Note that the maximum number of tasks will
 	// also be bounded by resources available on the machine. (Default: 8)
 	IndexingMaxTasks int
+
+	// EnableMarketBalanceManager enabled the task to automatically manage the market balance of the miner's market actor (Default: false)
+	EnableMarketBalanceManager bool
 }
 type CurioFees struct {
 	// maxBatchFee = maxBase + maxPerSector * nSectors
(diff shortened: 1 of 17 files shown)


deps/config/types.go
```diff
@@ -691,6 +696,18 @@ type MK12Config struct {
...
+	// utilized for deal collateral in market (f05) deals. If no wallet is set, worker wallet will be used for this.
+	DealCollateralWallet string
+
+	// CollateralAddThreshold is the balance below which more balance will be added to miner's market balance
+	// Accepts a decimal string (e.g., "123.45" or "123 fil") with optional "fil" or "attofil" suffix. (Default: "5 FIL")
+	CollateralAddThreshold types.FIL
+
+	// CollateralAddAmount is the amount added to miner's market balance when it falls below CollateralAddThreshold
+	// Accepts a decimal string (e.g., "123.45" or "123 fil") with optional "fil" or "attofil" suffix. (Default: "5 FIL")
+	CollateralAddAmount types.FIL
```
magik6k, line 709: Can we do "Low/High" watermark? If balance falls below "low", top up to "High"

tasks/storage-market/market_balance.go
```diff
@@ -0,0 +1,189 @@
...
+
+			msg := &types.Message{
+				To:     marketActor.Address,
+				From:   sender,
+				Value:  amount,
+				Method: marketActor.Methods.AddBalance,
+				Params: params,
+			}
+
+			mcid, err := m.sender.Send(ctx, msg, msp, "Add Market Collateral")
```
magik6k, line 156: All other places use this-kind-of-case, best to stay consistent, later on we can create a nice webui for viewing message sends by address/type/etc, showing e.g. where gas fees go

```suggestion
			mcid, err := m.sender.Send(ctx, msg, msp, "add-market-collateral")
```

deps/config/types.go
```diff
@@ -365,6 +367,9 @@ type CurioSubsystemsConfig struct {
 	// The maximum amount of indexing and IPNI tasks that can run simultaneously. Note that the maximum number of tasks will
 	// also be bounded by resources available on the machine. (Default: 8 - unlimited)
 	IndexingMaxTasks int
+
+	// EnableMarketBalanceManager enabled the task to automatically manage the market balance of the miner's market actor (Default: false)
+	EnableMarketBalanceManager bool
```
magik6k, line 372: For separate PR, but we could create a separate balance manager, similar to the lotus-shed balancer command (https://github.com/filecoin-project/lotus/blob/master/cmd/lotus-shed/balancer.go): Could define groups with source and destination addresses, and have it automatically shuffle funds e.g. to commit/precommit control addresses

magik6k (commented):
Generally this looks good, only semi-minor comment is that we probably don't want just one global source wallet given that most people use curio with multiple SP IDs often operated for entirely separate entities

deps/config/types.go
```diff
@@ -388,6 +397,10 @@ type CurioFees struct {
 
 	// Don't send collateral with messages even if there is no available balance in the miner actor (Default: false)
 	DisableCollateralFallback bool
+
+	// BalanceManagerConfig specifies the configuration parameters for managing wallet balances and actor-related funds,
+	// including collateral and other operational resources.
+	BalanceManager BalanceManagerConfig
```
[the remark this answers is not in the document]
LexLuthr: I agree that it gives more flexibility to add wallets per miner ID. Making TOML work with it is going to be a nightmare.

snadrus (approved):

PR #435, opened 2025-03-09
Feat/upload bigger than sector size

ZenGround0:
Support uploads working for files > 64 GiB
Update ABI
Also deal with a long overdue integration of RootsAdded event structure changing


ZenGround0:
Test failure is a CI build issue

PR #447, opened 2025-03-25
feat: add Curio PDP docs

rjan90:
All credits to: @ZenGround0 for writing the docs. These are just porting these over to the correct repo from https://github.com/filecoin-project/curio-docs/pull/2

### DIFF
--- a/documentation/en/curio-pdp.md
+++ b/documentation/en/curio-pdp.md
@@ -1,0 +1,105 @@
+---
+description: This page explains how to setup a Curio PDP SP
+---
+
+# Curio PDP
+
+This guide is focused on running a PDP-only SP.  Most of the same instructions will apply for running a curio node providing both PoRep and PDP storage though you will likely need to modify the configuration defaults specified here to keep the sealing pipeline running smoothly.
+
+## Setup curio 
+
+The first step to running a curio PDP SP is to install and setup curio.
+
+Some important callouts departing from the ordinary curio setup instructions
+* You MUST check out `feat/pdp` branch and build curio from source on this branch. PDP code is not yet on `main`
+* The `feat/pdp` branch has not been updated to go v1.24.x so running PDP SP requires building with go v1.23.x
+* Follow the guidelines under `Initiating a new Curio cluster` including the `guided-setup` command.  This means that you need to create a miner actor onchain (i.e. PoRep infrastructure). While unnecessary for PDP this is the quickest and easiest way to get PDP working on curio.
+
+
+With these points in mind proceed to the [curio setup instructions](setup.md#setup) to setup curio.
+
+## Curio configuration
+
+
+### Config 
+
+Curio PDP requires specific configuration options. Briefly you will need to modify your config to look like this: 
+
+```toml
+[HTTP]
+  DomainName = "your.domain.name.com"
+  Enable = true
+  IdleTimeout = 7200000000000
+  ListenAddress = "0.0.0.0:443"
+  ReadHeaderTimeout = 7200000000000
+  ReadTimeout = 7200000000000
+
+[Subsystems]
+  EnableCommP = true
+  EnableMoveStorage = true
+  EnablePDP = true
+  EnableParkPiece = true
+```
+
+### Explanation
+
+The domain name must point to the IP address your curio node is listening on. The purpose of setting this field is to allow lets encrypt ACME protocol to automatically issue a certificate to use TLS for encrypting access to the curio api. For let's encrypt policy reasons this will only work if curio listens on port 443.  
+
+The maxed out timeouts are a suggestion to maximally avoid data transfer timeout issues.
+
+The curio CommP, MoveStorage, PDP and ParkPiece subsystems are all necessary for pdp to function properly and must be enabled.
+
+### Setting config
+
+Copy the above toml to a file `pdp-layer` and then run `curio config create pdp-layer`.
+
+Then start up curio `./curio run --layers gui,pdp-layer`.
+
+Alternatively you can start curio with `./curio run --layers gui`, navigate to localhost:4701 -> Configurations -> Add Layer (pdp-layer) and then manually update the necessary subsystems and HTTP options.  Then restart with the new config layer applied `./curio run --layers gui,pdp-layer`
+
+To set your curio configuration values you will need to create a new pdp specific configuration layer.
+
+## Storage attach
+
+Running a PDP SP means storing data. Curio requires manually adding data storage paths to actually store data. Run `./curio cli storage attach --init --store /path/to/storage ` to setup a storage path. Now your curio node can store PDP files.
+
+## PDP Service
+
+To interact with the curio PDP api clients need to have authentication to access a PDP API Service. PDP SPs must create PDP Services for clients to use their API. Here is the creation flow.
+
+1. Client creates local authentication token offline, they then send PEM public key to curio PDP SP
+2. In curio web ui (localhost:4701) navigate to PDP and click "Add Service" 
+3. Fill out service name (i.e. clientX) and paste their public key
+
+## PDP owner
+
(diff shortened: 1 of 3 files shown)


ZenGround0 (approved):
LGTM though I noticed you haven't ported over your nice suggestions from the other PR

documentation/en/curio-pdp.md
LexLuthr, line 1: All features which are not yet part of the main branch should be housed under https://docs.curiostorage.org/experimental-features

We can move these out once pdp PR is merged to main.

Another issue is that this Page is PDP-Only SP which I don't think is the correct way to do things. So, instead of calling it pdp-only SP, let's reword things to "Enable PDP" for Curio. 

That way this has enough information to still be "pdp-only" SP but also PoRep+PDP SP.

documentation/en/curio-pdp.md
```diff
@@ -0,0 +1,105 @@
...
+description: This page explains how to setup a Curio PDP SP
+---
+
+# Curio PDP
+
+This guide is focused on running a PDP-only SP.  Most of the same instructions will apply for running a curio node providing both PoRep and PDP storage though you will likely need to modify the configuration defaults specified here to keep the sealing pipeline running smoothly.
+
+## Setup curio 
+
+The first step to running a curio PDP SP is to install and setup curio.
```
LexLuthr, line 11: Let's link the page here to installation and setup. With detailing where departure is from normal PoRep SP setup.

documentation/en/curio-pdp.md
```diff
@@ -0,0 +1,105 @@
...
+
+Some important callouts departing from the ordinary curio setup instructions
+* You MUST check out `feat/pdp` branch and build curio from source on this branch. PDP code is not yet on `main`
+* The `feat/pdp` branch has not been updated to go v1.24.x so running PDP SP requires building with go v1.23.x
+* Follow the guidelines under `Initiating a new Curio cluster` including the `guided-setup` command.  This means that you need to create a miner actor onchain (i.e. PoRep infrastructure). While unnecessary for PDP this is the quickest and easiest way to get PDP working on curio.
```
LexLuthr, line 16: We should link to exactly the section we want users to be looking at.

Maybe even add the link to PDP page from setup page. PDP documentation need not be limited to this page.

documentation/en/curio-pdp.md
```diff
@@ -0,0 +1,105 @@
...
+
+This guide is focused on running a PDP-only SP.  Most of the same instructions will apply for running a curio node providing both PoRep and PDP storage though you will likely need to modify the configuration defaults specified here to keep the sealing pipeline running smoothly.
+
+## Setup curio 
+
+The first step to running a curio PDP SP is to install and setup curio.
+
+Some important callouts departing from the ordinary curio setup instructions
+* You MUST check out `feat/pdp` branch and build curio from source on this branch. PDP code is not yet on `main`
+* The `feat/pdp` branch has not been updated to go v1.24.x so running PDP SP requires building with go v1.23.x
```
LexLuthr, line 15: This basically means, users cannot build on calibnet or mainnet. We should call that out.

documentation/en/curio-pdp.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Config 
+
+Curio PDP requires specific configuration options. Briefly you will need to modify your config to look like this: 
+
+```toml
+[HTTP]
+  DomainName = "your.domain.name.com"
+  Enable = true
+  IdleTimeout = 7200000000000
```
LexLuthr, line 32: All duration should be human friendly like "0h5m10s"

documentation/en/curio-pdp.md
```diff
@@ -0,0 +1,105 @@
...
+## Curio configuration
+
+
+### Config 
+
+Curio PDP requires specific configuration options. Briefly you will need to modify your config to look like this: 
+
+```toml
+[HTTP]
+  DomainName = "your.domain.name.com"
```
LexLuthr, line 30: Domain name is usually suggested to be in base layer. Otherwise other systems which use domain name like IPNI and UI cannot get access to it. But that might not apply to PDP only SP. So we should probably put a callout differentiating the 2 kind of SPs here.

LexLuthr (changes requested):
We need some changes to align with Curio docs.

documentation/en/curio-pdp.md
rjan90: Moved and renamed the page here: https://github.com/filecoin-project/curio/pull/447/commits/a5613c541ebb26367871dd5ad883588691fef855

documentation/en/curio-pdp.md
rjan90: Updated to human friendly here: https://github.com/filecoin-project/curio/pull/447/commits/919dc70bfc26bce3bc325392ca6f0ddaf9e226dc

documentation/en/curio-pdp.md
rjan90: I think it should still work with Go v1.23.7, but this needs confirmation from @ZenGround0 which is currently running the Calibnet Curio PDP SP displayed here: https://calibration.pdp-explorer.eng.filoz.org/

So maybe we only need to write that Go v1.23.7 is the only version you can be on at the moment

documentation/en/curio-pdp.md
```diff
@@ -0,0 +1,105 @@
...
+
+Then start up curio `./curio run --layers gui,pdp-layer`.
+
+Alternatively you can start curio with `./curio run --layers gui`, navigate to localhost:4701 -> Configurations -> Add Layer (pdp-layer) and then manually update the necessary subsystems and HTTP options.  Then restart with the new config layer applied `./curio run --layers gui,pdp-layer`
+
+To set your curio configuration values you will need to create a new pdp specific configuration layer.
+
+## Storage attach
+
+Running a PDP SP means storing data. Curio requires manually adding data storage paths to actually store data. Run `./curio cli storage attach --init --store /path/to/storage ` to setup a storage path. Now your curio node can store PDP files.
```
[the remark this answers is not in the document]
rjan90: Added link to the storage config page here: https://github.com/filecoin-project/curio/pull/447/commits/bba7e8a6b9558ea29e92de31f01fc93de30c13f0

documentation/en/curio-pdp.md
rjan90: @ZenGround0 can clarify if this is the case for PDP only SPs? Or if we can add it to the base layer?

documentation/en/curio-pdp.md
LexLuthr: Oh! My bad. I misread this. You can ignore this one.

LexLuthr:
@rjan90 I think you forgot to link the page to menu. It is not showing up in the docs. You can view the updated docs by clicking on Gitbook test.

rjan90:
Thanks for noticing @LexLuthr! Was a bit unfamiliar with how these were linked. But added it here: https://github.com/filecoin-project/curio/pull/447/commits/572f47a051eaa340f7d486af95bd793db24777f0

documentation/en/curio-pdp.md
ZenGround0: No need for IPNI access for PDP at least not yet.  I don't think we are indexing pdp pieces into IPNI.  Though @magik6k can correct me.

We don't need the UI on the public internet either.  

I'll put in a callout.

ZenGround0: Ah ok I can't push to this branch so for the callout I would just say "If DNS is configured in the base layer then the HTTP section of the config can be dropped"

We could add it to the base layer its just slightly more complicated for little gain.

ZenGround0:
@LexLuthr I think this is ready for another review

LexLuthr:
I am merging this PR as is. Once merged, I can edit the page to align it more to Curio docs.

LexLuthr (approved):

PR #448, opened 2025-03-26
Fix/miner premium

ZenGround0:
So I think this fixes the premium bug well enough without moving ethclient over to the curio internal shared api.  

One funny thing is that it only reduces premium on calibnet by 10x instead of 100 - 1000x because so many people are running 1nFIL premiums that the median as returned by lotus api is really high.

### DIFF
--- a/tasks/message/sender_eth.go
+++ b/tasks/message/sender_eth.go
@@ -11,6 +11,8 @@ import (
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/filecoin-project/go-address"
+	lotustypes "github.com/filecoin-project/lotus/chain/types"
 	"go.uber.org/multierr"
 	"golang.org/x/xerrors"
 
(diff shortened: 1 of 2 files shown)


LexLuthr (approved):

Stebalien:
Can we not use `client.SuggestGasTipCap`? Actually, we probably just want `client.SuggestGasPrice` which includes both the base fee and the premium (which should correspond to the tip?).

Stebalien:
Ok, Forrest corrected me here.

The real issue is that curio runs lotus _in-process_ so it can't make eth calls with the eth library. Instead, it's making eth calls against the calibnet API. That's kind of insane.

Can we not just directly use the lotus API for all of this? As-in, directly call `Eth` methods?

tasks/message/sender_eth.go
```diff
@@ -298,10 +306,21 @@ func (s *SenderETH) Send(ctx context.Context, fromAddress common.Address, tx *ty
 			return common.Hash{}, fmt.Errorf("base fee not available; network might not support EIP-1559")
 		}
 
+		// TODO hack
+		// lotus api barely looks at anything we pass it here so make stuff up
+		dummyAddr, err := address.NewIDAddress(42)
```
Stebalien, line 311: Nit: I'd just use 0, the system actor.

ZenGround0:
I think I'll update this PR to configure the eth rpc api which makes things less of a mess.

PR #450, opened 2025-03-27
fix: store log debug, seal poller log, batching, itest

LexLuthr:
This PR mostly have small fixes.
1. Fix Commit and Precommit batch condition
2. Update StartEpoch poller condition
3. Improve storage logging
4. Improve seal poller logging
5. Drops TestCurioNewActor from itest. We already cover this in TestCurioHappyPath
6. Improves the harmonyTak retry logic.

### DIFF
--- a/tasks/seal/poller_precommit_msg.go
+++ b/tasks/seal/poller_precommit_msg.go
@@ -21,6 +21,10 @@ import (
 )
 
 func (s *SealPoller) pollStartBatchPrecommitMsg(ctx context.Context) {
+	if !s.pollers[pollerPrecommitMsg].IsSet() {
+		return
+	}
+
 	ts, err := s.api.ChainHead(ctx)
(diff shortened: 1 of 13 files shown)


tasks/seal/poller_precommit_msg.go
```diff
@@ -21,6 +21,10 @@ import (
 )
 
 func (s *SealPoller) pollStartBatchPrecommitMsg(ctx context.Context) {
+	if !s.pollers[pollerPrecommitMsg].IsSet() {
```
snadrus, line 24: add a comment as to how this could happen

snadrus (approved):

tasks/seal/poller_precommit_msg.go
LexLuthr: Done

PR #458, opened 2025-03-31
fix: bulk pipeline restart

LexLuthr:
The bulk restart fail on snap function with error `failed to resume SnapDeals sector: ERROR: FOREACH expression must not be null (SQLSTATE 22004)`

- [x] Test

### DIFF
--- a/harmony/harmonydb/sql/20250331-fix-bulk-restart-func.sql
+++ b/harmony/harmonydb/sql/20250331-fix-bulk-restart-func.sql
@@ -1,0 +1,67 @@
+CREATE OR REPLACE FUNCTION unset_task_id(sp_id_param bigint, sector_number_param bigint)
+    RETURNS void AS $$
+DECLARE
+column_name text;
+    column_names text[] := ARRAY[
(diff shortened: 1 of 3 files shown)


harmony/harmonydb/sql/20250331-fix-bulk-restart-func.sql
```diff
@@ -0,0 +1,67 @@
+CREATE OR REPLACE FUNCTION unset_task_id(sp_id_param bigint, sector_number_param bigint)
```
snadrus, line 1: Consider making unset_task_id_ext that takes column_names. Then these 2 functions can be a one-liner. It'll also cement their relationship if something needs to change in the future.

snadrus (approved):

harmony/harmonydb/sql/20250331-fix-bulk-restart-func.sql
LexLuthr: As discussed, this also needs the table name so keeping the PR as is.

PR #460, opened 2025-04-01
Fix panic in allocateToExisting

strahe:
Fixed the nil panic issue that occurs when using ddo, where accessing piece.DealProposal leads to a null pointer dereference.


LexLuthr (changes requested):
Actually, we can use the method PieceCID() for the piece lpiece.PieceDealInfo to directly get the pieceCID regardless of type of deal.

LexLuthr:
> Actually, we can use the method PieceCID() for the piece lpiece.PieceDealInfo to directly get the pieceCID regardless of type of deal.

@strahe https://github.com/filecoin-project/curio/blob/ac01fcbda3b22b4d2f44f9833b974e7a169e621f/market/storageingest/deal_ingest_snap.go#L509 is the code you need

LexLuthr (approved):

PR #461, opened 2025-04-02
update lotus version for devnet, undo menu drawer

LexLuthr:
- Update lotus version for devnet images
- Undo menu drawer, there are some glitches that we need to fix
- Update Github Action files

### DIFF
--- a/.github/image/Dockerfile
+++ b/.github/image/Dockerfile
@@ -23,7 +23,7 @@ RUN git submodule update --init
 RUN go mod download
 
 # Stage 2: Install Lotus binary
-FROM ghcr.io/filecoin-shipyard/lotus-containers:lotus-v1.32.0-rc3-devnet AS lotus-test
+FROM ghcr.io/filecoin-shipyard/lotus-containers:lotus-v1.32.1-devnet AS lotus-test
 
 # Stage 3: Build the final image
 FROM myoung34/github-runner AS curio-github-runner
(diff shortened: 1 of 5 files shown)


.github/image/Dockerfile
```diff
@@ -23,7 +23,7 @@ RUN git submodule update --init
 RUN go mod download
 
 # Stage 2: Install Lotus binary
-FROM ghcr.io/filecoin-shipyard/lotus-containers:lotus-v1.32.0-rc3-devnet AS lotus-test
+FROM ghcr.io/filecoin-shipyard/lotus-containers:lotus-v1.32.1-devnet AS lotus-test
```
rjan90, line 26: Just noting as I came by and checked the Lotus version in this PR:

There was a v1.32.1 patch release published on Friday: https://github.com/filecoin-project/lotus/releases/tag/v1.32.1 which correctly set the F3 activation contract address on Mainnet, which v1.32.0 did not have. 

The dev-version was because of that patch release bumped to v1.32.2-dev: https://github.com/filecoin-project/lotus/blob/062ce4d52118a72512ace821744251b2ff5c5200/build/version.go#L9-L10

rjan90: But I might be misunderstanding the meaning of -dev in this change though, and that this is related to how Curio is handling things internally.

snadrus (approved):

PR #462, opened 2025-04-04
wdpost should cancel if deadline is closed

LexLuthr:


tasks/window/compute_task.go
```diff
@@ -175,25 +176,66 @@ func (t *WdPostTask) Do(taskID harmonytask.TaskID, stillOwned func() bool) (done
...
+
+		for {
+			select {
+			case <-finish:
+				return
+			case <-ticker.C:
+				h, err := stAPI.ChainHead(context.Background())
+				if err != nil {
+					counter++
+					if counter < 5 {
```
snadrus, line 208: Can we compute this constant?

snadrus (commented):
Let's have magic look over it.

tasks/window/compute_task.go
LexLuthr: Which constant?

tasks/window/compute_task.go
```diff
@@ -175,25 +176,66 @@ func (t *WdPostTask) Do(taskID harmonytask.TaskID, stillOwned func() bool) (done
...
+						log.Warnf("WdPostTask.Do() SP %d on Deadline %d and Partition %d failed to get chain head: %w", spID, dlIdx, partIdx, err)
+					}
+					cancel(xerrors.Errorf("WdPostTask.Do() SP %d on Deadline %d and Partition %d failed to get chain head: %w", spID, dlIdx, partIdx, err))
+					return
```
magik6k, line 212: Quite sure we don't want to cancel if the API happens to be down for 5 seconds, maybe could do 5 minutes of grace.

tasks/window/compute_task.go
```diff
@@ -175,25 +176,66 @@ func (t *WdPostTask) Do(taskID harmonytask.TaskID, stillOwned func() bool) (done
...
+				counter = 0
+				if h.Height() > deadline.Challenge {
+					cancel(xerrors.Errorf("WdPostTask.Do() SP %d on Deadline %d and Partition %d cancelling context as head %d is greater then deadline close %d", spID, dlIdx, partIdx, h.Height(), deadline.Close))
+					return
+				}
+			}
+		}
+	}(t.api, cancel, finish)
+
+	postOut, err := t.DoPartition(ctx, ts, maddr, deadline, partIdx, isTestTask())
```
magik6k, line 223: Do we propagate this context to ffiselect when computing the snark? Or do we only care about cancelling data reading?

magik6k (commented):
I like this idea, just one nit + question

tasks/window/compute_task.go
LexLuthr: That makes sense. I will bump up the grace period.

tasks/window/compute_task.go
LexLuthr: Looking at the FFISelect there is no context on rust side to we have no real way to cancel that context. On FFISelect RPC layer, the context will already be cancelled as we are cancelling the parent context itself.

magik6k: The idea would be that ffiselect just kills the child process when the context is cancelled.

PR #463, opened 2025-04-04
fix: UI sector list for expired sectors

LexLuthr:


web/api/sector/sector.go
```diff
@@ -199,7 +199,9 @@ func (c *cfg) getSectors(w http.ResponseWriter, r *http.Request) {
 				}
 				sectors[i].Proving = chainy.active
 				if st.Expiration < head.Height() {
-					sectors[i].Flag = true // Flag expired sectors
+					delete(sectorIdx, sectorID{minerID, uint64(st.SectorNumber)})
```
snadrus, line 202: Delete in a loop can be very inefficient.  But it's probably fine here if there's not many.

snadrus (approved):

web/api/sector/sector.go
magik6k: Yeah, https://github.com/filecoin-project/curio/pull/364 has an attempt at pagination, still not perfect tho.

magik6k (approved):
This is better, sooo on a boat it goes!

PR #465, opened 2025-04-06
Do the correct thing

ZenGround0:
I ignored the nice build flag stuff earlier, now curio will work with correct contract immediately -- nice


ZenGround0:
Merging quick to unblock myself

PR #469, opened 2025-04-10
http based 12 deals

LexLuthr:


market/libp2p/libp2p.go
```diff
@@ -100,6 +96,8 @@ func NewLibp2pHost(ctx context.Context, db *harmonydb.DB, cfg *config.CurioConfi
 		libp2p.EnableNATService(),
 		libp2p.BandwidthReporter(metrics.NewBandwidthCounter()),
 		libp2p.Identity(lcfg.priv),
+		libp2p.ResourceManager(&network.NullResourceManager{}),
+		libp2p.ConnectionManager(&connmgr.NullConnMgr{}),
```
magik6k, line 100: Really not ideal, we should have some config which works with some reasonable settings, but still better than the current completely broken one

magik6k (approved):
Looks great, really like how clean this turned out!

PR #480, opened 2025-04-16
feat: allow clients to pass extra-data for proof set creation and adding roots

aarshkshah1992:
The `PDPVerifier` contract allows passing an opaque byte array as "extradata" that gets passed to the listener as is. This is needed for allowing the client/provider to pass in payer/payee addresses and proofset/root metadata to the listener to build payment flows/better explorer UX.

### DIFF
--- a/cmd/pdptool/main.go
+++ b/cmd/pdptool/main.go
@@ -35,6 +35,21 @@ import (
 	curioproof "github.com/filecoin-project/curio/lib/proof"
 )
 
+// validateExtraData checks if the provided hex string is valid and within the size limit.
+func validateExtraData(extraDataHexStr string) error {
+	if extraDataHexStr == "" {
+		return nil // No data to validate
+	}
+	decoded, err := hex.DecodeString(strings.TrimPrefix(extraDataHexStr, "0x"))
+	if err != nil {
+		return fmt.Errorf("failed to decode hex in extra-data: %w", err)
+	}
+	if len(decoded) > 2048 {
+		return fmt.Errorf("decoded extra-data exceeds maximum size of 2048 bytes (decoded length: %d)", len(decoded))
+	}
+	return nil
+}
+
 func main() {
 	app := &cli.App{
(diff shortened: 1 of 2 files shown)


cmd/pdptool/main.go
```diff
@@ -1140,20 +1161,37 @@ var addRootsCmd = &cli.Command{
...
 	Action: func(cctx *cli.Context) error {
 		serviceURL := cctx.String("service-url")
 		serviceName := cctx.String("service-name")
 		proofSetID := cctx.Uint64("proof-set-id")
 		rootInputs := cctx.StringSlice("root")
+		extraDataHexStr := cctx.String("extra-data")
 
-		// Load the private key (implement `loadPrivateKey` according to your setup)
+		// Validate extraData hex string and its decoded length
+		if extraDataHexStr != "" {
```
ZenGround0, line 1178: nit: extract to a small validation and decode function since we use this code a bunch

pdp/handlers.go
```diff
@@ -575,22 +591,42 @@ func (p *PDPService) handleAddRootToProofSet(w http.ResponseWriter, r *http.Requ
...
+		extraDataHexStr := *payload.ExtraData
+		decodedBytes, err := hex.DecodeString(strings.TrimPrefix(extraDataHexStr, "0x"))
+		if err != nil {
+			log.Errorf("Failed to decode hex extraData: %v", err)
+			http.Error(w, "Invalid extraData format (must be hex encoded)", http.StatusBadRequest)
+			return
+		}
+		extraDataBytes = decodedBytes
+	} else {
+		extraDataBytes = []byte{} // Ensure it's an empty byte slice if not provided
```
ZenGround0, line 624: nit: just initialize extraDataBytes := []byte{} on line 613.  Then we can avoid the else statement

ZenGround0 (approved):
Looks good.  Will update my node to this and test it out live.

PR #483, opened 2025-04-17
PDP only cluster init

LexLuthr:
- [x] New PDP only cluster with `guided-setup`
- [x] Update `curio config new-cluster` for PDP only cluster
- [ ] Wait for PDP to merge
- [ ] Find a way to deploy PDP contracts on devnet 
- [ ] Stitch together PDP only devnet
- [ ] Verify that Curio no longer requires minerID
- [ ] Update web menu for PDP only cluster

### DIFF
--- a/cmd/curio/config_new.go
+++ b/cmd/curio/config_new.go
@@ -25,9 +25,13 @@ var configNewCmd = &cli.Command{
 			Hidden:  true,
 			Value:   "~/.lotus",
 		},
+		&cli.BoolFlag{
+			Name:  "pdp",
+			Usage: "initiate a PDP only cluster",
+		},
 	},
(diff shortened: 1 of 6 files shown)


cmd/curio/config_new.go
```diff
@@ -25,9 +25,13 @@ var configNewCmd = &cli.Command{
 			Hidden:  true,
 			Value:   "~/.lotus",
 		},
+		&cli.BoolFlag{
+			Name:  "pdp",
```
magik6k, line 29: Maybe `empty`? This option will apply to more than just pdp

ZenGround0:
Has feat/pdp merged?  Should this be based off of main?

PR #488, opened 2025-04-22
chore: update pdp docs

rjan90:
Running `make pdptool` fails, updating to be the correct checkout steps for a client


LexLuthr (changes requested):
I have highlighted this twice. Author needs to create the make command for pdptool. Curio does not support build commands outside `make`.

PR #490, opened 2025-04-22
feat: pending messages in UI

LexLuthr:
This PR adds pending message table to wallets page.

The color of the message string changes based on following:
1. < 30 minutes: Green
2. > 30 and < 60 minutes: Yellow
3. > 60 minutes: Red

Alert is also generates for messages > 60 minutes.

Once this PR is merged, I merge main into PDP and extend the functionality.

Screenshot 2025-04-22 at 4 40 41 PM
Screenshot 2025-04-22 at 4 44 51 PM
Screenshot 2025-04-22 at 5 33 45 PM

### DIFF
--- a/harmony/harmonydb/sql/20250422-msg-wait-timestamp.sql
+++ b/harmony/harmonydb/sql/20250422-msg-wait-timestamp.sql
@@ -1,0 +1,2 @@
+ALTER TABLE message_waits
+    ADD COLUMN added_at timestamptz NOT NULL DEFAULT TIMEZONE('UTC', NOW());
--- a/web/api/webrpc/wallet.go
+++ b/web/api/webrpc/wallet.go
@@ -4,6 +4,9 @@ import (
 	"context"
 	"errors"
 	"sync"
+	"time"
+
+	"golang.org/x/xerrors"
 
 	"github.com/filecoin-project/curio/harmony/harmonydb"
 )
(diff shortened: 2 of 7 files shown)


web/api/webrpc/wallet.go
```diff
@@ -99,3 +102,43 @@ func (a *WebRPC) WalletRemove(ctx context.Context, wallet string) error {
...
+}
+
+func (a *WebRPC) PendingMessages(ctx context.Context) (PendingMessages, error) {
+
+	var messages []struct {
+		MessageCid string    `db:"signed_message_cid"`
+		AddedAt    time.Time `db:"added_at"`
+	}
+
+	err := a.deps.DB.Select(ctx, &messages, `SELECT signed_message_cid, added_at FROM message_waits WHERE executed_tsk_cid IS NULL ORDER BY added_at DESC`)
```
magik6k, line 123: Not `ORDER BY created_at DESC`?

harmony/harmonydb/sql/20250422-msg-wait-timestamp.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE message_waits
+    ADD COLUMN added_at timestamptz NOT NULL DEFAULT TIMEZONE('UTC', NOW());
```
magik6k, line 2: The waits table holds historic data, we want a fancy conditional index on `created_at` + `where executed_tsk_cid is not null`

magik6k: https://www.postgresql.org/docs/current/indexes-partial.html

magik6k (commented):
Couple small nits, looks super useful.

Potentially later we can share some of this with the AlertManager

LexLuthr:
> Couple small nits, looks super useful.
> 
> Potentially later we can share some of this with the AlertManager

It is already plugged into AlertManager in this PR. If you are talking about generating alert from UI and passing it to AlertManager then I did not use the method by choice. Currently, AlertManager only care about what is in DB and if we use UI to generate the alert then we will depend on someone clicking in that page to actually generate the alert.

harmony/harmonydb/sql/20250422-msg-wait-timestamp.sql
LexLuthr: Forgot that we don't clean up this one

Reiers (approved):

magik6k (approved):

Issue #489, opened 2025-04-22
Default to use aggregation with prove commits to save significant gas usage


jennijuju:
### Checklist

- [x] This is **not** a new feature or an enhancement to the Filecoin protocol. If it is,  please open an [FIP issue](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0001.md).
- [x] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the Curio forum](https://github.com/filecoin-project/curio/discussions/categories/ideas) and select the category as `Ideas`.
- [x] I **have** a specific, actionable, and well motivated feature request to propose.

### Curio component

- [x] PoRep Pipeline
- [ ] Snap Deals
- [ ] Market
- [ ] IPNI provider
- [ ] Curio UI
- [ ] Other

### What is the motivation behind this feature request? Is your feature request related to a problem? Please describe.

FIP-100 removed the batch balancer to incentivize aggregation of prove commits to be always economically rational - it **always** costs way less gas to onboard the same amount of sectors using aggregation compared to onboard sectors individually. Similarly with batching for precommits.

<img width="598" alt="Image" src="https://github.com/user-attachments/assets/cc45ea60-a33c-4c1d-8385-0ddbf0e549aa" />

<img width="582" alt="Image" src="https://github.com/user-attachments/assets/842db1e2-8a44-4850-88da-d3c0874f4a7c" />

### Describe the solution you'd like

simplify the onboarding pipeline to default using batching and aggregation, and remove any fire-message logic thats based on base fee (user might still want to config the max # sectors and/or how long to wait for aggregation.

LexLuthr:
For implementor

- [ ] Remove base fee based logic for aggregation
    - [ ] Rewrite SQL functions
    - [ ] Test in devnet

PR #492, opened 2025-04-22
no fee based aggregation/batching

LexLuthr:
fixes https://github.com/filecoin-project/curio/issues/489

Should not break nodes running older code during upgrade

- [x] Test in devnet

### DIFF
--- a/deps/config/load.go
+++ b/deps/config/load.go
@@ -10,6 +10,7 @@ import (
 	"regexp"
 	"sort"
 	"strings"
+	"time"
 	"unicode"
 
 	"github.com/BurntSushi/toml"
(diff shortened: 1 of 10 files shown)


snadrus (approved):
fancy! I'd like Magik to check the SQL logic

LexLuthr:
> fancy! I'd like Magik to check the SQL logic

It is basically the same function but we don't check for fee anymore.

deps/config/load.go
magik6k, line 1: What do the changes here do?

harmony/harmonydb/sql/20250423-remove-fee-aggregation.sql
```diff
@@ -0,0 +1,335 @@
...
+            sectors_sdr_initial_pieces or fallback.
+        (3) Numbers them per group, forming sub-batches of size p_max_batch.
+        (4) Groups by (sp_id, reg_seal_proof, batch_index).
+        (5) Finally selects sp_id, reg_seal_proof, an array of sector_numbers,
+            plus the min start_epoch and earliest_ready for each batch.
+      Then we loop over those grouped results in ascending order.
+    */
+
+    FOR batch_rec IN
+        WITH locked AS (
```
magik6k, line 242: My understanding is that `WITH locked AS` does nothing at all in yugabyte

magik6k (commented):
Just the question about the config stuff. The rest looks good / makes sense.

deps/config/load.go
LexLuthr: Oh! Those are mostly due to how comp library works. It treats "0h5m10s" different from "5h0m0s" for some reason. 
The other change is because if we have 2 config parameters with same name then current load confuse the value because map we created was global. This change updates the map keys to be config section specific. To explain is simply:
The preCommit msg timeout of 1 hour (default 4 hour) will no longer be treated as default value due to commit msg timeout default being 1 hour.

harmony/harmonydb/sql/20250423-remove-fee-aggregation.sql
LexLuthr: I think you are confusing the table name with "FOR UPDATE;" at the end for actual locking. I can change the table name to avoid confusing.

magik6k: Yaaah it was 11 PM :sweat_smile:

magik6k (approved):

PR #498, opened 2025-04-24
pdp browser rendering

LexLuthr:
## DO NOT MERGE
- [x] Explore client side security considerations
1. Allows malicious builders to use PDP provider has host to malicious content
2. If malicious script is flagged, PDP provider's domain will be hit
3. Locking down CORS will mean external builders cannot serve directly from PDP provider
4. How do we find a balance? Where do we draw the line?

Added server side sniffing and only do "inline" if a valid MIME type is detected. This handles the security risks upto some degree.

### DIFF
--- a/market/retrieval/piecehandler.go
+++ b/market/retrieval/piecehandler.go
@@ -4,7 +4,9 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"mime"
 	"net/http"
+	"net/url"
 	"time"
 
 	"github.com/ipfs/go-cid"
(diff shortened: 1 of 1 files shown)


market/retrieval/piecehandler.go
```diff
@@ -58,19 +60,46 @@ func (rp *Provider) handleByPieceCid(w http.ResponseWriter, r *http.Request) {
...
+				encoded := url.PathEscape(filename)
+
+				w.Header().Set("Content-Disposition",
+					fmt.Sprintf(`inline; filename="%s"; filename*=UTF-8''%s`, filename, encoded))
+			}
+		}
+	} else {
+		etag := `"` + pieceCid.String() + `.gz"` // must be quoted
+		w.Header().Set("Etag", etag)
+		w.Header().Set("Content-Type", "application/piece")
```
magik6k, line 101: That's not really the standard way to do mime, I'd say either the .car mime, or do octet-stream. (I know it's what we used to do, but it's still wrong)

market/retrieval/piecehandler.go
```diff
@@ -58,19 +60,46 @@ func (rp *Provider) handleByPieceCid(w http.ResponseWriter, r *http.Request) {
...
 }
 
-func setHeaders(w http.ResponseWriter, pieceCid cid.Cid) {
+func setHeaders(w http.ResponseWriter, pieceCid cid.Cid, pdpPiece bool, contentType string) {
 	w.Header().Set("Vary", "Accept-Encoding")
-	etag := `"` + pieceCid.String() + `.gz"` // must be quoted
-	w.Header().Set("Etag", etag)
-	w.Header().Set("Content-Type", "application/piece")
 	w.Header().Set("Cache-Control", "public, max-age=29030400, immutable")
+	if pdpPiece {
```
magik6k, line 86: I wouldn't differentiate, piece is a piece.

Thinking in market v2 terms:
* Pieces are mostly separate to how they are proven on-chain
* Piece owners will be able to define some policy like ACL on a piece, but it could also include content-type settings
* Who a piece-owner is is a market v2 concern, so for now here we just have a single common default, always probing type.

Unless something would break, then this would make more sense.

LexLuthr: That makes sense. I will update the PR once testing for browser rendering is done

market/retrieval/piecehandler.go
LexLuthr: I would prefer to use "application/octet-stream" instead of car to align more with HTTP standards instead if Filecoin. A piece is in the end a byte array and it might not be a car file in market v2.

magik6k (approved):

PR #501, opened 2025-04-27
feat: add upload folder command and improvs

lordshashank:
- Implements `uploadFolder` CLI command, which recursively uploads all the files in folders to pdpservice.
- Implements optional parameter of `valid-time` for jwt creation and defaults it to 24 hours.
- Implements option `addRoots` parameter to `uploadFile` and `uploadFolder` command to upload and add root of particular file/folder in one go rather than running separate commands. 
- Refactors code to use functions for repetitive tasks.

Future improvs:-
- multiple root additions in single `POST` call to pdp-miner has some bugs i.e. is not working properly, hence currently makes multiple request for each roots.


cmd/pdptool/main.go
```diff
@@ -869,20 +1061,18 @@ var createProofSetCmd = &cli.Command{
...
 		recordKeeper := cctx.String("recordkeeper")
 		extraDataHexStr := cctx.String("extra-data")
-
 		// Validate extraData hex string and its decoded length
 		if err := validateExtraData(extraDataHexStr); err != nil {
 			return err
 		}
-
-		// Load the private key
+		// Load the private key (implement `loadPrivateKey` according to your setup)
```
LexLuthr, line 1068: Remove AI comments

cmd/pdptool/main.go
```diff
@@ -138,6 +361,13 @@ var authCreateJWTTokenCmd = &cli.Command{
 	Name:      "create-jwt-token",
 	Usage:     "Generate a JWT token using the service secret",
 	ArgsUsage: "[service_name]",
+	Flags: []cli.Flag{
+		&cli.IntFlag{
+			Name:  "valid-time",
+			Usage: "Validity time of the JWT token in hours",
+			Value: 24, // Default to 1 day
+		},
+	},
```
LexLuthr, line 370: Wouldn't this always set the JWT validity to 24 hours by default? I am not sure how PDP works but this means that JWT token handed off to SP will no longer be valid after 24 hours. How will client check anything after that?

lordshashank: I also don't know, but i created one for month and it is rendering me data until now.

PR #505, opened 2025-05-02
Making upgrading safe

cryptowhizzard:


LexLuthr:
This PR only handles a single table. Either we wrap all the SQL commands like this or continue with current approach. After discussing with @snadrus, it would be better to just continue with current approach. Closing the PR as not required.

PR #508, opened 2025-05-14
[WIP] feat: market 2.0

LexLuthr:


tasks/storage-market/mk20.go
```diff
@@ -0,0 +1,1085 @@
...
+			if piece.SourceOffline != nil {
+				offlinelist[downloadkey{ID: dealID, PieceCID: piece.PieceCID, Size: piece.Size}] = struct{}{}
+			}
+		}
+
+		pqBatch := &pgx.Batch{}
+		pqBatchSize := 20000
+
+		for k, _ := range toDownload {
+			pqBatch.Queue(`SELECT id FROM parked_pieces WHERE piece_cid = $1 AND piece_padded_size = $2`, k.PieceCID.String(), int64(k.Size)).QueryRow(func(row pgx.Row) error {
```
snadrus, line 305: Here's a hack to make this 1 query total (no loop) with half the params:
SELECT id FROM parked_pieces WHERE CONCAT(piece_cid, STRING(piece_padded_size)) = ANY($1)

tasks/storage-market/mk20.go
```diff
@@ -0,0 +1,1085 @@
...
+			continue
+		}
+		if !comm {
+			log.Errorf("inserting deal in pipeline: commit failed")
+			continue
+		}
+	}
+}
+
+func insertPiecesInTransaction(ctx context.Context, tx *harmonydb.Tx, deal *mk20.Deal) error {
```
snadrus, line 162: consider adding batch-with-flush behavior, maybe keyed on the length of args list staying low (10,000)

tasks/storage-market/mk20.go
```diff
@@ -0,0 +1,1085 @@
...
+		aggregation = data.Format.Aggregate.Type
+	} else {
+		aggregation = nil
+	}
+
+	// Insert pipeline when Data source is HTTP
+	if data.SourceHTTP != nil {
+		var pieceID int64
+		// Attempt to select the piece ID first
+		err = tx.QueryRow(`SELECT id FROM parked_pieces WHERE piece_cid = $1 AND piece_padded_size = $2`, data.PieceCID.String(), data.Size).Scan(&pieceID)
```
snadrus, line 190: consider an UPSERT

market/mk20/types.go
```diff
@@ -0,0 +1,244 @@
...
+	name := string(dso.Name())
+	for _, p := range dbDataSources {
+		if p.Name == name {
+			if p.Enabled {
+				return Ok, nil
+			}
+		}
+	}
+	return ErrUnsupportedDataSource, xerrors.Errorf("data source %s is not enabled", name)
+}
```
magik6k: Imho this method shouldn't really exist, instead we should have a `Dependencies() map[Name]struct{}` which returns self + in case of aggregate the sub types.

market/mk20/types.go
```diff
@@ -0,0 +1,244 @@
...
+
+// HttpUrl represents an HTTP endpoint configuration for fetching piece data.
+type HttpUrl struct {
+
+	// URL specifies the HTTP endpoint where the piece data can be fetched.
+	URL string `json:"url"`
+
+	// HTTPHeaders represents the HTTP headers associated with the URL.
+	HTTPHeaders http.Header `json:"httpheaders"`
```
magik6k, line 167: The struct is already named http, no need for that in variable names

```suggestion
	Headers http.Header `json:"headers"`
```

market/mk20/ddo_v1.go
```diff
@@ -0,0 +1,191 @@
...
+// for a DDO deal handling.
+type DDOV1 struct {
+
+	// Provider specifies the address of the provider
+	Provider address.Address `json:"provider"`
+
+	// Client represents the address of the deal client
+	Client address.Address `json:"client"`
+
+	// Actor able to with AuthorizeMessage (like f1/f3 wallet) able to authorize actions such as managing ACLs
```
magik6k, line 36:
```suggestion
	// Actor providing AuthorizeMessage (like f1/f3 wallet) able to authorize actions such as managing ACLs
```

market/mk20/ddo_v1.go
```diff
@@ -0,0 +1,191 @@
...
+	PieceManager address.Address `json:"piecemanager"`
+
+	// Duration represents the deal duration in epochs. This value is ignored for the deal with allocationID.
+	// It must be at least 518400
+	Duration abi.ChainEpoch `json:"duration"`
+
+	// AllocationId represents an aggregated allocation identifier for the deal.
+	AllocationId *verifreg.AllocationId `json:"aggregatedallocationid"`
```
magik6k, line 44: Lets make the fields not look like German.

```suggestion
	AllocationId *verifreg.AllocationId `json:"aggregated_allocation_id"`
```

Also not sure if here we really need the aggregated_ prefix

market/mk20/ddo_v1.go
```diff
@@ -0,0 +1,191 @@
...
+
+	// ContractAddress specifies the address of the contract governing the deal
+	ContractAddress string `json:"contractaddress"`
+
+	// ContractDealIDMethod specifies the method name to retrieve the deal ID for a contract
+	ContractDealIDMethod string `json:"contractdealidmethod"`
```
magik6k, line 50: The deal ID method is also where we do the first off-chain verification? Personally I'd call that method 'Verify' as duplicate id check is really just part of verification

magik6k (commented):
Just reviewing types for now, didn't look at much beyond that, but generally things make sense.

LexLuthr:
## Refactor

- [ ] Refactor Deal{} struct
- [ ] Figure out where to add market deal ID for PoRep
- [ ] Add PDP product
- [ ] Allow uploading data before data details
- [ ] Refactor chunk aggregation to work for PDP
- [ ] Set a limit to delete data if not finalized in certain time
- [ ] Create PDP pipeline
- [ ] Create method to update deal
- [ ] Update DB deal table to have complete, error columns for each product
- [ ] Create a clean up task for indexing/ipni based on expired sectors and completed PDP deals (deleteRoot)
- [ ] Refactor UI including PDP
- [ ] Remove unused PDP endpoints
- [ ] Move mk20 endpoints to directly under market
- [ ] Rename tables as required to differentiate between PoRep and PDP if required
- [ ] Figure out indexing when PoRep is later added to PDP deals (market piece deal table)
- [ ] Update retrievals to use market piece deal table instead of PDP workarounds. 
- [ ] Figure out IPNI piece only announcement
- [ ] Figure out IPNI announcements when PoRep is added to PDP deal
- [ ] Should PDP support other data sources? Especially aggregate (PODSI) data?

snadrus (changes requested):

tasks/pdp/task_prove.go
```diff
@@ -382,258 +384,170 @@ func padTo32Bytes(b []byte) []byte {
...
+			return contract.PDPVerifierProof{}, xerrors.Errorf("failed to get node: %w", err)
+		}
 
-	for i := len(subroots) - 1; i >= 0; i-- {
-		subroot := subroots[i]
-		level := proof.NodeLevel(subroot.SubrootSize/LeafSize, arity)
-		offset := (subroot.SubrootOffset / LeafSize) >> uint(level-1)
-		firstSubroot := i == 0
+		if !has {
+			// TODO: Trigger a Layer save task here and figure out if we should proceed or not
```
snadrus, line 454: timeline?

tasks/pdp/task_save_cache.go
```diff
@@ -0,0 +1,653 @@
...
+	}
+	cp.layerQueues[myIdx+1] = make(chan []byte, layerQueueDepth)
+	collectSnapshot := int(myIdx) == cp.snapShotLayerIdx-1
+	go func() {
+		var twinHold []byte
+
+		for {
+			slab, queueIsOpen := <-cp.layerQueues[myIdx]
+
+			// the dream is collapsing
```
snadrus, line 492: ?

web/api/webrpc/market.go
```diff
@@ -171,18 +187,19 @@ func (a *WebRPC) GetDealPipelines(ctx context.Context, limit int, offset int) ([
...
 	CreatedAt         time.Time      `db:"created_at" json:"created_at"`
 	SignedProposalCid string         `db:"signed_proposal_cid" json:"signed_proposal_cid"`
 	Offline           bool           `db:"offline" json:"offline"`
 	Verified          bool           `db:"verified" json:"verified"`
 	StartEpoch        int64          `db:"start_epoch" json:"start_epoch"`
 	EndEpoch          int64          `db:"end_epoch" json:"end_epoch"`
 	ClientPeerId      string         `db:"client_peer_id" json:"client_peer_id"`
-	ChainDealId       *int64         `db:"chain_deal_id" json:"chain_deal_id"`
-	PublishCid        *string        `db:"publish_cid" json:"publish_cid"`
+	ChainDealId       sql.NullInt64  `db:"chain_deal_id" json:"chain_deal_id"`
```
snadrus, line 198: if these are obnoxious, you can use in sql: COALESCE(chain_deal_id, 0) to get a zero value instead.

web/api/webrpc/market.go
```diff
@@ -417,39 +395,66 @@ func (a *WebRPC) MK12StorageDealList(ctx context.Context, limit int, offset int)
...
+			}
+			mk12Summaries[i].PieceCidV2 = pcid2.String()
 		}
-		mk12Summaries[i].Miner = addr.String()
 	}
 	return mk12Summaries, nil
+
 }
 
+// LegacyStorageDealList is deprecated
```
snadrus, line 417: rm?

market/mk20/pdp_v1.go
```diff
@@ -0,0 +1,138 @@
...
+	code, err := IsProductEnabled(db, p.ProductName())
+	if err != nil {
+		return code, err
+	}
+
+	if ok := p.CreateProofSet || p.DeleteProofSet || p.AddRoot || p.DeleteRoot; !ok {
+		return ErrBadProposal, xerrors.Errorf("deal must have one of the following flags set: create_proof_set, delete_proof_set, add_root, delete_root")
+	}
+
+	if p.CreateProofSet {
```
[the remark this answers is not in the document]
LexLuthr: Rename changes pushed.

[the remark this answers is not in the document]
LexLuthr: I will try to test in devnet again with latest contract but old service contract so verify that basic functionality is intact is after rename and I haven't broken things too badly. But, as this point, Synapse integration should be unblocked.

cmd/sptool/toolbox_deal_client.go
```diff
@@ -1561,3 +1569,1220 @@ var dealStatusCmd = &cli.Command{
...
+			if hurl.Scheme == "wss" {
+				hurl.Scheme = "https"
+			}
+			log.Debugw("converted multiaddr to URL", "url", hurl, "multiaddr", ma.String())
+			hurls = append(hurls, hurl)
+		}
+
+		addRoot := cctx.Bool("add-root")
+		addProofset := cctx.Bool("add-proofset")
+		removeRoot := cctx.Bool("remove-root")
```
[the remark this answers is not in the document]
ZenGround0: The core fields on `mk20.PDPV1` look correct, it's just the UX at the sptool level that is still using old names

market/mk20/types.go
```diff
@@ -0,0 +1,277 @@
...
+	// SourceHTTP represents the HTTP-based source of piece data within a deal, including raw size and URLs for retrieval.
+	SourceHTTP *DataSourceHTTP `json:"source_http,omitempty"`
+
+	// SourceAggregate represents an aggregated source, comprising multiple data sources as pieces.
+	SourceAggregate *DataSourceAggregate `json:"source_aggregate,omitempty"`
+
+	// SourceOffline defines the data source for offline pieces, including raw size information.
+	SourceOffline *DataSourceOffline `json:"source_offline,omitempty"`
+
+	// SourceHTTPPut // allow clients to push piece data after deal accepted, sort of like offline import
```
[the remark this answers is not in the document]
LexLuthr: Bad comment. You can make deal first and then upload data.

web/api/webrpc/market.go
LexLuthr: 0 is a valid chain deal ID

tasks/pdp/data_set_create_watch.go
```diff
@@ -213,16 +213,10 @@ func getProvingPeriodChallengeWindow(ctx context.Context, ethClient *ethclient.C
 		return 0, 0, xerrors.Errorf("failed to create proving schedule binding, check that listener has proving schedule methods: %w", err)
 	}
 
-	period, err := schedule.GetMaxProvingPeriod(&bind.CallOpts{Context: ctx})
+	config, err := schedule.GetPDPConfig(&bind.CallOpts{Context: ctx})
```
rvagg, line 216: @LexLuthr this is also going to need this, to optionally (I made it optional for simplicity but it could just as well be mandator) indirection through to a view (read-only) contract: https://github.com/filecoin-project/curio/pull/614

LexLuthr: @rvagg Can you please give me some context to these changes? I don't understand why we are doing it this way.

LexLuthr: Verifier contract has no such function.

rvagg: It sucks but it's a contract size limitation. You can only deploy contracts up to about 24k or some number and we quickly found ourselves at that limit. Every public function on a contract bloats the size. So we ended up dealing with it by treating WarmStorage as a "write" contract and then having a second contract off the size that's a "read" contract (FilecoinWarmStorageStateView, there's also a Library which can be used to read the same things but from another contract). It's able to reach into the state slots of the write contract and provide whatever methods we want, so we now have plenty of space to be able to make useful accessors while isolating the important functionality to the main service / write contract.

To make this practical, we have a double linking. The view contract has the address of the write contract because it needs that to do state loads. And then we also added the view contract's address to the write contract because we don't want to have to pass around multiple contract addresses or use the view contract as our primary contract address. And we have the additional complication that the write contract is the one that PDPVerifier uses as a listener, so we didn't want to do the indirection in there.

SO, for read methods that are typically done with an eth_call, we put them on the view contract, this is the rule going forward. To get to them you call `viewContractAddress()` and jump over to it. The pattern I introduced in the above commit was to first test if that function exists and move over to it for our read operations if it does. So, now as far as the relationship between client, Curio and PDPVerifier is concerned, PDPVerifier does the proving and WarmStorage does the listening. For everything else that the client or curio cares about, it has to go and ask the view contract.

LexLuthr:
# PDP pipeline

Screenshot 2025-09-09 at 6 32 56 PM

market/mk20/utils.go
```diff
@@ -0,0 +1,1018 @@
...
+		return "", nil, nil, fmt.Errorf("invalid signature")
+	}
+
+	return keyType, pubKey, sig, nil
+}
+
+func verifySignature(db *harmonydb.DB, keyType string, pubKey, signature []byte, cfg *config.CurioConfig) (bool, string, error) {
+	now := time.Now().Truncate(time.Hour)
+	minus1 := now.Add(-59 * time.Minute)
+	plus1 := now.Add(59 * time.Minute)
```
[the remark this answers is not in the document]
LexLuthr: We can change it to 1 hour. I should be fine. This was just to limit the time for testing period. Since, we are planning to switch to session keys, this would hold less significance as timestamp signing will move to original client signing the Auth for session key side. It allows us to allow auth with some expiry period.

market/mk20/utils.go
```diff
@@ -0,0 +1,1018 @@
...
+	return allowed, nil
+}
+
+func clientAllowed(ctx context.Context, db *harmonydb.DB, client string, cfg *config.CurioConfig) (bool, error) {
+	if !cfg.Market.StorageMarketConfig.MK20.DenyUnknownClients {
+		return true, nil
+	}
+
+	var allowed bool
+	err := db.QueryRow(ctx, `SELECT EXISTS (SELECT 1 FROM market_mk20_clients WHERE client = $1 AND allowed = TRUE)`, client).Scan(&allowed)
```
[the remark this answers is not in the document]
LexLuthr: List is flexible. You can mark a client allow/deny without separate lists.

market/mk20/utils.go
```diff
@@ -0,0 +1,1018 @@
...
+	return verifySignature(db, keyType, pubKey, sig, cfg)
+}
+
+func parseCustomAuth(header string) (keyType string, pubKey, sig []byte, err error) {
+
+	if !strings.HasPrefix(header, Authprefix) {
+		return "", nil, nil, errors.New("missing CustomAuth prefix")
+	}
+
+	parts := strings.SplitN(strings.TrimPrefix(header, Authprefix), ":", 3)
```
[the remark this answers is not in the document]
LexLuthr: We can switch. Again, this is just basic Auth for testing purposes.

market/mk20/utils.go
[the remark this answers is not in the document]
wjmelements: I am not saying you need separate lists. I am saying that this code is only a blacklist and does not function as a whitelist.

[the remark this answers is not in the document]
LexLuthr: Thanks! Yes, a logical bug.

documentation/en/market-2.0/README.md
```diff
@@ -0,0 +1,192 @@
...
+* `ddov1`: governs how the data should be stored and verified
+* (future) `aclv1`: may define retrieval access controls
+* (future) `retrievalv1`: may define retrieval SLA or payment terms
+
+### 🛠 Smart Contract Control
+
+Clients must select a contract that:
+
+* Is supported by the SP
+* Implements a deal validation method that returns a valid DealID
```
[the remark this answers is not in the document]
LexLuthr: This can be done later when we make changes to ddo_v1

docker/piece-server/sample/mk20-ddo.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+set -e
+
+ci="\e[3m"
```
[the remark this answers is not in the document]
LexLuthr: Leaving it for now. Will be cleaned up in followup PR.

docker/piece-server/sample/pdp.sh
```diff
@@ -0,0 +1,54 @@
...
+read COMMP_CID PIECE CAR < <(sptool --actor t01000 toolbox mk20-client commp $FILE 2>/dev/null | awk -F': ' '/CID/ {cid=$2} /Piece/ {piece=$2} /Car/ {car=$2} END {print cid, piece, car}')
+
+mv $FILE /var/lib/curio-client/data/$COMMP_CID
+
+miner_actor=$(lotus state list-miners | grep -v t01000)
+printf "$COMMP_CID\n"
+printf "$PIECE\n"
+printf "$CAR\n"
+
+#if [ "$put" == "true" ]; then
```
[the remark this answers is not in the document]
LexLuthr: Leaving it for now. Will be cleaned up in followup PR. All scripts are tested

tasks/indexing/task_indexing.go
```diff
@@ -191,63 +449,177 @@ loop:
...
+	if _, err := reader.Seek(int64(dsis), io.SeekStart); err != nil {
+		return 0, nil, false, xerrors.Errorf("seeking to data segment index start offset: %w", err)
+	}
+
+	idata, err := parseDataSegmentIndex(reader)
 	if err != nil {
-		return false, xerrors.Errorf("adding index to DB (interrupted %t): %w", interrupted, err)
+		return 0, nil, false, xerrors.Errorf("parsing data segment index: %w", err)
+	}
+	if len(idata.Entries) == 0 {
```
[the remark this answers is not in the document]
LexLuthr: No really. It will require changes to the data-segment library

tasks/pdp/task_save_cache.go
LexLuthr: Proving cache for large piece in PDP otherwise, we build whole Tree in memory which required 2x piece size memory

Reiers (approved):
Approved ✅ I didn’t review the code, I just trusted the gravitational pull of the diff.

PR #516, opened 2025-06-04
Fix/public service sql mig

ZenGround0:
This way curio will be able to actually use the default public service.


ZenGround0:
Upgrade gets applied automatically on restart as confirmed by database query
```
select * from curio.pdp_services
+----+----------------+---------------+------------------------------+
| id | pubkey         | service_label | created_at                   |
|----+----------------+---------------+------------------------------|
| 1  | \x1c0de4c0ffee | public        | 2025-06-04 05:37:29.19118+00 |
+----+----------------+---------------+------------------------------+
```

Issue #520, opened 2025-06-07
pdp API needs new endpoint: GET /pdp/proof-sets/{set-id}/roots/added/{txHash}


rvagg:
Currently the transaction hash gets hidden internally. I propose that we make `handleAddRootToProofSet` similar to similar to `handleCreateProofSet` and have it return (approximately) this:

```go
	w.Header().Set("Location", "/pdp/proof-sets/{set-id}/roots/added/{txHash}")
```

And then a new API, like `/pdp/proof-sets/created`, that can look up the status of the AddProofSet transaction and print out the status.

Then a client can know when, and whether, Curio has accepted the piece because it sees the chain has.

It would also be good to have `handleGetProofSetRoot` implemented so we can look up details once it's properly created.

rvagg:
@ZenGround0 similar (but not quite the same) to the create-proofset situation - Curio's timeline is different to the chain's. We get the tx from Curio for create-proofset, watch it land on chain, are happy with it from the client end, but it takes another 30-60s for Curio to report that it's happy with it. If we proceed after we see it land on chain then we get a failure.

That's handled here, but Curio is always well behind the chain when you watch it in action: https://github.com/FilOzone/synapse-sdk/blob/2482e2cdcbccb8941d3d1cc8365f2081ea6d3551/src/pandora/service.ts#L443-L459

It's not quite the same for add-roots because with create-proofsets we rely on Curio agreeing that it's created before we can add roots, but with add-roots it _should_ be a stand-alone operation. However, it would be nice to know (a) that Curio agrees that it's added and it's going to start proving it, just for completeness and knowing that Curio didn't error or find disagreement somewhere, and (b) I'm a bit concerned about state management since we need to bake in nextRootId, how rapidly can we follow up with another addRoot, can we do it when we see the tx land but before Curio is happy that it's landed?

Mostly I want it so the client can be happy that it's both landed on chain and that Curio is actively proving it.

rjan90:
This can be closed now that https://github.com/filecoin-project/curio/pull/530 has been merged, but I do not have permissions to click the close button.

rvagg:
https://github.com/filecoin-project/curio/pull/535 is still open, it's against main

@LexLuthr would you mind having a look at that PR? the equivalent on the synapse branch got reviewed a few times, tweaked and merged. That PR should match but it's got minor differences (mainly to account for the AuthService changes on synapse branch)

rjan90:
Closing now that both #530 and #535 has been merged

PR #530, opened 2025-06-12
feat(pdp): add transaction tracking for root additions (synapse branch)

rvagg:
- Add `Location` header to root addition response with tx hash
- Add `GET /proof-sets/{id}/roots/added/{txHash}` endpoint for status checks
- Implement `handleGetProofSetRoot` for retrieving root details
- Return confirmed root IDs after transaction confirmation
- **Use lowercase transaction hashes consistently in database operations**

### DIFF
--- a/pdp/handlers.go
+++ b/pdp/handlers.go
@@ -1036,18 +1036,18 @@ func (p *PDPService) handleGetRootAdditionStatus(w http.ResponseWriter, r *http.
 		return
 	}
 
+	// Determine unique roots list
+	uniqueRootMap := make(map[string]bool)
+	for _, ra := range rootAdds {
+		uniqueRootMap[ra.Root] = true
+	}
+
 	// Step 6: If transaction is confirmed and successful, get assigned root IDs
 	var confirmedRootIds []uint64
 	if txStatus == "confirmed" && len(rootAdds) > 0 && rootAdds[0].AddMessageOK != nil && *rootAdds[0].AddMessageOK {
-		// Get unique roots from the add transaction
-		uniqueRoots := make(map[string]bool)
-		for _, ra := range rootAdds {
-			uniqueRoots[ra.Root] = true
-		}
-
 		// Query pdp_proofset_roots for confirmed roots with their IDs
-		rootCids := make([]string, 0, len(uniqueRoots))
-		for root := range uniqueRoots {
+		rootCids := make([]string, 0, len(uniqueRootMap))
+		for root := range uniqueRootMap {
 			rootCids = append(rootCids, root)
 		}
 
@@ -1075,12 +1075,6 @@ func (p *PDPService) handleGetRootAdditionStatus(w http.ResponseWriter, r *http.
 	}
 
 	// Step 7: Build and send response
-	// Count unique roots for response
-	uniqueRootMap := make(map[string]bool)
-	for _, ra := range rootAdds {
-		uniqueRootMap[ra.Root] = true
-	}
-
 	// Check that all roots have the same RootsAdded value (consistency check)
 	if len(rootAdds) > 0 {
 		firstRootsAdded := rootAdds[0].RootsAdded
@@ -1329,7 +1323,6 @@ func (p *PDPService) handleGetProofSetRoot(w http.ResponseWriter, r *http.Reques
 		Subroots: make([]SubrootResponse, 0, len(subroots)),
 	}
 
-	// Convert database results to response format
 	for _, subroot := range subroots {
 		response.Subroots = append(response.Subroots, SubrootResponse{
 			SubrootCid:    subroot.SubrootCID,
--- a/pdp/handlers.go
+++ b/pdp/handlers.go
@@ -1001,12 +1001,13 @@ func (p *PDPService) handleGetRootAdditionStatus(w http.ResponseWriter, r *http.
 		SubrootOffset   int64  `db:"subroot_offset"`
 		SubrootSize     int64  `db:"subroot_size"`
 		AddMessageOK    *bool  `db:"add_message_ok"`
+		RootsAdded      bool   `db:"roots_added"`
 	}
 
 	var rootAdds []RootAddInfo
 	err = p.db.Select(ctx, &rootAdds, `
 		SELECT root, add_message_index, subroot, subroot_offset,
-		       subroot_size, add_message_ok
+		       subroot_size, add_message_ok, roots_added
 		FROM pdp_proofset_root_adds
 		WHERE proofset = $1 AND add_message_hash = $2
 		ORDER BY add_message_index, subroot_offset
@@ -1080,12 +1081,22 @@ func (p *PDPService) handleGetRootAdditionStatus(w http.ResponseWriter, r *http.
 		uniqueRootMap[ra.Root] = true
 	}
 
+	// Check if all roots have been processed
+	allRootsProcessed := true
+	for _, ra := range rootAdds {
+		if !ra.RootsAdded {
+			allRootsProcessed = false
+			break
+		}
+	}
+
 	response := struct {
 		TxHash           string   `json:"txHash"`
 		TxStatus         string   `json:"txStatus"`
 		ProofSetId       uint64   `json:"proofSetId"`
 		RootCount        int      `json:"rootCount"`
 		AddMessageOK     *bool    `json:"addMessageOk"`
+		RootsAdded       bool     `json:"rootsAdded"`
 		ConfirmedRootIds []uint64 `json:"confirmedRootIds,omitempty"`
 	}{
 		TxHash:           txHash,
@@ -1093,6 +1104,7 @@ func (p *PDPService) handleGetRootAdditionStatus(w http.ResponseWriter, r *http.
 		ProofSetId:       proofSetID,
 		RootCount:        len(uniqueRootMap),
 		AddMessageOK:     rootAdds[0].AddMessageOK,
+		RootsAdded:       allRootsProcessed,
 		ConfirmedRootIds: confirmedRootIds,
 	}
 
--- a/pdp/handlers.go
+++ b/pdp/handlers.go
@@ -1081,13 +1081,19 @@ func (p *PDPService) handleGetRootAdditionStatus(w http.ResponseWriter, r *http.
 		uniqueRootMap[ra.Root] = true
 	}
 
-	// Check if all roots have been processed
-	allRootsProcessed := true
-	for _, ra := range rootAdds {
-		if !ra.RootsAdded {
-			allRootsProcessed = false
-			break
+	// Check that all roots have the same RootsAdded value (consistency check)
+	if len(rootAdds) > 0 {
+		firstRootsAdded := rootAdds[0].RootsAdded
+		for _, ra := range rootAdds[1:] {
+			if ra.RootsAdded != firstRootsAdded {
+				http.Error(w, "Inconsistent rootsAdded state for this transaction's roots", http.StatusInternalServerError)
+				return
+			}
 		}
+	}
+	allRootsProcessed := false
+	if len(rootAdds) > 0 {
+		allRootsProcessed = rootAdds[0].RootsAdded
 	}
 
 	response := struct {
--- a/pdp/handlers.go
+++ b/pdp/handlers.go
@@ -225,7 +225,8 @@ func (p *PDPService) handleCreateProofSet(w http.ResponseWriter, r *http.Request
 	}
 
 	// Step 6: Insert into message_waits_eth and pdp_proofset_creates
-	err = p.insertMessageWaitsAndProofsetCreate(ctx, txHash.Hex(), serviceLabel)
+	txHashLower := strings.ToLower(txHash.Hex())
+	err = p.insertMessageWaitsAndProofsetCreate(ctx, txHashLower, serviceLabel)
 	if err != nil {
 		log.Errorf("Failed to insert into message_waits_eth and pdp_proofset_creates: %+v", err)
 		http.Error(w, "Internal server error", http.StatusInternalServerError)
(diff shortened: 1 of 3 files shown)


rvagg:
Forgot to mention this is related to this issue: https://github.com/filecoin-project/curio/issues/520 and it will be closed by merging.

pdp/handlers.go
```diff
@@ -909,15 +914,195 @@ func (p *PDPService) handleAddRootToProofSet(w http.ResponseWriter, r *http.Requ
...
+	}
+
+	// Step 4: Query pdp_proofset_root_adds for this transaction
+	type RootAddInfo struct {
+		Root            string `db:"root"`
+		AddMessageIndex int    `db:"add_message_index"`
+		Subroot         string `db:"subroot"`
+		SubrootOffset   int64  `db:"subroot_offset"`
+		SubrootSize     int64  `db:"subroot_size"`
+		AddMessageOK    *bool  `db:"add_message_ok"`
```
aarshkshah1992, line 1003: Will this be atomically yes or no for all subroots under a root that belong to one tx ?

aarshkshah1992: Because we do this below:

```
	if txStatus == "confirmed" && len(rootAdds) > 0 && rootAdds[0].AddMessageOK != nil && *rootAdds[0].AddMessageOK {

```

aarshkshah1992, line 1003: why do we use a pointer for reading `AddMessageOK ` but values for everything else ?

aarshkshah1992: What is the difference between `AddMessageOK =true` and `txStatus == "confirmed"` ?

pdp/handlers.go
```diff
@@ -1018,74 +1203,128 @@ func (p *PDPService) handleDeleteProofSetRoot(w http.ResponseWriter, r *http.Req
...
+	}
+
+	// Step 3: Verify ownership and get root details
+	var rootCID string
+	err = p.db.QueryRow(ctx, `
+		SELECT DISTINCT r.root
+		FROM pdp_proofset_roots r
+		JOIN pdp_proof_sets ps ON ps.id = r.proofset
+		WHERE r.proofset = $1 AND r.root_id = $2 AND ps.service = $3
+	`, proofSetID, rootID, serviceLabel).Scan(&rootCID)
```
aarshkshah1992, line 1270: Auth here works because we make the `serviceLabel ` part of the query.

pdp/handlers.go
```diff
@@ -169,7 +171,7 @@ func (p *PDPService) handleCreateProofSet(w http.ResponseWriter, r *http.Request
 	}
 
 	// Decode extraData if provided
-	var extraDataBytes = []byte{}
+	extraDataBytes := []byte{}
```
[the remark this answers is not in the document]
rvagg: no, and tbh my editor auto did this and I didn't notice until commit; I didn't care enough to back them out!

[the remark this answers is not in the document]
rvagg: Ugh, I can't, `go fmt` is insisting in my editor.

pdp/handlers.go
```diff
@@ -909,15 +914,195 @@ func (p *PDPService) handleAddRootToProofSet(w http.ResponseWriter, r *http.Requ
...
+	if proofSetService != serviceLabel {
+		// Same response as not found to avoid leaking information
+		http.Error(w, "Proof set not found", http.StatusNotFound)
+		return
+	}
+
+	// Step 4: Query pdp_proofset_root_adds for this transaction
+	type RootAddInfo struct {
+		Root            string `db:"root"`
+		AddMessageIndex int    `db:"add_message_index"`
```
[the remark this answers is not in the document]
rvagg: I'm open to not including this on the response btw, I think it's mainly useful for sorting and I _think_ there's an implicit assumption that this will be 0...n in the results anyway that should hole true.

rvagg:
I added a new commit here after I ran into problems actually using this - the immediate deletion of the record when the tx lands means that querying for it fails. But we don't do this for createproofset so I've now aligned the impl to that:


```
fix(pdp): retain root addition records to prevent 404s on status queries

Previously, pdp_proofset_root_adds records were deleted after processing,
causing clients to receive 404 errors when querying transaction status
after successful root additions. This created a poor UX as clients couldn't
verify transaction outcomes if they queried too late.

This change aligns root addition tracking with proof set creation by:
- Adding roots_added column to mark processed records instead of deleting
- Updating watcher to UPDATE instead of DELETE when processing completes
- Including roots_added status in API response for client visibility

The fix ensures transaction status remains queryable indefinitely while
maintaining backward compatibility. Existing records will be processed
normally with roots_added=FALSE by default.

API enhancement: /pdp/proof-sets/{id}/roots/added/{txHash} now includes
"rootsAdded" field indicating if roots were fully processed.
```

tasks/pdp/proofset_addroot_watch.go
```diff
@@ -59,7 +59,7 @@ func processPendingProofSetRootAdds(ctx context.Context, db *harmonydb.DB, ethCl
 	err := db.Select(ctx, &rootAdds, `
         SELECT DISTINCT proofset, add_message_hash
         FROM pdp_proofset_root_adds
-        WHERE add_message_ok = TRUE
+        WHERE add_message_ok = TRUE AND roots_added = FALSE
```
aarshkshah1992, line 62: @rvagg Same as below, why do you need to filter on `roots_added = FALSE` ? This update is idempotent.

rvagg: Cause I don't want to do extra work. Strictly yes, it'd be idempotent, but we know the limiting condition here so we can `if len(rootAdds) == 0 {` and bail early if there's nothing and we don't spin our wheels.

rvagg: Remember the code before this would return `len(rootAdds)` because they were deleted previously

tasks/pdp/proofset_addroot_watch.go
```diff
@@ -210,18 +210,18 @@ func extractAndInsertRootsFromReceipt(ctx context.Context, db *harmonydb.DB, rec
...
 		}
 
-		// Delete from pdp_proofset_root_adds
+		// Mark as processed in pdp_proofset_root_adds (don't delete, for transaction tracking)
 		_, err = tx.Exec(`
-            DELETE FROM pdp_proofset_root_adds
-            WHERE proofset = $1 AND add_message_hash = $2
+            UPDATE pdp_proofset_root_adds
+            SET roots_added = TRUE
+            WHERE proofset = $1 AND add_message_hash = $2 AND roots_added = FALSE
```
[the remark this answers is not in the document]
aarshkshah1992: "good to stick a return check on this to make sure something did get updated like we expect it to"

This makes a lot of sense, if we are absolutely sure that there's no code-path that leads here twice for roots added in the same tx.

aarshkshah1992 (approved):
lgtm

pdp/handlers.go
```diff
@@ -909,15 +914,213 @@ func (p *PDPService) handleAddRootToProofSet(w http.ResponseWriter, r *http.Requ
...
 	// Step 10: Respond with 201 Created
+	w.Header().Set("Location", path.Join("/pdp/proof-sets", proofSetIDStr, "roots/added", txHashLower))
 	w.WriteHeader(http.StatusCreated)
 }
 
+// handleGetRootAdditionStatus handles GET /pdp/proof-sets/{proofSetID}/roots/added/{txHash}
+func (p *PDPService) handleGetRootAdditionStatus(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+
+	// Step 1: Verify that the request is authorized using ECDSA JWT
```
ZenGround0, line 931: Comment misleading since in general its not JWT and its hardcoded to not be JWT right now

pdp/handlers.go
```diff
@@ -909,15 +914,195 @@ func (p *PDPService) handleAddRootToProofSet(w http.ResponseWriter, r *http.Requ
...
+			http.Error(w, "Transaction status not found", http.StatusNotFound)
+			return
+		}
+		http.Error(w, "Failed to query transaction status: "+err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	// Step 6: If transaction is confirmed and successful, get assigned root IDs
+	var confirmedRootIds []uint64
+	if txStatus == "confirmed" && len(rootAdds) > 0 && rootAdds[0].AddMessageOK != nil && *rootAdds[0].AddMessageOK {
```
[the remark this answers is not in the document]
ZenGround0: nit: checking len > 0 is a little too defensive for me since you've already failed on 1020 if this case isn't met.

pdp/handlers.go
```diff
@@ -909,15 +914,213 @@ func (p *PDPService) handleAddRootToProofSet(w http.ResponseWriter, r *http.Requ
...
+			// Extract just the root IDs
+			for _, cr := range confirmedRoots {
+				confirmedRootIds = append(confirmedRootIds, cr.RootID)
+			}
+		}
+	}
+
+	// Step 7: Build and send response
+	// Count unique roots for response
+	uniqueRootMap := make(map[string]bool)
```
ZenGround0, line 1079: Am I missing something or is this just AI being AI?  Declare this before the if statement so you don't have to do this twice.

pdp/handlers.go
```diff
@@ -909,15 +914,213 @@ func (p *PDPService) handleAddRootToProofSet(w http.ResponseWriter, r *http.Requ
...
+	for _, ra := range rootAdds {
+		uniqueRootMap[ra.Root] = true
+	}
+
+	// Check that all roots have the same RootsAdded value (consistency check)
+	if len(rootAdds) > 0 {
+		firstRootsAdded := rootAdds[0].RootsAdded
+		for _, ra := range rootAdds[1:] {
+			if ra.RootsAdded != firstRootsAdded {
+				http.Error(w, "Inconsistent rootsAdded state for this transaction's roots", http.StatusInternalServerError)
```
ZenGround0, line 1089: I'm not so sure what value we get out of this.  It feels like we're testing the consistency model of yugabyte and I don't really care as a user for the purposes of this call.  I actually just care about the blockchain consistency model.

ZenGround0: Similar feeling about RootsAdded return value.  As user I care about chain and upload commitment and this seems to be some extraneous other bookkeeping.

pdp/handlers.go
```diff
@@ -909,15 +914,213 @@ func (p *PDPService) handleAddRootToProofSet(w http.ResponseWriter, r *http.Requ
...
+		var confirmedRoots []ConfirmedRoot
+		err = p.db.Select(ctx, &confirmedRoots, `
+			SELECT DISTINCT root_id, root
+			FROM pdp_proofset_roots
+			WHERE proofset = $1 AND root = ANY($2)
+			ORDER BY root_id
+		`, proofSetID, rootCids)
+		if err != nil {
+			log.Warnf("Failed to query confirmed roots: %v", err)
+			// Don't fail the request, just log the warning
```
ZenGround0, line 1068: Whats the rationale on returning anyway with no root IDs?  It would be simpler to just fail right?

pdp/handlers.go
```diff
@@ -1018,74 +1221,128 @@ func (p *PDPService) handleDeleteProofSetRoot(w http.ResponseWriter, r *http.Req
...
+		RootCid  string            `json:"rootCid"`
+		Subroots []SubrootResponse `json:"subroots"`
+	}{
+		RootId:   rootID,
+		RootCid:  rootCID,
+		Subroots: make([]SubrootResponse, 0, len(subroots)),
+	}
 
-	// Respond with root details
+	// Convert database results to response format
```
ZenGround0, line 1332: nit: comment too verbose better to remove

ZenGround0 (approved):
Overall I like it

pdp/handlers.go
rvagg: Yeah, but I'm just following the pattern in the file that exists now:

```
$ grep 'Verify that the request' pdp/handlers.go 
        // Verify that the request is authorized using ECDSA JWT
        // Step 1: Verify that the request is authorized using ECDSA JWT
        // Step 1: Verify that the request is authorized using ECDSA JWT
        // Step 1: Verify that the request is authorized using ECDSA JWT
        // Step 1: Verify that the request is authorized using ECDSA JWT
        // Step 1: Verify that the request is authorized using ECDSA JWT
        // Step 1: Verify that the request is authorized using ECDSA JWT
        // Step 1: Verify that the request is authorized using ECDSA JWT
```

Do you want me to remove all of those, or just these new ones?

pdp/handlers.go
rvagg: OK, your call, Aarsh and I went back and forth on this and I don't think I care. I'll remove the checks and `rootsAdded`.

rvagg: Oh, actually no I remember now - we _added_ this new field to the database, `roots_added`, it's the final check that tells us that Curio has fully processed it and it's onboarded. It's the direct equivalent of `proofset_created`. SO I need `roots_added`, it's just a question of whether we're OK trusting that the first row we get from the DB is representative of the whole lot. Perhaps there's a Curio-bug-reason that makes the first one `false` and the rest `true`; that's the logic that lead to all the looping and checking. So, your call on this but I'd really like the `roots_added` from for the SDK

pdp/handlers.go
rvagg: Because we could be in an in-between state. We get into this branch because `add_message_ok` is set on the `pdp_proofset_root_adds` table, but it's set because of a change in transaction status on the `message_waits_eth` table by the `pdp_proofset_add_message_status_change` stored procedure. Then, separate to that the `WatcherRootAdd` task polls `pdp_proofset_root_adds` to look for these `add_message_ok` and `roots_added` status changes and eventually updates `pdp_proofset_roots` which is what we are reading here.

So there is a moment where `rootAdds[0].AddMessageOK` is true but this `SELECT` statement won't return anything because our background task hasn't followed up on it. If we fail at that point then we're telling the user that there's a problem completing the root add, but in reality we're just closing it out and the next time they query it should return everything is 👍 .

pdp/handlers.go
ZenGround0: Ok this makes sense you might not eter the if above but still need this.  But can you do this above the if statement?  Its the exact map constructed from the exact same data in both cases right?  I dont understand why we would do this twice

ZenGround0: Im talking about 1043 to 1046not confirmedRoots

rvagg: yeah, sorry, I get you now, I'm just being blinkered on the steps here; if we just do this calc outside of the first `if` then we're all good.

PR #533, opened 2025-06-13
fix(pdp): look in pdp piecerefs to check for actual readiness

rvagg:
Also from the `synapse` branch. A bugfix that prevented "findPiece" from finding pieces, it was not looking in the right table.


LexLuthr (approved):

rvagg:
Good question and the original bug we were fixing is lost in my memory, I just know Zen fixed it with this. I'll try and dig up the details.

PR #545, opened 2025-06-20
PDP service contract whitelist

ZenGround0:
This provides some protection for curio PDP operators running publically accessible PDP services.  Only the pandora PDP service contract is allowed upon proofset creation over the publically available PDP service.  

The whitelist is network dependent -- only calibnet specified as pandora only launched on calibnet.

Closes https://github.com/FilOzone/synapse-sdk/issues/34

### DIFF
--- a/pdp/contract/addresses.go
+++ b/pdp/contract/addresses.go
@@ -12,7 +12,8 @@ import (
 )
 
 type PDPContracts struct {
-	PDPVerifier common.Address
+	PDPVerifier                common.Address
+	AllowedPublicRecordKeepers []common.Address
 }
 
 func ContractAddresses() PDPContracts {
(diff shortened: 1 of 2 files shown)


pdp/contract/whitelist.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+	
+	// If allowedList is empty (but not nil), no recordkeepers are allowed
+	if len(allowedList) == 0 {
+		return false
+	}
+	
+	// Check if the recordkeeper is in the whitelist
+	for _, allowed := range allowedList {
+		if recordKeeper == allowed {
```
rvagg, line 48: Do you know if `common.Address` is going to be happy with case-insensitive forms of the address? Should we just lower-case all addresses to avoid any issues with case sensitivity and not demand the user provide the correct validation form?

ZenGround0:
> fair enough if this is only for the synapse branch, but maybe it should be a config option rather than defined in code?

I see where you're coming from but since it only applies to the public service I am ok with leaving it as a build constant like the existing verifier.  Merging synapse back to main will require some conceptual merging of the public service construct and we can revisit this question then as part of that.

pdp/contract/whitelist.go
ZenGround0: Good question.  After looking into it the answer is that case insensitivity is handled correctly.

Tracing through common.Address everything relies in a straightforward way on goland stdlib hex encode decode.  Stdlib reverse lookup from ascii code to hex value is here: https://cs.opensource.google/go/go/+/refs/tags/go1.24.4:src/encoding/hex/hex.go;l=25 and you can see capital letters are treated correctly.

rjan90 (approved):

PR #547, opened 2025-06-23
fix(pdp): s/piece_cid/pieceCid to make JSON response casing consistent

rvagg:
This is the only place in the pdp handlers where we have _'s. This is probably good to fix earlier than later. On the client side, in my code at least, I'll make it accept both forms for now and remove the older form when we think we have it rolled out properly.

### DIFF
--- a/pdp/handlers_upload.go
+++ b/pdp/handlers_upload.go
@@ -497,7 +497,7 @@ func (p *PDPService) handleFindPiece(w http.ResponseWriter, r *http.Request) {
 	}
 
 	response := struct {
-		PieceCID string `json:"piece_cid"`
+		PieceCID string `json:"pieceCid"`
 	}{
 		PieceCID: pieceCid.String(),
 	}


rjan90:
@LexLuthr does the change here look okay to you, and can be merged?

LexLuthr:
This change requires coordination with Storacha as we are updating JSON response. In case, they are using this API, their system also needs to be updated. Don't we need to client side changes or was that done in some other PR?

pdp/handlers_upload.go
```diff
@@ -497,7 +497,7 @@ func (p *PDPService) handleFindPiece(w http.ResponseWriter, r *http.Request) {
 	}
 
 	response := struct {
-		PieceCID string `json:"piece_cid"`
+		PieceCID string `json:"pieceCid"`
 	}{
 		PieceCID: pieceCid.String(),
 	}
```
frrist, line 503: Rather than define this as an anonymous struct, could it be defined as a named struct? Ideally along with all the other request response type curio returns? 

This could allow Storacha, and other services depending on Curio, to import the types directly, rather than manual coordination.

frrist:
What version of Curio is this change expected to land in?

LexLuthr:
> What version of Curio is this change expected to land in?

Next release v1.25.2

pdp/handlers_upload.go
LexLuthr: I do agree with your ask to have proper names structs all over the place but Curio team is prioritising the market v2 work instead of enhancements to existing implementation. You can open a PR which can be reviewed by both Curio and FilOz if you are interested in contributing this change. Otherwise, Market PR is likely to be merged in next 2 weeks.

Issue #548, opened 2025-06-23
PDP: addRoots intermittently fails - chain watcher stops processing transactions


rjan90:
## Description:
Curio daemon intermittently stops processing root additions to proof sets, causing StorageService addRoots operations to fail with timeout errors. The transactions are confirmed on-chain but the server fails to acknowledge them.

```
StorageService addRoots failed: Failed to add root to proof set
Caused by: Failed to verify root addition after 420 seconds: Server did not provide confirmation. The transaction was confirmed on-chain but the server failed to acknowledge it.
```

We have logs (stderr and stdout), but can’t find any explanation for the chain watcher stopping watching and processing these root adds.

```
SELECT
       add_message_ok,
       roots_added,
       COUNT(*)
   FROM curio.pdp_proofset_root_adds
   GROUP BY add_message_ok, roots_added;
+----------------+-------------+-------+
| add_message_ok | roots_added | count |
|----------------+-------------+-------|
| <null>         | False       | 14    |
| True           | True        | 12    |
+----------------+-------------+-------+
```

## Temporary Workaround:
- Restarting the Curio daemon process resolves the issue temporarily 
- Issue is likely to recur

## Investigation Notes:
- Memory issues suspected but adding 32GB swap didn't resolve it
- Need to catch the exact moment when chain watcher stops to narrow down timing for log analysis

rvagg:
Debugging datapoint: for the proofset create I can see this:

```
> select * from curio.message_waits_eth where signed_tx_hash = '0x3005f4a91b26648cf54d94933d147798885aa38cce2dc6cdebb66f53693cc75d';
+--------------------------------------------------------------------+-------------------+------------------------+-------------------+-------------------+-----------+------------+------->
| signed_tx_hash                                                     | waiter_machine_id | confirmed_block_number | confirmed_tx_hash | confirmed_tx_data | tx_status | tx_receipt | tx_suc>
|--------------------------------------------------------------------+-------------------+------------------------+-------------------+-------------------+-----------+------------+------->
| 0x3005f4a91b26648cf54d94933d147798885aa38cce2dc6cdebb66f53693cc75d | 1                 | <null>                 | <null>            | <null>            | pending   | <null>     | <null>>
+--------------------------------------------------------------------+-------------------+------------------------+-------------------+-------------------+-----------+------------+------->
SELECT 1
```

with no change in status, in fact there's a lot of pending transactions in there which is odd. 

Running this to reset it to get it reassigned and re-watched:

```
[...]

PR #552, opened 2025-06-27
fix(chain): improve chain scheduler robustness and eth tx monitoring

rvagg:
Still working on https://github.com/filecoin-project/curio/issues/548, it's still not fixed. And my latest diagnosis from looking at the logs is that the `MessageWatcherEth` just stops being called at some point so it doesn't even have a chance to run. Which leaves us looking at `CurioChainSched` as a potential culprit. But I don't have any useful logs that tell me what might be wrong.

So this PR:

* Adds a mutex around `callbacks`, which is not a likely culprit since I think all listeners are added at construction time, synchronously, but this seems like a good thing to do because there's no docs or guards that would stop new callbacks being added async
* Lots more logging so we get insight into its operation
* A secondary watcher ticker that will fire every 60 seconds to make sure we're still getting notifications, and if we haven't had a notification in 60 seconds then assume something is wrong and set up a new channel

### DIFF
--- a/tasks/message/watch_eth.go
+++ b/tasks/message/watch_eth.go
@@ -25,7 +25,7 @@ import (
 const (
 	// defaultEthCallTimeout is the timeout for sets of Ethereum client calls per transaction
 	// (i.e. receipt and transaction data)
-	defaultEthCallTimeout = time.Second
+	defaultEthCallTimeout = 30 * time.Second
 )
 
 // EthClient is an interface for the Ethereum client operations we need
@@ -129,6 +129,9 @@ func (mw *MessageWatcherEth) update() {
 
 	// Check if any of the transactions we have assigned are now confirmed
 	for _, txHashStr := range txHashes {
+		processed++
+		processStart := time.Now()
+
 		txHash := common.HexToHash(txHashStr)
 		log.Debugw("Checking transaction", "txHash", txHash.Hex())
 
@@ -244,8 +247,14 @@ func (mw *MessageWatcherEth) update() {
 		}
 
 		confirmed++
-		processed++
 		log.Infow("Successfully confirmed transaction", "txHash", txHash.Hex())
+
+		processDuration := time.Since(processStart)
+		if processDuration > 5*time.Second {
+			log.Warnw("Transaction processing took longer than expected",
+				"txHash", txHash.Hex(),
+				"duration", processDuration)
+		}
 	}
 
 	log.Infow("MessageWatcherEth update completed",


lib/chainsched/chain_sched.go
```diff
@@ -41,7 +48,8 @@ func (s *CurioChainSched) AddHandler(ch UpdateFunc) error {
 	if s.started {
 		return xerrors.Errorf("cannot add handler after start")
 	}
-
+	s.callbacksLk.Lock()
+	defer s.callbacksLk.Unlock()
```
magik6k, line 52: I'd put the log above started. Even though I'm 99% sure that all hook adds happen in construction, if we want to use the lock as a guard, for it to actually do anything it has to be above the `if started`

magik6k (commented):
Generally good to be defensive here so makes sense to me; One comment to address then it should be good.

Would be good to get in the habit of writing tests for the more critical components like this, but not going to block on that..

Fwiw in the years of having curio in production I've never had seen a single issue originating from the chain watcher so your problem is a little suspect. I'd suggest looking at `http://node-addr:12300/debug/pprof/goroutine?debug=1`, I'd guess that the watcher is running, just stuck somewhere (e.g. long-running postgres operation because e.g. some query is missing some index, or some index is a hash index (yb default and a big footgun) and not sorted - which has historically been the case quite often)

rvagg:
> that the watcher is running, just stuck somewhere

so after much tinkering and eventually deploying https://github.com/filecoin-project/curio/pull/553 this has fixed the problem. The call to ethclient's `TransactionByHash` just locks up and doesn't return. On the lotus side we have an equally weird error and my hunch is that lotus is returning an error in a form that ethclient isn't built to recognise so we don't get a return. On the lotus side I suspect this might be to do with a low-resource machine hitting splitstore compaction and then having some trouble properly processing the request resulting in an error. So, probably things to fix on the lotus side, but also here we need to be more resillient.

I'll combine the two into one, PR, remove some of the cruft and get it in better shape for merging.

rvagg:
@magik6k I incorporated the changes in #553, added a bunch of tests. The line that actually fixes the problem is https://github.com/filecoin-project/curio/blob/c52f983e5cc816074c6db75f07d44491285f9a6b/tasks/message/watch_eth.go#L130.

Just timing out the ethclient calls is enough to keep it all running. But I've left the other pieces in place because IMO they add a bit more robustness. I've had to do some restructuring to make some of these testable, unfortunately it increases the noise of the diff but they should be an improvement overall in being able to add tests in the future.

#553 is identical to this now but it's for the `synapse` branch, while this is for `main`.

magik6k:
Looks good now, needs conflicts resolved

rvagg:
welp, most of this came in with the synapse branch merge, but there's some review items in here now to land

rjan90:
@magik6k can you sanity check the latest commit here after you approved? And maybe push the ⛙ button if all is okay?

PR #559, opened 2025-07-15
fix(pdp): GET /pdp/proof-sets/{id} - handle uninitialised proof sets

rvagg:
When there are no roots in a proof set we don't yet have a `prove_at_epoch` so we need to gracefully handle that case.

Current code does:

500 Internal Server Error - Failed to retrieve next challenge epoch: can't scan into dest[0]: cannot scan NULL into *int64

### DIFF
--- a/pdp/handlers.go
+++ b/pdp/handlers.go
@@ -509,8 +509,8 @@ func (p *PDPService) handleGetProofSet(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// Step 6: Get the next challenge epoch
-	var nextChallengeEpoch int64
+	// Step 6: Get the next challenge epoch (can be NULL for uninitialized proof sets)
+	var nextChallengeEpoch *int64
 	err = p.db.QueryRow(ctx, `
         SELECT prove_at_epoch
         FROM pdp_proof_sets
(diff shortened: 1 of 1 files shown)


LexLuthr (commented):
This will need @frrist to review once as we are changing the response here.

pdp/handlers.go
```diff
@@ -488,14 +488,22 @@ func (p *PDPService) handleGetProofSet(w http.ResponseWriter, r *http.Request) {
...
+
 	response := struct {
 		ID                 uint64      `json:"id"`
 		Roots              []RootEntry `json:"roots"`
 		NextChallengeEpoch int64       `json:"nextChallengeEpoch"`
 	}{
 		ID:                 proofSet.ID,
-		NextChallengeEpoch: nextChallengeEpoch,
+		NextChallengeEpoch: epochValue,
+		Roots:              []RootEntry{}, // Initialize as empty array, not nil
```
frrist, line 506: @LexLuthr I don't think the response payload is changing here, right?
Note: it would be _really nice_ if these were typed structs instead of anonymous.

frrist:
LGTM

LexLuthr:
@BigLep Failing action needs to be fixed to allow merging.

PR #560, opened 2025-07-16
Narrow mutex scope to only protect skipped append, improving concurrency

asamuj:


LexLuthr (changes requested):
Please add details to the PR. What problem you are solving, reproducer step and how you are fixing it. Otherwise, PR will be rejected.

PR #567, opened 2025-07-25
fix: per sector commit simulation

LexLuthr:
This PR adds a per sector simulation to Commit Submit job which works in batches. This should allow users to easily spot the faulty sector and process will also skip it for a higher success rate with major service disruption. Failure should leave the sector in pipeline without removing the submit task ID. This should make it easier to identify the problematic sectors.


magik6k (approved):
Makes sense to me, would be good to test - but I guess we'll find out very quickly if it's broken

PR #570, opened 2025-07-31
fix: misc fixes to get pdp enabled server to start

rvagg:


harmony/harmonytask/task_type_handler.go
```diff
@@ -346,10 +346,12 @@ func (h *taskTypeHandler) AssertMachineHasCapacity() error {
 		return xerrors.Errorf("Did not accept %s task: out of available GPU: required %f available %f)", h.Name, h.Cost.Gpu, r.Gpu)
 	}
 
+	/*
```
rvagg, line 349: I don't know what's going on here but I've had this cause problems on both Curio nodes I've tried to run; I thought it was because it insisted on having a sector size's space available but on the first server where I had this be a problem on it was set to 64G sectors and I freed up over 90G and it still complained about this. On the server I'm now running mkv2 on I have 402G free but every "piece" task fails with this so I give up.

market/mk20/http/http.go
```diff
@@ -130,7 +130,8 @@ func APIRouter(mdh *MK20DealHandler, domainName string) http.Handler {
 	SwaggerInfo.Version = version
 	mux := chi.NewRouter()
 	mux.Use(dealRateLimitMiddleware())
-	mux.Use(AuthMiddleware(mdh.db))
+	// Disabled, far too complex to deal with for now
+	// mux.Use(AuthMiddleware(mdh.db))
```
rvagg, line 134: Couldn't get this to work, too complex particularly when working with Ethereum addresses and this is built for Filecoin addresses using the specific Go encoding forms that aren't as readily accessible in other languages. This auth thing needs a rethink to be more accessible and usable.

LexLuthr:
I have created a client for testing and already fixed this and more. I will continue testing and fixing till all paths are tested. 
I am closing this PR as next commit on the branch will supersede this.

PR #574, opened 2025-08-03
feat: wallet exporter

magik6k:
Another extension of the named wallet feature, essentially extending it to export wallet activity to prometheus for analysis in Grafana.

Could potentially also make this data viewable in the webui, but the gymnastics needed for that aren't worth it imo.


magik6k:
Tested on my setup and appears to work great

LexLuthr (approved):

magik6k:
2025-08-08-181321_1601x284_scrot

Don't have much now, but generally it gives a pretty high-res view over everything. Tho with this kind of data it only becomes interesting  / useful after a few months when you can really zoom out to see patterns

PR #604, opened 2025-08-22
fix(pdp): generate CommPv2 in GetDataSet based on raw piece sizes

rvagg:


rvagg:
This lead to a detour which resulted in this: https://github.com/FilOzone/pdp/issues/189

But yes, subpieces are broken right now, and achieving that linked issue will involve fixing them properly.

PR #607, opened 2025-08-26
fix gc mark snapdeal sector query

magik6k:
We were basically comparing apples to oranges. Every sector was considered as snapped - this was still mostly safe as for correct files to be removed there was still the separate check that update files existed, but this would have caused us to mark too many files for GC in some scenarios.


magik6k:
(also this comparison was extremely expensive as it's basically O(n^2) in how yugabyte run it by default. New one seems much faster)

PR #608, opened 2025-08-26
fix proofshare log spam

magik6k:
Those logs don't really say anything useful and are extreeeemely spammy, GBs per day.


magik6k:
Coalescing another small fix into here to not spam too many PRs

LexLuthr (approved):

PR #624, opened 2025-09-09
Enhance node info page with detailed fields

strahe:
## Changes
- Add supported tasks, miners, startup time, restart requests to node info page
- Add message wait status for Filecoin/Ethereum transactions
- Add storage URL health monitoring
- Enhance task tables with retries and execution details

## Purpose
Improve node visibility for better cluster management and troubleshooting.

iShot_2025-09-09_15 23 54

### DIFF
--- a/web/api/webrpc/cluster.go
+++ b/web/api/webrpc/cluster.go
@@ -196,18 +196,25 @@ type MachineInfo struct {
 
 	// Message Waits
 	MessageWaits []struct {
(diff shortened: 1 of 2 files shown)


LexLuthr:
Can you explain a little bit more about the message waits? Why do we have those and I don't see a screenshot for them. We already have message waits in a another page /wallet. Please take a look at that code and ensure we are not doing same thing.

PR #637, opened 2025-09-12
fix: Handle OPTIONS requests for CORS preflight

strahe:
Fixes CORS preflight OPTIONS requests returning 404 for API endpoints.

The previous CORS middleware in #629 worked fine for registered routes, but I missed that OPTIONS requests were hitting 404 before the middleware could handle them. This happens because Gorilla Mux returns 404 when no route matches the HTTP method.

Added a catch-all OPTIONS handler so all API endpoints can receive CORS preflight requests properly.

My bad for not testing this thoroughly in the first PR!


LexLuthr:
Superseded by https://github.com/filecoin-project/curio/pull/658

PR #657, opened 2025-09-24
herd-db-upgrade

snadrus:
This PR helps the very trusting SP who upgrades all their servers at once without testing.

It avoids problems of multiple DB upgrade attempts conflicting which can potentially block all servers from starting-up. 

Avoided Approaches: Since Yugabyte doesn't support SAVEPOINT on DDL, transactions won't help.  Fake transactions would be troubled in the case of server-died or partial-application-before-breakage. 

Approach: Make SQL idempotent. Verify with tests that run each twice.

### DIFF
--- a/market/mk20/http/docs.go
+++ b/market/mk20/http/docs.go
@@ -853,7 +853,6 @@ const docTemplate = `{
         },
         "github_com_filecoin-project_go-state-types_builtin_v16_verifreg.AllocationId": {
             "type": "integer",
-            "format": "int64",
             "enum": [
                 0
             ],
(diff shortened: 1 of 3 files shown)


snadrus:
I put the entire change into a single exec statement, so that part will be
transactional. That way, any conflict will come from after that transaction
completes, in which case it's just a matter of pulling on. If they are done
or trying again.
I built it the other way first, where we wrapped it in a regular begin
transaction.And yougabyte was very unhappy with it.

Andrew Jackson
President of Curio Storage, Inc. <http://curiostorage.org>

snadrus:
Ah, you're correct, Yugabyte does transactions different. 
In-fact DDL doesn't support SAVEPOINT at all, so herds could cross-conflict each other.
I'm going to change this to:
Lock, Check, Exec-Upgrades, Update base, Unlock. 

That way the only risk is a server that dies during upgrade.

PR #658, opened 2025-09-24
correct secondary ui server spec

snadrus:
This allows limiting secondary servers to a specific list


LexLuthr (commented):
We should ship some default or document them. The cors in market http server are meant to allow serving static websites on PDP. We should let users decided what degree of risk they are okay with but we must either ship or document good default.

Also, do we want to share the CORS between UI and Market server since they are meant to serve different things?

snadrus:
Changes I will make today:
1. better docs: "*" allows everyone, it's best to specify the UI server's hostname. 
2. rm from market server.

LexLuthr:
Thank you @strahe. We just wanted to test the UI part only

PR #664, opened 2025-09-26
Dynamic Config

snadrus:
Benefit: 
Compatible configs populate the latest value (with respect to the layers) every 30 seconds on running servers without restarting. 

Pieces:
- config.NewDynamic() to create a dynamic type (which can live outside the config tree for computed types)
 - outfit the first config
 - Ensure docs-gen reports the right type and "Updated instantly." suffix.
 - tests
 - Change Notifier which allows computed objects to. adapt to changes.

Decision-Points:
 - 30 seconds poll?

### DIFF
--- a/deps/config/dynamic.go
+++ b/deps/config/dynamic.go
@@ -3,6 +3,7 @@ package config
 import (
 	"context"
 	"fmt"
+	"math/big"
 	"reflect"
 	"strings"
 	"sync"
@@ -16,6 +17,11 @@ import (
 )
 
 var logger = logging.Logger("config-dynamic")
+
+// bigIntComparer is used to compare big.Int values properly
+var bigIntComparer = cmp.Comparer(func(x, y big.Int) bool {
+	return x.Cmp(&y) == 0
+})
 
 type Dynamic[T any] struct {
 	value T
@@ -70,7 +76,7 @@ func (d *Dynamic[T]) MarshalTOML() ([]byte, error) {
 // Equal is used by cmp.Equal for custom comparison.
 // If used from deps, requires a lock.
 func (d *Dynamic[T]) Equal(other *Dynamic[T]) bool {
-	return cmp.Equal(d.value, other.value)
+	return cmp.Equal(d.value, other.value, bigIntComparer)
 }
 
 type cfgRoot[T any] struct {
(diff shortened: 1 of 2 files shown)


deps/config/dynamic.go
```diff
@@ -0,0 +1,120 @@
...
+			}
+		}
+	}
+	walker(reflect.ValueOf(orig), result)
+	r.treeCopy = result.Interface().(*CurioConfig)
+	return nil
+}
+
+func (r *cfgRoot) changeMonitor() {
+	lastTimestamp := time.Now().Add(-30 * time.Second) // plenty of time for start-up
```
magik6k, line 90: I'd prefer some global flag for this. Large storage servers with millions of files on hard drives may need a few minutes for proper startup

deps/config/dynamic.go
```diff
@@ -0,0 +1,120 @@
...
+		if err != nil {
+			logger.Errorf("error getting configs: %s", err)
+			continue
+		}
+
+		// 2. lock "dynamic" mutex
+		func() {
+			DynamicMx.Lock()
+			defer DynamicMx.Unlock()
+			ApplyLayers(context.Background(), r.treeCopy, configs)
```
magik6k, line 116: Linter somewhat rightfully complains about a missing err check

deps/config/dynamic.go
```diff
@@ -0,0 +1,120 @@
...
+		}()
+		DynamicMx.Lock()
```
magik6k, line 118: Stray lock?

magik6k (commented):
Reads really well, I like how this refreshes fixed values a lot more than having that bloat directly in each instance of Dynamic.

I'd really like to have some Unit tests for the cursed reflect logic, AI should be pretty great at writing that sort of thing.

Also linters are complaining.

magik6k (approved):
Can't wait to use this. Couple of probably just questions. Would be nice to ask some AI model to write more unit tests for the `dynamic.go`

deps/config/dynamic.go
```diff
@@ -0,0 +1,255 @@
...
+func (d *Dynamic[T]) Get() T {
+	dynamicLocker.RLock()
+	defer dynamicLocker.RUnlock()
+	return d.value
+}
+
+// UnmarshalText unmarshals the text into the dynamic value.
+// After initial setting, future updates require a lock on the DynamicMx mutex before calling toml.Decode.
+func (d *Dynamic[T]) UnmarshalText(text []byte) error {
+	return toml.Unmarshal(text, d.value)
```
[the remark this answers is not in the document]
snadrus: corrected & test added (yes)

PR #670, opened 2025-10-01
Backport index and ipni pdp tasks to PDPv0

ZenGround0:
Backporting the indexing and ipni tasks from mkv2 to the pdpv0 branch.

Code is at least not obviously broken but testing is a work in progress.

### DIFF
--- a/documentation/en/configuration/default-curio-configuration.md
+++ b/documentation/en/configuration/default-curio-configuration.md
@@ -722,7 +722,7 @@ description: The default curio configuration
       # heads.
       #
       # type: []string
-      #DirectAnnounceURLs = ["https://cid.contact/ingest/announce"]
+      #DirectAnnounceURLs = ["https://cid.contact/ingest/announce", "https://filecoinpin.contact/announce"]
 
     # Indexing configuration for deal indexing
     #


tasks/indexing/task_pdp_ipni.go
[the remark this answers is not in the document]
ZenGround0: +1 That's fine for now.  I just want to get ipfs adds working first.

pdp/handlers.go
```diff
@@ -1050,7 +1050,46 @@ func (p *PDPService) handleAddPieceToDataSet(w http.ResponseWriter, r *http.Requ
...
+	if err != nil {
+		// Hard to differenctiate between nsupported listener type OR internal error
+		// So we log on debug and abort indexing attempt
+		log.Infow("Failed to get data set metadata, skipping indexing ", "error", err, "dataSetId", dataSetId)
+	} else {
+		if exists {
+			log.Debugw("Data set metadata exists, marking all subpiecesas needing indexing", "dataSetId", dataSetId)
+			// Note: it's possible to update a duplicate piece that has already completed the indexing step
+			// but task_pdp_indexing handles pieces that have already been indexed smoothly
+			_, err := p.db.Exec(ctx, `UPDATE pdp_piecerefs
```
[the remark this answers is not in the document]
ZenGround0: Thanks!

market/ipni/ipni-provider/ipni-provider.go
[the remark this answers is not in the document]
rvagg: does it really matter if it's retrievable?
what's the GC like in cid.contact these days, will it figure out that the announcements are invalid after some period of time? does it periodically check endpoints for new chains and expire old ones if the SP isn't available anymore or the chain it's providing is entirely new?

market/ipni/ipni-provider/ipni-provider.go
```diff
@@ -73,6 +74,7 @@ type Provider struct {
 	indexStore    *indexstore.IndexStore
 	sc            *chunker.ServeChunker
 	keys          map[string]*peerInfo // map[peerID String]Private_Key
+	latest        map[string]cid.Cid   // map[peerID String]last published head, used to avoid duplicate announce
```
rvagg, line 77: the changes in here wrt `latest` look like bugfixes, or is it just the fact that 30 seconds now make this task more onerous so it's worth being spare with what we provide?

pdp/handlers.go
```diff
@@ -974,7 +974,30 @@ func (p *PDPService) handleAddPieceToDataSet(w http.ResponseWriter, r *http.Requ
...
+	listenerAddr, err := pdpVerifier.GetDataSetListener(nil, dataSetId)
+	if err != nil {
+		log.Errorw("Failed to get listener address for data set", "error", err, "dataSetId", dataSetId)
+		http.Error(w, "Internal server error", http.StatusInternalServerError)
+		return
+	}
+	mustIndex, _, err := contract.GetDataSetMetadataAtKey(listenerAddr, p.ethClient, dataSetId, "withIPFSIndexing")
+	if err != nil {
+		// Hard to differenctiate between unsupported listener type OR internal error
+		// So we log on debug and skip indexing attempt
```
rvagg, line 994: this is what I think you should move up into `GetDataSetMetadataAtKey` - we can differentiate so let's just do it there

tasks/indexing/task_pdp_indexing.go
```diff
@@ -0,0 +1,270 @@
...
+	return nil
+}
+
+func (P *PDPIndexingTask) Adder(taskFunc harmonytask.AddTaskFunc) {}
+
+var _ harmonytask.TaskInterface = &PDPIndexingTask{}
+var _ = harmonytask.Reg(&PDPIndexingTask{})
+
+func IndexCAR(r io.Reader, buffSize int, recs chan<- indexstore.Record, addFail <-chan struct{}) (int64, bool, error) {
+	blockReader, err := carv2.NewBlockReader(bufio.NewReaderSize(r, buffSize), carv2.ZeroLengthSectionAsEOF(true))
```
rvagg, line 235:
```suggestion
	// ZeroLengthSectionAsEOF is not strictly needed here as it exists for the PoRep case where
	// padding pieces with zero bytes to get them to be a larger size is reasonable. This isn't
	// expected to be the case with PDP, but we'll stay consistent.
	blockReader, err := carv2.NewBlockReader(bufio.NewReaderSize(r, buffSize), carv2.ZeroLengthSectionAsEOF(true))
```

ZeroLengthSectionAsEOF was introduced specifically for this case, it's a weird feature that probably shouldn't exist. I wouldn't mind removing it from PDP but I can also imagine someone coming up with a whacky use for it... I wouldn't mind documenting it though for the future spelunker / maintainer.

tasks/indexing/task_pdp_ipni.go
```diff
@@ -0,0 +1,427 @@
...
+	blockReader, err := carv2.NewBlockReader(bufio.NewReaderSize(reader, 4<<20), opts...)
+	if err != nil {
+		return false, fmt.Errorf("getting block reader over piece: %w", err)
+	}
+
+	chk := chunker.NewInitialChunker()
+
+	blockMetadata, err := blockReader.SkipNext()
+	for err == nil {
+		if err := chk.Accept(blockMetadata.Hash(), int64(blockMetadata.SourceOffset), blockMetadata.Size+40); err != nil {
```
rvagg, line 132:
```suggestion
		// CAR sections are [varint (length), CID, blockData]
		combinedSize := blockMetadata.Size+c.ByteLen()
		lenSize := varint.UvarintSize(combinedSize)
		sectionSize := combinedSize + lenSize
		if err := chk.Accept(blockMetadata.Hash(), int64(blockMetadata.SourceOffset), sectionSize); err != nil {
```

Looking at the implementation I think this matters: https://github.com/filecoin-project/curio/blob/547a713b148c05ae319ad5dbb7700afdbb0f286c/market/ipni/chunker/initial-chunker.go#L81, although maybe not critical because it's doing chunk size processing? Not hard to be accurate here though.

Hard-coding `40` assumes a certain length varint at the start, followed by a certain length CID, both of which are actually variable.

Also @LexLuthr this is the other spot I was mentioning in https://github.com/filecoin-project/curio/pull/681#issuecomment-3379509599 where you still have `blockMetadata.Offset` in tasks/indexing/task_ipni.go, it should be `SourceOffset` because the chunker seems to care about the first offset in particular.

LexLuthr: https://github.com/filecoin-project/curio/blob/547a713b148c05ae319ad5dbb7700afdbb0f286c/tasks/indexing/task_ipni.go#L312 is where we take the data in ipni task in main. We are reading a channel here. This channel is fed by https://github.com/filecoin-project/curio/blob/547a713b148c05ae319ad5dbb7700afdbb0f286c/tasks/indexing/task_indexing.go#L438 which uses [sourceOffset](https://github.com/filecoin-project/curio/blob/547a713b148c05ae319ad5dbb7700afdbb0f286c/tasks/indexing/task_indexing.go#L461)

LexLuthr: In this PR, ipni task is using separate reading mechanism so it needs to be fixed in both places. In main there is only 1 function to read the actual indexes.

rvagg (approved):
nothing fatal in here, this seems pretty good; but I've left a bunch of inline suggestions, giving approval in the assumption that those will be reasonably considered at least

tasks/indexing/task_pdp_ipni.go
rvagg: oh, sorry, I wasn't looking on `main`, I saw the duplicates on `pdpv0` but you've already squashed it down to one place; my bad but I now understand what you were talking about over there

other issue about `sectionSize` vs `40` still stands here, maybe in `main` too

harmony/harmonydb/sql/20240823-ipni.sql
```diff
@@ -8,7 +8,7 @@ CREATE TABLE ipni_peerid (
 CREATE TABLE ipni (
     order_number BIGSERIAL PRIMARY KEY, -- Unique increasing order number
     ad_cid TEXT NOT NULL,
-    context_id BYTEA NOT NULL, -- abi.PieceInfo in Curio
+    context_id BYTEA NOT NULL, -- abi.PieceInfo || PDPIPNIContext in Curio
```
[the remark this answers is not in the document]
LexLuthr: Already discussed with Spark about this. They agreed with the new contextID.

tasks/indexing/task_pdp_ipni.go
```diff
@@ -0,0 +1,427 @@
...
+var _ harmonytask.TaskInterface = &PDPIPNITask{}
+var _ = harmonytask.Reg(&PDPIPNITask{})
+
+// PdpIpniContext is used to generate the context bytes for PDP IPNI ads
+type PdpIpniContext struct {
+	// PieceCID is piece CID V2
+	PieceCID cid.Cid
+
+	// Payload determines if the IPNI ad is TransportFilecoinPieceHttp or TransportIpfsGatewayHttp
+	Payload bool
```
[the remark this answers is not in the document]
LexLuthr: Multiple reasons:
1. Early on I was told, we should be able to publish IPFS style and piece multihash only ads for each piece in PDP. So, pieceCID will be same in that case. I am not sure what you mean by "PieceCID is going to be different for payload vs piece anyway"
2. If we decide to some kind of checker stuff down the line (Spark), this allows decoding and making sense. Better safe then sorry as it will require another ad chain migration which is not cheap.

Issue #679, opened 2025-10-07
Update Curio to use the M3 FWSS contracts


rjan90:
This is a tracking issue for updating curio to use the M3 FWSS Calibnet Contracts

rjan90:
Closing as the Calibnet FWSS GA contracts are in the `pdpNext` branch

PR #680, opened 2025-10-07
chore(pdp): M3 Calibnet contract addresses and ABIs

rjan90:
Closes: #679


wjmelements:
> @wjmelements I see that we have checked in some .json files here https://github.com/filecoin-project/curio/blob/main/pdp/contract/IPDPProvingSchedule.json, are those actually needed with the above approach?

The json files seem redundant and can be removed. I don't think anyone is using them except for their abi. If we're getting the abi from somewhere else we can remove them. But you should also remove their makefile recipe, replacing it with a comment on where they come from.

wjmelements (approved):

PR #684, opened 2025-10-07
fix: index migration and IPNI canAccept

LexLuthr:
I will be dropping index migration due to high op cost for now. This PR will add some conversion to allow v1 and v2 lookup both.


market/indexstore/indexstore.go
magik6k, line 602: When is this expected to run? This function reads like it's immensely expensive to run.

Issue #678, opened 2025-10-07
PDPv0: uploadAndCreate


rvagg:
Operation supported in PDP, we need a way to use it in Curio. We may be able to just extend `POST /pdp/data-sets/` to take an optional `pieces` property that mirrors what we provide to `handleAddPieceToDataSet`,  rather than making a new endpoint.

BigLep:
@rjan90 @jennijuju : what milestone is this part of?  Is it M3 or M4?  I assume M3, but it affects https://github.com/FilOzone/synapse-sdk/issues/264 which is relevant for the Filecoin Pin website.  For now, I"m going to assume M4.

rvagg:
@virajbhartiya is keen to tackle this.

LexLuthr:
If we decide to reuse the current endpoint, then modify the endpoint and [proofset_create_watch.go](https://github.com/filecoin-project/curio/blob/pdpv0/tasks/pdp/proofset_create_watch.go) to have the new functionality.

PR #685, opened 2025-10-07
chore: fix breaking changes due to PDP v2.2.0

rjan90:
On top of https://github.com/filecoin-project/curio/pull/680

- Adapt to PDP v2.2.0 breaking changes for dataset creation
- Re-enable and update createDataSetCmd to use combined dataset creation + piece addition flow

### DIFF
--- a/cmd/pdptool/main.go
+++ b/cmd/pdptool/main.go
@@ -66,8 +66,7 @@ func main() {
 			uploadFileCmd,   // upload a file to a pdp service in many chunks
 			downloadFileCmd, // download a file from curio
 
-			// createDataSetCmd,    // REMOVED in PDP v2.2.0 - createDataSet() no longer exists
-			// Datasets are now created implicitly when adding the first piece via addPieces
+			createDataSetCmd,    // Create dataset with initial pieces (combined flow)
 			getDataSetStatusCmd, // get the status of a data set creation on the PDP service
 			getDataSetCmd,       // retrieve the details of a data set from the PDP service
 


cmd/pdptool/main.go
```diff
@@ -885,6 +886,10 @@ var uploadFileCmd = &cli.Command{
 	},
 }
 
+// BREAKING CHANGE: createDataSet has been removed in PDP v2.2.0
+// Datasets are now created implicitly when adding the first piece via addPieces
+// To create a dataset, use the add-pieces command with your first piece
+/* DISABLED - createDataSetCmd removed in PDP v2.2.0
 var createDataSetCmd = &cli.Command{
```
LexLuthr, line 893: We should probably update this command instead of deleting it as there is no way to create dataSet now

rjan90: Makes sense, I have updated the createDataSetCmd in https://github.com/filecoin-project/curio/pull/685/commits/7666fb428d8c4f216a96ea8976a417f3bdef3f02.

rvagg:
Sadly this isn't going to work and it's going to require more surgery than this.

`handleGetDataSetCreationStatus` requires a data set id and for it to exist in `pdp_data_sets`, which will fail (chicken and horse).

There's 2 paths here:

1. Data set doesn't exist, first pieces should create it
2. Data set exists, add pieces to that data set

In https://github.com/filecoin-project/curio/issues/678 I suggested that we should keep the `POST /pdp/data-sets` (`handleCreateDataSet`) to _create and add pieces_, but the alternative is to delete that and use a data set ID of `0` to indicate "not created yet" and use `POST /pdp/data-sets/0/pieces` (this is the one that leads to `handleAddPieceToDataSet`) to do both jobs - and the fact that it's `0` is a special case that causes it to skip the "do I have this?" check and instead flip into "create+add" mode. Otherwise if it's a valid ID that I have in my db then it's just "add" mode.

I don't really mind which way, perhaps it's a little cleaner if we just delete `handleCreateDataSet` and go with `0` rather than the path I outlined in #678 and @virajbhartiya started on in https://github.com/filecoin-project/curio/pull/687.

So I think that @virajbhartiya you should start on this branch and make it work with new data sets, where `0` is a special case. Trace through `handleAddPieceToDataSet` and figure out where it's going to fail with `0`.

But you also need to get an end-to-end setup going so you can try this out with the new contracts that it's supposed to work with and Synapse that it's supposed to interact with so we're not just guessing that it might work.

PR #692, opened 2025-10-09
feat(pdp): support create and add endpoint

Kubuxu:


tasks/pdp/proofset_addroot_watch.go
```diff
@@ -111,6 +111,30 @@ func processDataSetPieceAdd(ctx context.Context, db *harmonydb.DB, ethClient *et
...
+				WHERE id = $1
+			)`, resolvedDataSetId.Int64).Scan(&exists)
+		if err != nil {
+			return fmt.Errorf("failed to check if data set exists: %w", err)
+		}
+		if !exists {
+			// XXX: maybe return nil instead to avoid warning?
+			return fmt.Errorf("data set %d not found in pdp_data_sets", resolvedDataSetId.Int64)
+		}
+	}
```
LexLuthr: I would prefer not try to create_watch here as well. Just scan from DB and if it is NULL then just skip processing addPiece for this tipset. But that might cause unnecessary delay of 1 tipset. Any other ideas do this this without 2 watches?

Kubuxu: > Just scan from DB and if it is NULL then just skip processing addPiece for this tipset

That is what I'm doing here. If we get an add_piece triggered, and the dataset doesn't exist yet, we leave it alone and wait for another trigger for when the dataset is created.
A cleaner way might be to get `create_watch` to be processed with a higher priority than `add_pieces`. This could be achieved by combining them into one watcher and executing sequentially with create first.

pdp/contract/PDPVerifier.go
```diff
@@ -197,6 +197,37 @@ func (_PDPVerifier *PDPVerifierTransactorRaw) Transact(opts *bind.TransactOpts,
 	return _PDPVerifier.Contract.contract.Transact(opts, method, params...)
 }
 
+// EXTRADATAMAXSIZE is a free data retrieval call binding the contract method 0x029b4646.
```
rvagg: this is a different ABI to the 3.0.0 that got tagged where these were removed: https://github.com/FilOzone/pdp/commits/v3.0.0/src/PDPVerifier.sol

I guess you need to update/sync something?

pdp/handlers_create.go
```diff
@@ -4,16 +4,150 @@ import (
...
+
+	extraDataBytes, err := decodeExtraData(reqBody.ExtraData)
+	if err != nil {
+		http.Error(w, "Invalid extraData format (must be hex encoded)", http.StatusBadRequest)
+		return
+	}
+
+	pieceDataArray, subPieceInfoMap, _, err := p.transformAddPiecesRequest(ctx, serviceLabel, reqBody.Pieces)
+	if err != nil {
+		http.Error(w, "Failed to transform addPieces request: "+err.Error(), http.StatusBadRequest)
```
rvagg:
```suggestion
		http.Error(w, "Failed to process request: "+err.Error(), http.StatusBadRequest)
```

"transform" isn't helpful to the user here I think

pdp/handlers_add.go
```diff
@@ -350,6 +247,106 @@ func (p *PDPService) handleAddPieceToDataSet(w http.ResponseWriter, r *http.Requ
...
+			http.Error(w, "Invalid extraData format (must be hex encoded)", http.StatusBadRequest)
+			return
+		}
+		extraDataBytes = decodedBytes
+	}
+
+	// Step 4: Prepare piece information
+	pieceDataArray, subPieceInfoMap, subPieceCidList, err := p.transformAddPiecesRequest(ctx, serviceLabel, payload.Pieces)
+	if err != nil {
+		http.Error(w, "Failed to transform add pieces request: "+err.Error(), http.StatusBadRequest)
```
rvagg:
```suggestion
		http.Error(w, "Failed to process request: "+err.Error(), http.StatusBadRequest)
```

pdp/handlers_add.go
```diff
@@ -350,6 +247,106 @@ func (p *PDPService) handleAddPieceToDataSet(w http.ResponseWriter, r *http.Requ
...
+	if payload.ExtraData != nil {
+		extraDataHexStr := *payload.ExtraData
+		decodedBytes, err := hex.DecodeString(strings.TrimPrefix(extraDataHexStr, "0x"))
+		if err != nil {
+			log.Errorf("Failed to decode hex extraData: %v", err)
+			http.Error(w, "Invalid extraData format (must be hex encoded)", http.StatusBadRequest)
+			return
+		}
+		extraDataBytes = decodedBytes
+	}
```
rvagg: there's a `decodeExtraData` now for this

tasks/pdp/proofset_addroot_watch.go
```diff
@@ -11,47 +12,30 @@ import (
...
-	DataSet         uint64 `db:"data_set"`
-	Piece           string `db:"piece"`
-	AddMessageHash  string `db:"add_message_hash"`
-	AddMessageIndex uint64 `db:"add_message_index"`
-	SubPiece        string `db:"sub_piece"`
-	SubPieceOffset  int64  `db:"sub_piece_offset"`
-	SubPieceSize    int64  `db:"sub_piece_size"`
-	PDPPieceRefID   int64  `db:"pdp_pieceref"`
-	AddMessageOK    *bool  `db:"add_message_ok"`
-	PDPDataSetId    uint64 `db:"data_set"`
```
rvagg: hah, what was going on here ..

rvagg:
looks fine overall, just a bunch of relatively minor items in this review pass

rvagg (approved):

PR #700, opened 2025-10-10
feat: Option to specify custom client address in mk12 deal proposals

kacperzuk-neti:
## Context

FIDL has introduced client smart contracts for use in Fil+. Instead of assigning DataCap directly to clients, it's now assigned to a smart contract. This provides more control over how the DataCap is used and allows allocators to revoke it from clients without needing to involve RKH. This system is already live on mainnet for DDO.

However, this introduces a challenge: from the perspective of the Market actor, the smart contract is the client. The Market actor will call the smart contract to authenticate the deal, but the contract still needs a way to verify that the *actual* client (the one using the DataCap) initiated the deal, so it can correctly deduct their allowance.

## This PR

This pull request proposes a general solution: a new option that allows specifying the client address in the deal proposal separately from the wallet used to sign it.

This means a client can sign the proposal with their own wallet, but set the `clientAddress` field to point to the smart contract.

The change is fully backwards compatible. If the `client-address` option is not provided, the system will default to using the signing wallet address, which is the current behavior.

### DIFF
--- a/cmd/sptool/toolbox_deal_client.go
+++ b/cmd/sptool/toolbox_deal_client.go
@@ -289,7 +289,11 @@ var dealFlags = []cli.Flag{
 	},
 	&cli.StringFlag{
 		Name:  "wallet",
-		Usage: "wallet address to be used to initiate the deal",
+		Usage: "wallet address to be used to sign the deal",
+	},
+	&cli.StringFlag{
+		Name:  "client-address",
+		Usage: "address to be used to initiate the deal; if empty, the signing wallet will be used",
 	},
 	&cli.BoolFlag{
 		Name:  "skip-ipni-announce",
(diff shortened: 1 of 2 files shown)


LexLuthr (changes requested):
I don't understand what this PR is doing? Why do we need a separate flag to specify a different wallet here?

LexLuthr:
In case this is not obvious, the client field in the proposal must have signed the proposal otherwise they will be rejected. It seems like you are signing with a different wallet here.

~~I still don't understand why do we insist on making the contract as client?~~ It seems like verifreg design doesn't allow for it.

Also, this tool is not meant to be used in any production environment. It is a testing tool for Curio team and devnet testing.

kacperzuk-neti:
> In case this is not obvious, the client field in the proposal must have signed the proposal otherwise they will be rejected. It seems like you are signing with a different wallet here.

Not completely accurate. Market actor doesn't validate the signature directly, [it calls](https://github.com/filecoin-project/builtin-actors/blob/master/actors/market/src/lib.rs#L1698) `authenticateMessage` on the client. For accounts it indeed makes sure that signature matches the address, but for smart contracts it can be implemented in any other way. For our use-case we want to validate it against other known pubkey.

> It seems like verifreg design doesn't allow for it.

This was tested and core flow works. Both Verifreg and Market are OK with it as long as the contract implements proper methods.

> Also, this tool is not meant to be used in any production environment. It is a testing tool for Curio team and devnet testing.

Ah, I wasn't aware of that. In that case, please disregard this PR. Thanks!

PR #709, opened 2025-10-14
pdp(chore): extract upload and create handlers into separate files

Kubuxu:
To reduce conflicts with #692


LexLuthr (approved):
Good idea!

PR #710, opened 2025-10-15
Apply dyn-cfg to p1 cases

snadrus:
Make our Ingest & Address configuration dynamic such that the values apply instantly. 

 Market will ERROR on 0 <--> 1 that a restart is required to enable the correct services. 

__Discoveries__
 Computed Values: These become their own config.NewDynamic() which gets updated by a callback on the source(s). 
    I could be persuaded to turn these into event streams into reactive programming: https://github.com/samber/ro

I don't think we should be stopping & starting services on these changes. That will simply be unsupported. 

Bad configuration is setup to throw error and reuse the original in all the secondary-interpretations, otherwise it'll stream out errors until fixed in the direct consumptions. These seem reasonable. 

I couldn't get TOML to do value unmarshal as it knows documents only at the top level, so I built an ugly wrapper full of reflection that we should use instead of TOML unmarshal. It works, but it's ugly. 

BTW, also made descriptions visible (was a mouseover). This will make it easier to see what's dynamic and what's not. 
image

### DIFF
--- a/web/api/config/config.go
+++ b/web/api/config/config.go
@@ -77,14 +77,14 @@ func getSch(w http.ResponseWriter, r *http.Request) {
 		},
 	}
 	sch := ref.Reflect(config.UnwrapDynamics(config.CurioConfig{}))
-	// add comments
-	for k, doc := range config.Doc {
-		item, ok := sch.Definitions[k]
-		if !ok {
-			continue
+
+	// Helper to add comments to a schema's properties
+	addComments := func(targetSchema *jsonschema.Schema, docEntries []config.DocField) {
+		if targetSchema == nil || targetSchema.Properties == nil {
+			return
 		}
-		for _, line := range doc {
-			item, ok := item.Properties.Get(line.Name)
+		for _, line := range docEntries {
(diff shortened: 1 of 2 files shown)


tasks/storage-market/market_balance.go
magik6k: We have f05 support in Wallet Manager, this is the legacy version of that - I'd just drop it

deps/config/TRANSPARENT_WRAPPER_MIGRATION.md
magik6k: Agent report?

snadrus: yes, seemed mildly interesting at the time, but I should just remove it.

tasks/storage-market/market_balance.go
snadrus: @LexLuthr https://github.com/filecoin-project/curio/pull/431 just added this. Can you weigh in?

LexLuthr:
We should Test this with UI and see if new encode/decode works with our weird UI setup.

magik6k (approved):

PR #725, opened 2025-10-23
fix: edge case for proving subpiece

LexLuthr:
We seems to be missing the edge case where challenged leaf's index is 0.


rvagg (approved):
would also impact when the offset is right at the start of a subpiece too, yeah? we only experience this as the "first" because we only use one subpiece so `0` is the only way to trigger it now, but if we had 2 subpieces then I suppose that if we were challenged at the start of that piece then we'd hit this as well?

PR #755, opened 2025-10-29
guided-setup needs a fudge factor to create miners

snadrus:
DO NOT USE until merged (and therefore tested):This may drain your wallet!

1.5 factor hard-coded


lib/createminer/create_miner.go
```diff
@@ -29,6 +29,8 @@ type CreateMinerChainAPI interface {
 	StateMinerCreationDeposit(ctx context.Context, tsk types.TipSetKey) (types.BigInt, error)
 }
 
+const DepositMarginFactor = 1.5
```
beck-8: why not use it below ？

Issue #761, opened 2025-10-29
pdp: division by zero in PDP Proving task


Kubuxu:
Possible cause is an empty dataset, possible dataset IDs 1, 11, 12
based on `select * from curio.pdp_data_sets WHERE init_ready = TRUE AND challenge_request_msg_
 hash is NULL`
```
Recovered from a serious error while processing PDPProve task 105710: division by zero Stack: goroutine 15174 [running]:
github.com/filecoin-project/curio/harmony/harmonytask.(*taskTypeHandler).considerWork.func3.1()
	/home/kubuxu/curio/harmony/harmonytask/task_type_handler.go:200 +0xd4
panic({0x437fe80?, 0x52395f0?})
	/home/kubuxu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.6.linux-amd64/src/runtime/panic.go:792 +0x132
math/big.nat.div({0x0?, 0xda5a924f2910ed0?, 0x4d7daa?}, {0x0?, 0x7bb0b10d6148a194?, 0x31dda9a7a3db3c52?}, {0xc001384500?, 0xc001162d38?, 0x4d8106?}, {0x0, ...})
	/home/kubuxu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.6.linux-amd64/src/math/big/natdiv.go:520 +0x32e
math/big.(*Int).QuoRem(0xc001162d50, 0xc001347700, 0xc001347720, 0xc001347740)
	/home/kubuxu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.6.linux-amd64/src/math/big/int.go:293 +0x7b
math/big.(*Int).Mod(0xc001347740, 0x0?, 0x0?)
	/home/kubuxu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.6.linux-amd64/src/math/big/int.go:324 +0x217
github.com/filecoin-project/curio/tasks/pdp.generateChallengeIndex({0xc000c05740, 0x20, 0x21}, 0xb, 0x0, 0x0)
	/home/kubuxu/curio/tasks/pdp/task_prove.go:397 +0x405
github.com/filecoin-project/curio/tasks/pdp.(*ProveTask).GenerateProofs.func1(...)
	/home/kubuxu/curio/tasks/pdp/task_prove.go:347
github.com/samber/lo.Times[...](...)
[...]

PR #765, opened 2025-10-30
fix(pdp): don't try proving empty datasets

Kubuxu:
Empty datasets set proving epoch to zero (NO_CHALLENGE_SCHEDULED).
Number of leaves is also zero, which was causing https://github.com/filecoin-project/curio/issues/761 .

This creates a bit of an edge case when piece deletions, additions and proving happen in the same time window:
- data set gets used, pieces are added and proven
- all pieces are deleted from it
- nextProvingPeriod gets called on an empty dataset
- a new piece gets added, it sets `init_ready = TRUE`, but it is true already
- prove task fires, detects that proving set is empty, challenge epoch is 0 (as the proving set is empty),
		proving gets disabled
Now the dataset won't get proven until one more piece gets added to set `init_ready = TRUE`.
Better pattern here would be to react to events emitted in our messages from the transactions we send to PDPVerifier.
As ordering can get even more tricky if you consider that transactions are sent async.

EDIT:
We also need to stop calling `nexProvingPeriod` and relying on `NextPDPChallengeWindowStart` giving us a value, because these don't work when the dataset is empty. This is why I'm kicking the dataset out of the proving cycle.

### DIFF
--- a/tasks/pdp/task_prove.go
+++ b/tasks/pdp/task_prove.go
@@ -170,7 +170,7 @@ func (p *ProveTask) disableProving(ctx context.Context, dataSetId int64) error {
 	// As ordering can get even more tricky if you consider that transactions are sent async.
 	_, err := p.db.Exec(ctx, `
 		UPDATE pdp_data_sets
-		SET challenge_request_msg_hash = NULL, prove_at_epoch = NUL, init_ready = FALSE,
+		SET challenge_request_msg_hash = NULL, prove_at_epoch = NULL, init_ready = FALSE,
 			prev_challenge_request_epoch = NULL
 		WHERE id = $1
 		`, dataSetId)


tasks/pdp/task_prove.go
```diff
@@ -154,13 +154,40 @@ func NewProveTask(chainSched *chainsched.CurioChainSched, db *harmonydb.DB, ethC
...
+	// 		proving gets disabled
+	// Now the dataset won't get proven until one more piece gets added to set `init_ready = TRUE`.
+	// Better pattern here would be to react to events emitted in our messages from the transactions we send to PDPVerifier.
+	// As ordering can get even more tricky if you consider that transactions are sent async.
+	_ , err := p.db.Exec(ctx, `
+		UPDATE pdp_data_sets
+		SET challenge_request_msg_hash = NULL, prove_at_epoch = NUL, init_ready = FALSE,
+			prev_challenge_request_epoch = NULL
+		WHERE id = $1
+		`, dataSetId)
```
LexLuthr: I think this will cause PPInit every time we enable and disable it. I am pretty sure that would revert from contract

LexLuthr (commented):
We can just skip proving if total leaves are 0 instead of enable/disable which is likely to have some edge case as you already stated in the comment.

PR #789, opened 2025-11-06
Upgrade to Gcc12 Cuda13

snadrus:
SupraSeal BETA support in Curio (a long-running Beta) needs to become stabilized as we plan to build atop it for other non-beta features. 

This brings it into compatibility with the latest Ubuntu & Cuda (and it's held-back gcc12 while 14 is current).


magik6k (approved):
Amazing to see CI for this

PR #791, opened 2025-11-07
chore(pdp): fix log initPP, fix error in in asPieceCIDv1

Kubuxu:
Just misc error and log output fixes


Kubuxu:
Ahhh, I closed the wrong PR

PR #799, opened 2025-11-13
return 404 for piece retrievals correctly

LexLuthr:


lib/cachedreader/cachedreader.go
```diff
@@ -283,18 +285,37 @@ func (cpr *CachedPieceReader) GetSharedPieceReader(ctx context.Context, pieceCid
...
+				if retCode == http.StatusNotFound && !errors.Is(err, ErrNoDeal) {
+					retCode = http.StatusInternalServerError
+				}
+
+				var finalErr error
+
+				if retCode == http.StatusNotFound {
+					finalErr = xerrors.Errorf("failed to get piece reader from sector or piece park: %w, %w, %w", err, serr, ErrNotFound)
+				} else {
+					finalErr = xerrors.Errorf("failed to get piece reader from sector or piece park: %w, %w", err, serr)
```
Kubuxu: You cannot do double/triple `%w %w` with xerrors. fmt.Errorf supports it

PR #806, opened 2025-11-17
Unified Snap

snadrus:
This addition to the fast snap mainline modifies supraseal:
- No more objcopy: An interface glues the filereader with the NVMe driver.
- Faster build using parallelism
- AMD64v3 becomes the default, as Curio has had for a while (2013-2015 era introduction). 
- Added a fallback for CPUs without SHA-NI (v3 but not v4) to be able to use Supraseal. 

This should make fastsnap the only path needed for Curio. 

- mbx2 is the crazy SHA for Zen: supraseal/sha/sha_mbx_wrapper.cpp

__Cases__
 old libfilecoin-ffi snap:
- non-Linux: mbx2 fallback 
- Linux Snap without CUDA
- Linux before AMD64v4

Use supraseal fastsnap: 
- Linux with Cuda with AMD64v4

Use supraseal PC1:
- Linux + CUDA + Zen only. TODO!!!!!!!: Fix mbx2 to run only in this case. 

DEBs to cover all cases:
1. OPENGL
2. CUDA
3. ZEN

---

> [!NOTE]
> Unifies SupraSeal snap path with CPU/GPU fallback, adds a `curio batch setup` SPDK command, refactors/builds the SupraSeal library, and overhauls CI/Docker to Ubuntu 24.04 with GCC 12 and CUDA.
> 
> - **SupraSeal core/library**:
>   - Unifies PC2 via templated readers (NVMe and file) and removes objcopy/symbol hacks.
>   - Parallelizes build (Poseidon constants, objects); adjusts NVCC flags; builds `libsupraseal.a` (no tool binaries in CI).
>   - Adds CUDA-only TreeR path with CPU/GPU feature detection; fallback to `filecoin-ffi` when unsupported.
>   - Fixes memory delete (`delete[]`) in `c1/challenge.hpp`.
> - **CLI/runtime**:
>   - Adds `curio batch setup` to auto-configure SPDK (hugepages, NVMe binding).
>   - Auto SPDK setup and hugepage checks in batch task; supports `DISABLE_SPDK_SETUP` env.
> - **Go bindings** (`lib/supraffi`, `ffidirect`):
>   - Splits NVMe vs generic CUDA paths; adds CPU feature and CUDA probing; TreeR fallback in FFI wrapper.
> - **Build/Makefile**:
>   - Makes SupraSeal a default Linux build dep; sets `GOAMD64=v3`; removes legacy batch targets.
> - **CI/Docker**:
>   - Migrates to `ubuntu-24.04`, replaces custom Go action, sets GCC 12 via alternatives.
>   - Installs CUDA 12+, configures stubs and env; recursive submodules; caches proof params.
>   - Adds dedicated SupraSeal library build job/artifact and local test Dockerfile.
> - **Docs**:
>   - Updates install deps (adds `libgmp-dev`, `libconfig++-dev`), switches build instructions to `make curio/sptool`.
>   - Documents `curio batch setup` and new env var in EN/ZN SupraSeal guides.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 23c6b6afd705fd27b347d404de08d32dac5e5efc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lib/ffiselect/ffidirect/ffi-direct.go
+++ b/lib/ffiselect/ffidirect/ffi-direct.go
@@ -98,15 +98,15 @@ func (FFI) TreeRFile(lastLayerFilename, dataFilename, outputDir string, sectorSi
 		// Convert sector size to RegisteredPoStProof (WindowPoSt version)
 		var postProof abi.RegisteredPoStProof
 		switch sectorSize {
-		case 34359738368: // 32GiB
+		case 32 << 30: // 32GiB
 			postProof = abi.RegisteredPoStProof_StackedDrgWindow32GiBV1_1
-		case 68719476736: // 64GiB
+		case 64 << 30: // 64GiB
 			postProof = abi.RegisteredPoStProof_StackedDrgWindow64GiBV1_1
-		case 2147483648: // 2KiB
+		case 2 << 10: // 2KiB
 			postProof = abi.RegisteredPoStProof_StackedDrgWindow2KiBV1_1
-		case 8388608: // 8MiB
+		case 8 << 20: // 8MiB
 			postProof = abi.RegisteredPoStProof_StackedDrgWindow8MiBV1_1
-		case 536870912: // 512MiB
+		case 512 << 20: // 512MiB
 			postProof = abi.RegisteredPoStProof_StackedDrgWindow512MiBV1_1
 		default:
 			return xerrors.Errorf("unsupported sector size for TreeR fallback: %d", sectorSize)
--- a/lib/supraffi/cuda_linux.go
+++ b/lib/supraffi/cuda_linux.go
@@ -33,15 +33,13 @@ int main() {
 func HasUsableCUDAGPU() bool {
 	cudaProbeOnce.Do(func() {
 		cudaProbeOK = detectCUDAWithNvcc()
+		logger.Infof("cuda probe result: %t", cudaProbeOK)
 	})
 
 	return cudaProbeOK
 }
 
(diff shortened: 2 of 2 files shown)


lib/supraffi/spdk_setup.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	spdkLog.Info("SPDK setup completed successfully")
+	return nil
+}
+
+// CheckAndSetupSPDK checks if SPDK is set up, and if not, runs the setup script.
+// This is a convenience function that combines checking and setup.
+// It ensures SPDK is downloaded and available, then runs the setup script if hugepages aren't configured.
+func CheckAndSetupSPDK(nrHuge int, minPages int) error {
```
magik6k: I think this should be optional (default enabled but optional) so that more advanced deployments can do this with customization (e.g. controlling hugepage-numa assignment, controlling exact nvme used, etc.)

snadrus: done

extern/supraseal/sha/sha_ext_mbx2_wrapper.cpp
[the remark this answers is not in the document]
magik6k: Is V3 == -march=zenver2 or are those kinda different?

snadrus:
Cannot proceed until the self-hosted runners get to Ubuntu 24.04 (currently 20.04 which was also referred to as "ubuntu-latest")

PR #807, opened 2025-11-18
feat(pdp): enforce extraData size limits

### LINKED ISSUE #713: PDP: implement EXTRA_DATA_MAX_SIZE limit
## Context

The `EXTRA_DATA_MAX_SIZE` limit has been removed from PDPVerifier (FilOzone/pdp#224) as it should be enforced at the service level instead. This allows different services to set appropriate limits based on their specific requirements.

## Task

Implement an `EXTRA_DATA_MAX_SIZE` limit in Curio to prevent excessively large metadata payloads. This should validate extraData size before calling PDPVerifier functions.

Recommended limits: https://github.com/FilOzone/pdp/issues/224#issuecomment-3420907995

## References

- PDPVerifier removal: FilOzone/pdp#224

siddharthbaleja7:
## Description
This PR implements service-level validation for the `extraData` payload size in PDP (Proof of Data Possession) requests. Following the removal of the hard limit in the PDPVerifier contract (FilOzone/pdp#224), this logic is now enforced within Curio to prevent excessively large metadata payloads from being processed. I have verified these changes locally to ensure that the limits are enforced correctly and oversized payloads are rejected.

Fixes #713

### 1. Defined Size Limits
* **`MaxCreateDataSetExtraDataSize`**: Set to **4KB** (4096 bytes) for `createDataSet`.
* **`MaxAddPiecesExtraDataSize`**: Set to **8KB** (8192 bytes) for `addPieces` and `createDataSetAndAddPieces`.

### 2. Validation Logic (Fail Fast)
Implemented size checks in the following handlers:
* `handleCreateDataSet`
* `handleCreateDataSetAndAddPieces`
* `handleAddPieceToDataSet`

## Checklist
- [x] Code compiles correctly
- [x] Changes verified locally
- [x] Limits match recommendations from FilOzone/pdp#224

### DIFF
--- a/pdp/handlers.go
+++ b/pdp/handlers.go
@@ -877,7 +877,8 @@ func (p *PDPService) handleDeleteDataSetPiece(w http.ResponseWriter, r *http.Req
 			return
 		}
 		if len(extraDataBytes) > MaxDeletePieceExtraDataSize {
-			http.Error(w, "extraData too long (max 256 bytes)", http.StatusBadRequest)
+			errMsg := fmt.Sprintf("extraData size (%d bytes) exceeds the maximum allowed limit for DeletePiece (%d bytes)", len(extraDataBytes), MaxDeletePieceExtraDataSize)
+			http.Error(w, errMsg, http.StatusBadRequest)
 			return
 		}
 	}


pdp/handlers.go
```diff
@@ -865,6 +865,10 @@ func (p *PDPService) handleDeleteDataSetPiece(w http.ResponseWriter, r *http.Req
 			http.Error(w, "Invalid extraData format (must be hex encoded)", http.StatusBadRequest)
 			return
 		}
+		if len(extraDataBytes) > 256 {
```
rvagg: needs a constant like the other two please

but now they are spread across two files, how about you move the constants into this file, handlers.go, since it's the parent file of the set.

PR #813, opened 2025-11-24
feat(pdp): add idempotency to create and add handlers

Kubuxu:
Introduces a `IdempotencyKey` to JSON paylods of these handlers, which
deboounces requests and if the request with the same key completed,
redirects to the created resource.

Users should use UUIDs or ULIDs as their IdempotencyKey-s

### DIFF
--- a/pdp/idempotency.go
+++ b/pdp/idempotency.go
@@ -11,12 +11,8 @@ import (
 	"regexp"
 	"time"
 
-	logger "github.com/ipfs/go-log/v2"
-
 	"github.com/filecoin-project/curio/harmony/harmonydb"
 )
-
-var logIdempotency = logger.Logger("pdp/idempotency")
 
 // Ensure harmonydb import is used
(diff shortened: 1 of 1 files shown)


Kubuxu:
Yeah, I was going to create a more expansive issue for this.
I'm mainly targeting the "bad network" case. We tried sending a request, but didn't get a response back for some reason. It will allow us to retry the same request without triggering side effects if the first request was received, but we failed to receive a response.
The key is generated on the client side and reused if the request is equivalent to one that has already been sent.
It can be "the request didn't go through due to network issues" or "application crashed between the steps of sending the request and saving the transaction hash" (if we were going for very strict transactionality on the client side).

This should be an implementation detail ideally not exposed to the end user. In my initial stab at this, I was considering using the extra data payload as the key, but this is less constraining. @hugomrdias might have more ideas on how to utilise it effectively within Synapse, as this is something we discussed back when we met.

PR #820, opened 2025-12-02
more efficient CI

snadrus:
No production impact!

Lets have CI complete faster & more accurately:
1. the different builds actually build the right tags now
2. cache stuff that rarely changes: Per-PR: deps, ffi, Global: params
3. don't waste CI resources doing the same work in parallel
5. Fixes: some old idempotent SQL matters
6. in itest:
 - unit-test & itest[1,2] run separately, some using parallelism
 - share yugabyte
 - only upgrade 1  database, then copy to others.(1s vs 48s each)
 
image

### DIFF
--- a/.github/actions/setup-build-env/action.yml
+++ b/.github/actions/setup-build-env/action.yml
@@ -28,13 +28,17 @@ runs:
       run: go mod download
       shell: bash
 
-    # Cache FFI build based on submodule commit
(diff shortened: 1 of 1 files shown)


LexLuthr:
@snadrus Shall we close this?

snadrus:
> @snadrus Shall we close this?

Let's be sure all tests run, and yes

PR #854, opened 2025-12-29
Accept Task List

snadrus:
HarmonyTask can now accept a set of tasks at once. CanAccept indicates it wants a set of tasks (instead of just one) and the machine accepts as many as it has capacity. This is meant to improve task acceptance at scale.

The code to accept a group of tasks was optimized to avoid locked task rows (almost certainly being updated to claimed), so a single call claims as many as possible and returns that list. 

Storage headroom is attempted as an allocation for each task (as before), but any number 1 or greater proceeds to find a task for that storage lease. 

No list caching is done as there's a possibility that something more important will arrive (fail out from elsewhere) and need priority. 

CanAccept's List ordering is respected such that we get the first N (headroom) tasks that are currently unclaimed. 

All GOTOs are removed as we call CanAccept() once max from considerWork().

Balancing concern: we now have a higher risk of starvation if task ordering is poor.

### DIFF
--- a/tasks/pdp/pdp_sync.go
+++ b/tasks/pdp/pdp_sync.go
@@ -159,8 +159,8 @@ func (P *PDPSyncTask) Do(taskID harmonytask.TaskID, stillOwned func() bool) (don
 	return true, nil
 }
 
-func (P *PDPSyncTask) CanAccept(ids []harmonytask.TaskID, engine *harmonytask.TaskEngine) (*harmonytask.TaskID, error) {
-	return &ids[0], nil
+func (P *PDPSyncTask) CanAccept(ids []harmonytask.TaskID, engine *harmonytask.TaskEngine) ([]harmonytask.TaskID, error) {
+	return ids, nil
 }
 
 func (P *PDPSyncTask) TypeDetails() harmonytask.TaskTypeDetails {


snadrus:
@Reiers give this a try on calibration-net and let me know how it goes. Watch the first 20 minutes to see if task processing halts.

harmony/harmonydb/userfuncs.go
```diff
@@ -12,11 +12,25 @@ import (
...
-const InitialSerializationErrorRetryWait = 5 * time.Second
+var backoffs = lo.Map([]int{200, 400, 800, 1600, 2400, 5000, 8000}, func(item int, _ int) time.Duration {
+	return time.Duration(item) * time.Millisecond
+})
+
+func backoffForSerializationError[T any](f func() (T, error)) (whatever T, err error) {
+	for _, backoff := range backoffs {
+		res, err := f()
+		if !IsErrSerialization(err) {
+			return res, nil
```
magik6k: I'm quite positive this should be res, err

harmony/harmonytask/task_type_handler.go
```diff
@@ -103,140 +102,170 @@ top:
...
 	}
-	if tID == nil {
-		log.Infow("did not accept task", "task_id", ids[0], "reason", "CanAccept() refused", "name", h.Name)
+	if len(tIDs) == 0 {
+		log.Infow("did not accept task", "task_ids", ids, "reason", "CanAccept() refused", "name", h.Name)
 		return false
 	}
 
-	releaseStorage := func() {
+	/// TODO FROM HERE handle multiple task IDs
```
magik6k: Not really TODO anymore?

harmony/harmonytask/task_type_handler.go
```diff
@@ -333,27 +362,43 @@ VALUES ($1, $2, $3, $4, $5, $6, $7, $8) RETURNING id`, tID, h.Name, postedTime.U
 	}
 }
 
-func (h *taskTypeHandler) AssertMachineHasCapacity() error {
+func (h *taskTypeHandler) AssertMachineHasCapacity() (int, error) {
 	r := h.TaskEngine.ResourcesAvailable()
-
+	headroom := 100
```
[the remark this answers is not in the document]
magik6k: Really I'd like to be able to set 'MaxTasksPerSchedule' in the config or at least as an envvar

PR #855, opened 2025-12-31
improve canAccept() performance

LexLuthr:


tasks/indexing/task_ipni.go
```diff
@@ -197,7 +187,7 @@ func (I *IPNITask) Do(taskID harmonytask.TaskID, stillOwned func() bool) (done b
 				PreviousID: cidlink.Link{Cid: prevCID},
 				Provider:   a.Provider,
 				Addresses:  strings.Split(a.Addresses, "|"),
-				Entries:    cidlink.Link{Cid: e},
+				Entries:    schema.NoEntries,
```
magik6k: Interesting, I assume it is supposed to be like this

market/indexstore/indexstore.go
```diff
@@ -847,3 +847,34 @@ func (i *IndexStore) GetPDPNode(ctx context.Context, pieceCidV2 cid.Cid, layerId
...
+
+func (i *IndexStore) GetMultihashes(ctx context.Context, piecev2 cid.Cid) (<-chan multihash.Multihash, <-chan error) {
+	out := make(chan multihash.Multihash, 1000)
+	errCh := make(chan error, 1)
+
+	go func() {
+		defer close(out)
+		defer close(errCh)
+
+		qry := "SELECT PayloadMultihash FROM PieceBlockOffsetSize WHERE PieceCid = ? ORDER BY PayloadMultihash ASC"
```
magik6k: This will blow up with larger indexes, the bulk GetMultihashes was one of the top issues in boost data indexing. There is ~90% chance this method will fail with more than >10M multihashes.

Also not a fan of loading everything into memory. Ideally we can just stream chained calls to `GetPieceHashRange` with EntryCount (iirc default 16k) entries, which will result in much less memory use

tasks/indexing/task_ipni.go
```diff
@@ -291,101 +281,29 @@ func (I *IPNITask) Do(taskID harmonytask.TaskID, stillOwned func() bool) (done b
...
-					found = true
-				}
-				if !found {
-					return false, xerrors.Errorf("no sub pieces for aggregate mk20 deal")
-				}
-				_, _, interrupted, err = IndexAggregate(pcid2, reader, pi.Size, subPieces, recs, addFail)
-			} else {
-				return false, xerrors.Errorf("invalid aggregate type")
-			}
+		serr := chk.Accept(rec)
```
magik6k: I'd make Accept just accept an array now, this for loop doesn't need to exist now, at least in this form

Really here we should be calling `idx.GetPieceHashRange` in a loop and call sth like `chk.NextEntry`

tasks/indexing/task_ipni.go
```diff
@@ -57,16 +52,18 @@ type IPNITask struct {
...
 	cfg           *config.CurioConfig
 	max           taskhelp.Limiter
+	idx           *indexstore.IndexStore
 }
 
-func NewIPNITask(db *harmonydb.DB, sc *ffi.SealCalls, pieceProvider *pieceprovider.SectorReader, cpr *cachedreader.CachedPieceReader, cfg *config.CurioConfig, max taskhelp.Limiter) *IPNITask {
+func NewIPNITask(db *harmonydb.DB, sc *ffi.SealCalls, pieceProvider *pieceprovider.SectorReader, cpr *cachedreader.CachedPieceReader, cfg *config.CurioConfig, max taskhelp.Limiter, idx *indexstore.IndexStore) *IPNITask {
 	return &IPNITask{
 		db:            db,
 		pieceProvider: pieceProvider,
```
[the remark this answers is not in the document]
LexLuthr: Done

tasks/indexing/task_ipni.go
```diff
@@ -291,101 +281,29 @@ func (I *IPNITask) Do(taskID harmonytask.TaskID, stillOwned func() bool) (done b
...
-		}
-	}
-
-	recs := make(chan indexstore.Record, 1)
-
-	var eg errgroup.Group
-	addFail := make(chan struct{})
-	var interrupted bool
-	var subPieces []mk20.DataSource
 	chk := chunker.NewInitialChunker()
```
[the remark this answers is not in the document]
LexLuthr: Cleaned up car. Redesigning the stream thing

tasks/seal/task_finalize.go
```diff
@@ -189,39 +184,20 @@ func (f *FinalizeTask) CanAccept(ids []harmonytask.TaskID, engine *harmonytask.T
...
-		SELECT p.task_id_finalize, p.sp_id, p.sector_number, l.storage_id FROM sectors_sdr_pipeline p
-			INNER JOIN sector_location l ON p.sp_id = l.miner_id AND p.sector_number = l.sector_num
-			WHERE task_id_finalize = ANY ($1) AND l.sector_filetype = 4
-`, indIDs)
+	var acceptedID harmonytask.TaskID
+	err := f.db.QueryRow(ctx, `
+		SELECT p.task_id_tree_c FROM sectors_sdr_pipeline p
+			INNER JOIN sector_location l ON p.sp_id = l.miner_id AND p.sector_number = l.sector_num AND l.sector_filetype = 4
+			INNER JOIN storage_path sp ON sp.storage_id = l.storage_id
+			WHERE task_id_tree_r = ANY ($1) AND sp.urls IS NOT NULL AND sp.urls LIKE '%' || $2 || '%' LIMIT 1`, indIDs, engine.Host()).Scan(&acceptedID)
```
snadrus: https://github.com/filecoin-project/curio/pull/854 plan?

PR #856, opened 2025-12-31
fix raw size

LexLuthr:
This PR fixes the missing raw size from market_piece_deal table after Boost migration.


tasks/storage-market/task_fix_rawSize.go
```diff
@@ -0,0 +1,296 @@
...
+		IAmBored: passcall.Every(time.Minute, func(taskFunc harmonytask.AddTaskFunc) error {
+			return f.schedule(context.Background(), taskFunc)
+		}),
+	}
+}
+
+func (f *FixRawSize) schedule(ctx context.Context, taskFunc harmonytask.AddTaskFunc) error {
+	var stop bool
+
+	for !stop {
```
magik6k: I would do one-by-one scheduling here, without the loop; 1000s of tasks sitting in the scheduler are bad

tasks/storage-market/task_fix_rawSize.go
```diff
@@ -0,0 +1,296 @@
...
+}
+
+func (f *FixRawSize) TypeDetails() harmonytask.TaskTypeDetails {
+	return harmonytask.TaskTypeDetails{
+		Max:  taskhelp.Max(16),
+		Name: "FixRawSize",
+		Cost: resources.Resources{
+			Cpu: 1,
+			Gpu: 0,
+			Ram: 64 << 30,
```
magik6k: 64G is really high, I'd say maybe 4G

tasks/storage-market/task_fix_rawSize.go
```diff
@@ -0,0 +1,296 @@
...
+			Number: abi.SectorNumber(sectorNumer),
+		},
+		ProofType: abi.RegisteredSealProof(proof),
+	}, storiface.PaddedByteIndex(pieceOffset).Unpadded(), abi.PaddedPieceSize(pieceSize).Unpadded(), pcid)
+
+	if err != nil {
+		return false, xerrors.Errorf("getting piece reader: %w", err)
+	}
+
+	var carSize uint64
```
magik6k: I'd put everything from here that extracts rawSize from a reader and put it in a separate function and add unit tests covering v1/v2 cases at least somewhat, would be an easy way to have at least some reasonable confidence

tasks/storage-market/task_fix_rawSize.go
magik6k: Or at minimum maybe limit to 100 entries in fix_raw_size at one time

magik6k (commented):
Would be really great to also have a minimal "repairs" tab on the market page which shows ongoing fixes / allows admins easy restart / matching task to deal/sector

harmony/harmonydb/sql/20251231-fix-raw-size.sql
```diff
@@ -0,0 +1,4 @@
+CREATE TABLE IF NOT EXISTS fix_raw_size (
```
magik6k: nit: I'd prefix table name with market_

LexLuthr:
> Can we write a file per storage area indicating that the storage path is clear of zero length sectors so it saves us from a triple join every startup?

Storage is dyamic, users can add remove files any time and storage picks it up. A static file would make it hard to maintain consistent state.

LexLuthr:
> Would be really great to also have a minimal "repairs" tab on the market page which shows ongoing fixes / allows admins easy restart / matching task to deal/sector

I do agree but I am going to defer that to another PR. This needs to be shipped quickly to allow repairs to work fix broken UI and index check tasks.

PR #864, opened 2026-01-09
feat(pdp): add SP-to-SP piece pull endpoint

rvagg:
Implement POST /pdp/piece/fetch for fetching pieces from other SPs. Includes idempotent request handling, failure tracking with retry status, and CommP verification for external sources.

New files:
- handlers_fetch.go: fetch endpoint and database store implementation **See HandleFetch() Godoc for comprehensive details about new functionality.**
- fetch_types.go: request/response types with source URL validation
- piece_cid.go: PieceCIDv2 parsing utilities (consolidated existing PieceCID code)
- 20251212-pdp-fetch.sql: tracking tables for fetch requests

Changes:
- task_park_piece.go: verify CommP for external sources, reduce MaxFailures to 5
- handlers.go: refactor to use shared PieceCID utilities

Pulls in lib/ffi/piece_funcs.go from main branch as is.

### DIFF
--- a/pdp/handlers.go
+++ b/pdp/handlers.go
@@ -18,11 +18,7 @@ import (
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/go-chi/chi/v5"
-	"github.com/ipfs/go-cid"
-	"github.com/multiformats/go-multicodec"
 	"github.com/yugabyte/pgx/v5"
-
-	commcid "github.com/filecoin-project/go-fil-commcid"
 
 	"github.com/filecoin-project/curio/harmony/harmonydb"
 	"github.com/filecoin-project/curio/lib/paths"
@@ -56,6 +52,8 @@ type PDPService struct {
 	sender    *message.SenderETH
 	ethClient *ethclient.Client
 	filClient PDPServiceNodeApi
+
+	pullHandler *PullHandler
 }
 
 type PDPServiceNodeApi interface {
@@ -64,14 +62,20 @@ type PDPServiceNodeApi interface {
 
 // NewPDPService creates a new instance of PDPService with the provided stores
 func NewPDPService(ctx context.Context, db *harmonydb.DB, stor paths.StashStore, ec *ethclient.Client, fc PDPServiceNodeApi, sn *message.SenderETH) *PDPService {
+	auth := &NullAuth{}
+	pullStore := NewDBPullStore(db)
+	pullValidator := NewEthCallValidator(ec, db)
+
 	p := &PDPService{
-		Auth:    &NullAuth{},
+		Auth:    auth,
 		db:      db,
 		storage: stor,
 
 		sender:    sn,
 		ethClient: ec,
 		filClient: fc,
+
+		pullHandler: NewPullHandler(auth, pullStore, pullValidator),
 	}
 
 	go p.cleanup(ctx)
@@ -146,6 +150,9 @@ func Routes(r *chi.Mux, p *PDPService) {
 
 	// POST /pdp/piece/uploads/{uploadUUID}
 	r.Post(path.Join(PDPRoutePath, "/piece/uploads/{uploadUUID}"), p.handleFinalizeStreamingUpload)
+
+	// POST /pdp/piece/pull - Pull pieces from other SPs
+	r.Post(path.Join(PDPRoutePath, "/piece/pull"), p.pullHandler.HandlePull)
 }
 
 // Handler functions
@@ -181,12 +188,12 @@ func (p *PDPService) handleGetPieceStatus(w http.ResponseWriter, r *http.Request
 	}
 
 	// Convert to v1 format (database stores v1)
-	pieceCidV1, err := asPieceCIDv1(pieceCidStr)
+	info, err := ParsePieceCid(pieceCidStr)
 	if err != nil {
 		http.Error(w, "Invalid pieceCid format: "+err.Error(), http.StatusBadRequest)
 		return
 	}
-	pieceCidV1Str := pieceCidV1.String()
+	pieceCidV1Str := info.CidV1.String()
 
 	// Query status from database
 	var result struct {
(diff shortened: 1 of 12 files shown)


tasks/piece/task_park_piece.go
```diff
@@ -229,7 +272,7 @@ func (p *ParkPieceTask) TypeDetails() harmonytask.TaskTypeDetails {
 			Ram:     64 << 20,
 			Storage: p.sc.Storage(p.taskToRef, storiface.FTPiece, storiface.FTNone, maxSizePiece, storageType, ParkMinFreeStoragePercent),
 		},
-		MaxFailures: 10,
+		MaxFailures: 5,
```
LexLuthr: Again, this impacts PoRep as well. It should stay at 10.

LexLuthr (commented):
Looks good. Just need to change the ParkPiece strategy.

snadrus:
@rvagg Can we have the spec for this ticket? 
I would like the Curio team to implement it in a supportable way.

tasks/piece/task_park_piece.go
```diff
@@ -183,12 +190,48 @@ func (p *ParkPieceTask) Do(taskID harmonytask.TaskID, stillOwned func() bool) (d
...
 			}
 
-			if err := p.sc.WritePiece(ctx, &taskID, pnum, pieceData.PieceRawSize, upr, storageType); err != nil {
-				merr = multierror.Append(merr, xerrors.Errorf("write piece: %w", err))
-				continue
+			// Check if this is an internal custore:// source (already verified during upload)
+			// or an external URL (needs CommP verification)
+			isCustore := strings.HasPrefix(refData[i].DataURL, dealdata.CustoreScheme+"://")
+
+			if isCustore {
```
magik6k: Please keep old behavior for mk12/mk20 pipelines.


Separately:

Ideally this retrieval would happen as some sort of "verifiable piece retrieval" which IMHO should be a light FRC, for example:
1. Fetch some middle layer cache of TreeD `/piece/ba...2abc4?tree=leaf32m` -> returns commp stream where each node is 32MiB of data below
2. Start a fetch where we stream-verify the result - this way you don't just fail the entire task when the retrieval inevitably fails in the middle with no serious retry strategy 

You are implementing a retrieval pathway here, and moving a lot of data - this must be one of the most defensive and robust piece of the codebase with no shortcuts, otherwise you're signing up for months of suffering.

(For context filecoin-gateway v1 has multiple layers to recover from bad providers, up to being able to recover gracefully without losing progress even if .car stream contains bitflips in any part of the stream including varints - https://github.com/CIDgravity/gw/blob/main/ributil/repair_car.go)

It is 100% guaranteed that here you will see:
* Bitflips
* Slow providers
* Multiple slow providers
* Providers who just disappeared suddenly and the other ones have issues above

lib/ffi/piece_funcs.go
```diff
@@ -83,3 +87,82 @@ func (sb *SealCalls) PieceReader(ctx context.Context, id storiface.PieceNumber)
 func (sb *SealCalls) RemovePiece(ctx context.Context, id storiface.PieceNumber) error {
 	return sb.Sectors.storage.Remove(ctx, id.Ref().ID, storiface.FTPiece, true, nil)
 }
+
+func (sb *SealCalls) WriteUploadPiece(ctx context.Context, pieceID storiface.PieceNumber, size int64, data io.Reader, storageType storiface.PathType, verifySize bool) (abi.PieceInfo, uint64, error) {
```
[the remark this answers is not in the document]
magik6k: If we need to add commp verification either it should be immediate or it should be part of the main WritePiece which can be extended to support proper incremental verification (likely coupled to a small FRC describing how that is supposed to work). Just copy-pasting 90% of the code is honestly a bit sloppy

pdp/handlers_fetch.go
```diff
@@ -0,0 +1,798 @@
...
+	commcid "github.com/filecoin-project/go-fil-commcid"
+
+	"github.com/filecoin-project/curio/harmony/harmonydb"
+	"github.com/filecoin-project/curio/pdp/contract"
+)
+
+// FetchRecord represents a fetch request record from the database
+type FetchRecord struct {
+	ID            int64
+	Service       string
```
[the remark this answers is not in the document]
rvagg: Effectively we do but the code still has the pathways for alternatives to `"public"` so I'm just working with that. We could treat this API as assumed-`"public"` if we want?

rvagg:
@LexLuthr in a new `fixup!` commit I've reworked this based on your feedback. Restored task_park_piece.go and lib/ffi/piece_funcs.go to their original state, so no changes to `ParkPiece` or the shared piece funcs.

Created a separate `PDPv0_FetchPiece` in tasks/pdp/task_fetch_piece.go that handles the fetch flow independently: downloads from source URL, computes and verifies CommP while streaming, stores to custore, then creates a `parked_pieces` entry for `StorePiece` to pick up. Tracking happens via `pdp_piece_fetch_items` which ties into the fetch request lifecycle, cleanup still happens via the existing CASCADE delete when fetch records age out.

The task runs up to 4 concurrent downloads, each isolated so a slow/bad source doesn't block others. Has a 1 hour total fetch timeout (these are capped at 1016 MiB total) and exponential backoff on retries (5 max). Still room for verified streaming retrieval in future but I hope this should be solid enough for a starter?

rjan90:
> @rjan90 we're going to get this landed before you cut a tag this week, keeping it open while I'm still testing on it in case something comes up, but it's ready as is, works, tested, just not heavily exercised.

Discussed in sync today to merge this, so that we can cut a tag and get it shipped out to PDP SPs.

pdp/handlers.go
```diff
@@ -1155,6 +1120,13 @@ func (p *PDPService) cleanup(ctx context.Context) {
 				log.Errorw("failed to delete non-finalized uploads", "error", err)
 			}
 		}
+
+		// Clean up old piece fetch records (older than 5 days)
+		// CASCADE deletes pdp_piece_fetch_items automatically
+		_, err = db.Exec(ctx, `DELETE FROM pdp_piece_fetches WHERE created_at < NOW() - INTERVAL '5 days'`)
```
beck-8: @rvagg  Can you look over here? There may be some errors here.

```
  {"level":"error","ts":"2026-02-05T15:14:47.655+0800","logger":"pdp","caller":"pdp/handlers.go:1128","msg":"failed to delete old piece fetch records","error":"ERROR: relation \"pdp_piece_fetches\" does not exist (SQLSTATE 42P01)"}
```

PR #869, opened 2026-01-10
fix(pdp): detect contract reverts and exclude terminated datasets from scheduling

rvagg:
Stop infinite retry loop when datasets are terminated on-chain. Detects termination errors at gas estimation, marks datasets terminated, and excludes them from scheduling. Unknown contract reverts get exponential backoff.

## Reverting

If an SP wants to try running this code and decides to back out, there is a database migration in here. A revert is possible but you should run:

```sql
  ALTER TABLE pdp_data_sets DROP COLUMN IF EXISTS terminated_at_epoch;
  ALTER TABLE pdp_data_sets DROP COLUMN IF EXISTS consecutive_prove_failures;
  ALTER TABLE pdp_data_sets DROP COLUMN IF EXISTS next_prove_attempt_at;
```

The old code ignores the new columns, so in theory it will run fine without a DB migration rollback. But you probably should clean up your db schema to be on the safe side for future updates.

### DIFF
--- a/tasks/pdp/failure_handling.go
+++ b/tasks/pdp/failure_handling.go
@@ -1,0 +1,156 @@
+package pdp
+
+import (
+	"context"
+
+	"golang.org/x/xerrors"
+
+	"github.com/filecoin-project/curio/harmony/harmonydb"
+)
+
+const (
+	// MaxConsecutiveFailures is the threshold for giving up on a dataset.
+	// Used by ApplyProvingBackoff: after this many consecutive contract reverts
+	// without a successful prove, the dataset is marked as terminated.
+	// This gives time for external resolution (e.g., client adds funds).
+	MaxConsecutiveFailures = 5
+
+	// BaseBackoffBlocks is the initial delay after the first contract revert.
+	// Used by CalculateBackoffBlocks: subsequent failures double this value.
+	BaseBackoffBlocks = 100
+
+	// MaxBackoffBlocks prevents unbounded exponential growth.
+	// Used by CalculateBackoffBlocks to cap the delay. In practice,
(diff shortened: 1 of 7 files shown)


rvagg:
> Too FWSS specific

The FWSS specific bits are kept ~generic - we're just matching known terminal reverts but it's not exhaustive. Only if we _know_ its terminal do we proceed to clean up, otherwise we go into a retry cycle and eventually terminate. So it's not locked to FWSS, it'll still clean up regardless, it just has a shortcut.

tasks/pdp/failure_handling.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	return backoff
+}
+
+// MarkDatasetTerminated marks a dataset as terminated, stopping all future proving attempts.
+// This is called when a termination error (like DataSetPaymentBeyondEndEpoch) is detected.
+func MarkDatasetTerminated(ctx context.Context, db *harmonydb.DB, dataSetId int64, currentHeight int64) error {
+	_, err := db.Exec(ctx, `
+		UPDATE pdp_data_sets
+		SET terminated_at_epoch = $2,
```
ZenGround0: One thing I want to fix is the problematic naming collision between this and service termination (which #871 is all about hooking this function up to)

_Service termination_ today begins when payer is in default.  It involves terminating rails and entering the grace period but NOT halting proving operations until the moment we delete the dataset in order to not be penalized on the final period of fee collection by not submitting proofs. 

The concept here which I'm calling UnrecoverableProvingFailure should terminate service as well but has the additional property that we are fully done with all proving too.

It's possible we'll want to update datasets to unrecoverableProvingFailure after _dataset deletion_ but I think we'll probably just clean up all of the datasets state at that point anyway.

PR #875, opened 2026-01-12
PDPv0 DB efficiency

snadrus:
These indexes will fix the hot DB queries seen in a trace done during backlog handling. The intent is to see an 80% reduction in CPU for equivalent DB work completed.


snadrus:
@TippyFlitsUK This is the one from your scan. You're welcome to try out this branch to see if it's helpful, but the SQL downgrade flow isn't ported to pdpv0 yet, so removal would require running some SQL commands to `DELETE INDEX abc...`

rvagg:
@TippyFlitsUK no these are in addition; I'll merge this in.

beck-8:
I tested it and saw that the database load was reduced from 100% to 80%, and the task processing speed seemed to be improved.

PR #884, opened 2026-01-15
Fix UPDATE orphanage

ZenGround0:
We haven't been checking the output rows effected by tx.Exec UPDATE and apparently this doesn't actually fail during a race but succeeds and returns n==0.  This means that we have been allowing tasks with no matching pdp piece ref to get scheduled.  In the case of indexing this closely matches the error behavior we are seeing on ezpdpz.  Unclear if this is the actual cause but it is a possible match we should try.


snadrus (approved):

rvagg:
Should be duplicated into task_indexing.go too I think so we don't end up losing this when we transition back. task_ipni.go is different and might already have this.

Issue #880, opened 2026-01-14
pdpv0: Task backlog will affect normal data uploading


beck-8:
I'm using a 32G 28 thread wsl2 instance. use pdpv0(Contains two fix PRs)

After confirming that there are no tasks and no load, perform a stress test.
```
node dist/index.js parallel --contexts 10 --uploads 50 --rounds 10 --delay 5000 --interval 100 --provider 6
```

The above tasks have not been completed. 0/5000, the success rate is 0, and it ends quickly with a lot of errors along the way.

**The test ends at 18:30. Not restored yet (20:45).**

process
1. End of test script
2. It is observed that a large number of storepiece tasks are slowly increasing and then slowly decreasing.
3. Wait until the storepiece ends and PDPv0_Notify slowly increases.
4. My instance is stuck (in this case, the system memory is almost 0)
5. Restart the instance, restart lotus, database, curio
6. curio OOM (it may also be due to increased load that other components occupy more memory)
7. Adjust machine memory to 50G+
8. After curio was started, it was found that it occupied about 30G. After a while, curio only occupied 300M
9. PDPv0_Notify is still being processed and data cannot be uploaded.

<img width="79" height="91" alt="Image" src="https://github.com/user-attachments/assets/bb00819f-7e33-4aff-9cbd-863286efc816" />

beck-8:
OOM should be resolved.
https://github.com/filecoin-project/curio/pull/883#issuecomment-3758139424

> Indexing improvement 

I tested it and saw that the database load was reduced from 100% to 80%, and the task processing speed seemed to be improved.

beck-8:
> if we get bit by this more or there is a conviction that this is too big of a risk to take for GA

I think this must be done before. Reason: It is convincing enough that an attacker can paralyze all nodes in the entire network by spending only a small amount of money and resources.

beck-8:
Record current test results
In the case of 100 contexts/datasets at the same time, a 1KiB file is uploaded 5 times in each round. Except for a few failures, everything was successful. There is no backlog of tasks.
`StorageContext addPieces failed: Failed to add piece to data set - Request timed out after 300000ms`

4d53c8017ad345410adfd80794fd7518b49c9128

```
node dist/index.js parallel --contexts 100 --uploads 1 --rounds 5 --delay 1000 --interval 100 --provider 6
```
```
Context 90:
  Provider ID: 6
  Dataset ID: 9228 (existing: false)
  Uploads: 5/5 successful
[...]

PR #900, opened 2026-01-21
fix: resolve goroutine leak in CommP calculation

beck-8:
https://github.com/filecoin-project/curio/issues/880#issuecomment-3777759004

The underlying library `github.com/filecoin-project/go-fil-commp-hashhash` spawns background goroutines (addLayer) to handle parallel tree construction. These goroutines are designed to exit only when:

1. The calculation finishes successfully (Digest completes and closes the channel).
2. Reset() is explicitly called.

However, if Digest() encounters an error (e.g., input data is smaller than MinPiecePayload (65 bytes), or io.Copy fails), it returns immediately without closing the internal channels. This leaves the background addLayer goroutines blocked indefinitely on channel reads, causing a silent accumulation of goroutines (observed ~500+ leaked goroutines in production during idle state).

lib/proof/datacid.go There also be this problem, fix it by the way

### DIFF
--- a/cmd/pdptool/main.go
+++ b/cmd/pdptool/main.go
@@ -297,6 +297,7 @@ func loadPrivateKey() (*ecdsa.PrivateKey, error) {
 func preparePiece(r io.ReadSeeker) (cid.Cid, uint64, []byte, []byte, error) {
 	// Create commp calculator
 	cp := &commp.Calc{}
+	defer cp.Reset()
 
 	// Copy data into commp calculator
 	rawSize, err := io.Copy(cp, r)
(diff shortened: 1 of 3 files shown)


rvagg:
@beck-8 great fine, but you're missing one location: https://github.com/filecoin-project/curio/blob/4ccd24f251ca742a9f5f8c63d6af8f153b8ae240/pdp/handlers_upload.go#L219

There's also a couple in cmd/pdptool/main.go that are lower priority but we should do it for completeness (for anyone copying code in particular).

main branch also has some, including a mk20 one that we'll need to make sure get fixed too, maybe you could open a separate main branch PR that does this stuff and the rest?

Issue #918, opened 2026-01-24
pdpv0: Obsolete PDPv0_Prove tasks should be discarded


beck-8:
This is currently the case
1. After restarting/shutdown for a while, a large number of PDPv0_Prove tasks will be generated
2. A large number of tasks will consume all the memory during scheduling, causing other tasks to fail to proceed normally.
3. The accumulated tasks cannot be completed normally. New tasks will continue to be blocked due to queue reasons.

```
error: failed to submit possesion proof for dataset 9699: failed to get chain randomness from beacon for pdp prove: failed to get randomness digest from tickets: cannot draw randomness from the future
```

<img width="2135" height="950" alt="Image" src="https://github.com/user-attachments/assets/08b359f0-2628-4116-a717-4b305696cafa" />

---
> Of course, this is also related to the short certification interval of the test network.

- Is there a way for us to discard tasks in advance that we shouldn’t continue doing?
- Do we want to limit the number of parallel PDPv0_Prove tasks?

beck-8:
@rvagg It looks very good, and I have a backlog of several thousand tasks that I have finished in 15 minutes.

PR #934, opened 2026-01-28
pdp: reset to next proving period when challenge epoch is in the future

beck-8:
https://github.com/filecoin-project/curio/issues/918

### DIFF
--- a/tasks/pdp/task_prove.go
+++ b/tasks/pdp/task_prove.go
@@ -222,9 +222,9 @@ func (p *ProveTask) Do(taskID harmonytask.TaskID, stillOwned func() bool) (done 
 	var proveAtEpoch *int64
 	var challengeWindow *int64
 	err = p.db.QueryRow(ctx, `
-	SELECT prove_at_epoch, challenge_window
-	FROM pdp_data_sets
-	WHERE id = $1
+		SELECT prove_at_epoch, challenge_window
+		FROM pdp_data_sets
+		WHERE id = $1
 	`, dataSetId).Scan(&proveAtEpoch, &challengeWindow)
 	if err != nil {
 		return false, xerrors.Errorf("failed to check task timeliness: %w", err)


rvagg (changes requested):
eh I shouldn't have approved with big suggestions

ZenGround0 (changes requested):

tasks/pdp/failure_handling.go
```diff
@@ -119,6 +119,51 @@ func ResetProvingFailures(ctx context.Context, db *harmonydb.DB, dataSetId int64
...
+// NextPP calls NextPDPChallengeWindowStart on the contract, which advances the
+// deadline in whole-period increments from the original cadence rather than
+// re-initializing.
+// If NextPP discovers the on-chain state is actually uninitialized, it will call
+// ResetDatasetToInitPP itself.
+func ResetDatasetToNextPP(ctx context.Context, db *harmonydb.DB, dataSetId int64) error {
+     log.Infow("resetting dataset to next proving period state", "dataSetId", dataSetId)
+     _, err := db.Exec(ctx, `
+             UPDATE pdp_data_sets
+             SET challenge_request_task_id = NULL,
```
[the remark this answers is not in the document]
rvagg: You could put the task id in the query, plumb it through into the function from the caller: `WHERE id = $1 AND challenge_request_task_id = $2`, that would make this a bit safer? A next_pp might have claimed this task and that would make the update a noop.

tasks/pdp/failure_handling.go
```diff
@@ -119,6 +119,28 @@ func ResetProvingFailures(ctx context.Context, db *harmonydb.DB, dataSetId int64
 	return err
 }
 
+// ResetDatasetToInitPP resets a dataset so that InitProvingPeriodTask picks it up.
+// This is only appropriate for datasets whose on-chain proving period was never
+// initialized (e.g. ProvingPeriodNotInitialized error from the contract). InitPP
+// computes a fresh challenge window from config.InitChallengeWindowStart, which is
+// only valid for first-time initialization. For datasets that have an established
+// proving cadence but missed a window, use ResetDatasetToNextPP instead.
+func ResetDatasetToInitPP(ctx context.Context, db *harmonydb.DB, dataSetId int64) error {
```
rvagg: may as well put this back to `kickToInit` in tasks/pdp/task_next_pp.go since we don't need it outside of there now

tasks/pdp/task_next_pp.go
```diff
@@ -156,8 +163,8 @@ func (n *NextProvingPeriodTask) Do(taskID harmonytask.TaskID, stillOwned func()
 		// not my favourite way to handle this but pragmatic
 		// for some reason we are in a proving loop running but it is not initialized
 		if strings.Contains(err.Error(), "0x999010d5") { // Error.ProvingPeriodNotInitialized
-			if err := n.kickToInit(ctx, dataSetId); err != nil {
-				return false, xerrors.Errorf("failed to kick back to init: %w", err)
+			if err := ResetDatasetToInitPP(ctx, n.db, dataSetId); err != nil {
```
rvagg:
```suggestion
			if err := resetDatasetToInitPP(ctx, n.db, dataSetId); err != nil {
```

tasks/pdp/task_prove.go
```diff
@@ -208,6 +208,52 @@ func (p *ProveTask) Do(taskID harmonytask.TaskID, stillOwned func() bool) (done
...
+				"proveAtEpoch", *proveAtEpoch,
+				"challengeWindow", *challengeWindow,
+				"currentEpoch", int64(ts.Height()),
+			)
+			return true, nil
+		}
+		// Missed challenge window and next_proving_period already reset.
+		// Noop and wait for prove task scheduling.
+		if *proveAtEpoch > int64(ts.Height()) {
+			log.Errorw("Prove task awoke too late, fault registered, proving period rescheduled",
```
rvagg:
```suggestion
			log.Errorw("Prove task missed challenge window, period will be faulted, next period already scheduled (proving may be backlogged)",
```

PR #937, opened 2026-01-29
Settle watcher inefficient query 

### LINKED ISSUE #923: pdpv0:  slow SQL: SettleWatcher scans 650k+ rows every tipset even when no pending settlements
## Environment                                                                                                                                               
                                                                                                                                                               
  - **Date:** Jan 25, 2026                                                                                                                                     
  - **Version:** curio version 1.27.0+calibnet+git_0e56f82e_2026-01-22T18:51:12-05:00                                                                           
  - **Node:** calib2.ezpdpz.net (ID 2)                                                                                                                           
                                                                                                                                                               
  ## Description                                                                                                                                               
                                                                                                                                                               
  The `SettleWatcher` runs on every tipset (~30 seconds) and executes two queries that JOIN `filecoin_payment_transactions` with `message_waits_eth`. These    
  queries perform a sequential scan of 650k+ rows in `message_waits_eth` even when `filecoin_payment_transactions` is empty.                                   
                                                                                                                                                               
  ## Log Output                                                                                                                                                
                                                                                                                                                               
```
  Jan 25 18:55:36.223 |WARN| Slow SQL run rowCt=0 err=null logger=harmonydb milliseconds=5125 caller=harmonydb/harmonydb.go:163 query=                         
          SELECT fpt.tx_hash, fpt.rail_ids                                                                                                                     
          FROM filecoin_payment_transactions fpt                                                                                                               
          JOIN message_waits_eth mwe ON fpt.tx_hash = mwe.signed_tx_hash                                                                                       
          WHERE mwe.tx_status = 'confirmed' AND mwe.tx_success = TRUE       
```                                                                                   
                                                                                                                                                               
  ## Query Plan Analysis                                                                                                                                       
                                                                                                                                                               
  YB Batched Nested Loop Join  (cost=0.00..226.50 rows=1000 width=64) (actual time=5331.362..5331.366 rows=0 loops=1)                                          
     Join Filter: (fpt.tx_hash = mwe.signed_tx_hash)                                                                                                           
     ->  Seq Scan on message_waits_eth mwe  (actual time=4.525..268.883 rows=652887 loops=1)                                                                   
           Storage Filter: (tx_success AND (tx_status = 'confirmed'::text))                                                                                    
     ->  Index Scan using filecoin_payment_transactions_pkey on filecoin_payment_transactions fpt                                                              
           Index Cond: (tx_hash = ANY (ARRAY[mwe.signed_tx_hash, ...]))                                                                                        
   Execution Time: 5332.159 ms                                                                                                                                 
                                                                                                                                                               
  ## Root Cause                                                                                                                                                
                                                                                                                                                               
  1. `NewSettleWatcher` (watcher.go:29) registers a handler on `CurioChainSched` that runs every tipset                                                        
  2. `processPendingTransactions` executes two queries (lines 44-48 and 66-70) that JOIN with `message_waits_eth`                                              
  3. No index exists on `message_waits_eth` for `tx_status = 'confirmed'` - existing indexes only cover `pending` status                                       
  4. YugabyteDB scans all 652k confirmed rows in `message_waits_eth` before checking the (empty) `filecoin_payment_transactions` table                         
                                                                                                                                                               
  ## Impact                                                                                                                                                    
                                                                                                                                                               
  - 5+ second query executed every ~30 seconds                                                                                                                 
  - Unnecessary database load when no payment settlements are pending                                                                                          
  - Contributes to overall system latency                                                                                                                      
                                                                                                                                                               
  ## Affected Code                                                                                                                                             
                                                                                                                                                               
  `tasks/pay/watcher.go:41-93` - `processPendingTransactions()`                                                                                                
                                                                                                                                                               
  ## Existing Indexes on message_waits_eth                                                                                                                     
                                                                                                                                                               
  idx_message_waits_eth_pending              WHERE tx_status = 'pending'                                                                                       
  idx_message_waits_eth_waiter_pending       WHERE tx_status = 'pending'                                                                                       
                                                                                                                                                               
  Note: No index for `confirmed` status.

ZenGround0:
Closes #923 by trying to force the query optimizer to use fpt as driving table.

### DIFF
--- a/tasks/pay/watcher.go
+++ b/tasks/pay/watcher.go
@@ -55,7 +55,7 @@ func processPendingTransactions(ctx context.Context, db *harmonydb.DB, ethClient
 	SELECT fpt.tx_hash, fpt.rail_ids
 	FROM filecoin_payment_transactions fpt`)
 	if err != nil {
-		return xerrors.Errorf("failed to get failed settlements from DB: %w", err)
+		return xerrors.Errorf("failed to get settlements from DB: %w", err)
 	}
 	if len(settles) == 0 {
 		return nil
@@ -93,6 +93,8 @@ func processPendingTransactions(ctx context.Context, db *harmonydb.DB, ethClient
 				return xerrors.Errorf("failed to delete failed settlement from DB: %w", err)
 			}
 		}
+
+		// since mwe updated atomically with fpt iterating all mwe should iterate all settles
 	}
 
 	serviceAddr := contract.ContractAddresses().AllowedPublicRecordKeepers.FWSService


ZenGround0:
For what its worth I get a good EXPLAIN ANALYZE on the new bottleneck query on my local 100k mwe table: 

```
| QUERY PLAN                                                                          >
|------------------------------------------------------------------------------------->
| Index Scan using message_waits_eth_pkey on message_waits_eth  (cost=100.00..104.12 r>
|   Index Cond: (signed_tx_hash = ANY ($0))                                           >
|   InitPlan 1 (returns $0)                                                           >
|     ->  Seq Scan on new_table  (cost=0.00..100.00 rows=1000 width=32) (actual time=0>
| Planning Time: 0.054 ms                                                             >
| Execution Time: 0.693 ms                                                            >
| Peak Memory Usage: 152 kB 
```

This doesn't mean much since I similarly get an index seq on the JOIN that I originally tried

ZenGround0:
I agree this is a problem.  I'm looking to see if there is any obvious bound on pending txs in the eth message waits watcher that helps us here.

ZenGround0:
Yeah as you'd expect there's nothing I can see preventing massive hangs.  The problem of infinitely stuck pending transactions should be solved at the level of the message_waits_eth subsystem.  One proposal would be to include a new status "dropped" alongside "pending" and "confirmed" that we periodically assign to transactions that have been hanging for a long time.  Then the settlement watcher task can handle submitting retries for these hanging messages.

I would hold off on doing this until we observe or otherwise suspect the frequent existance of large scale mpool hanging as a significant issue.  The fact that the PoRep `message_waits` table doesn't deal with the stuck-in-mpool case more directly makes me suspicious that lower down in the message sending system we have strategies to avoid this that generally work.  But since we are tracking by message cids across the board I suspect that we don't fully have premium / fee cap increase logic yet.

PR #938, opened 2026-01-29
always fast make-gen

snadrus:
From 5-ish minutes to under 30 seconds.
We've all got to run it often (more and more all the time), so lets make it snappy. 

This isn't a "fastpath" but literally makes the "slow path" complete faster, mostly: 
- fix-imports does only run on files that were changed since the branch-from-main. 
- gotext extraction is more efficient (replaced). application is still slow, but rare


LexLuthr (approved):
Seems like slowest part is now docsgen-cli. We should rethink that as well. Current iteration is a Lotus copy. I think our toy main trick can be reused there as well. That would bring our make-gen to probably under 3 minutes. This is still the most time taking job.

PR #943, opened 2026-02-03
fix(pdp): register pulled pieces in pdp_piecerefs for service ownership

rvagg:
This is a follow-up to #864:

Pieces fetched via SP-to-SP pull were not being registered in pdp_piecerefs, causing "subPiece CID not found or does not belong to service" errors when attempting to add them to datasets. The pull task now creates the pdp_piecerefs entry (paralleling notify_task.go for uploads) after storing the piece.


rvagg:
fwiw my e2e testing was mostly limited to parking the piece and confirming its availability; not the next bit - actually using it to confirm on chain .. ooops

PR #945, opened 2026-02-05
fix(pay): handle nullable tx_success in settle watcher

rvagg:
message_waits_eth.tx_success is NULL for pending transactions. Use *bool to allow NULL values and safely dereference after the status check.

Fixes: https://github.com/filecoin-project/curio/pull/937#issuecomment-3851229263


ZenGround0 (approved):
thx!

PR #946, opened 2026-02-05
pdpv0: Gracefully terminate pdp proving

ZenGround0:
This is a follow on from #934.

#934 handles the much more reachable case of a proving task getting woken up too late and needing to gracefully exit the system.  

This PR is hoping to further reduce the less likely but more nefarious state of nextProvingPeriod task getting shut down midway and leaving blockchain and db in bad state which leads to unschedulable prove task for that period.  To do this we add `PDPv0_ProvPeriod` to the graceful shutdown list.  

I'm also adding `PDPv0_Prove` to the list because it will help SPs avoid missing proofs.  But this is a slighly lower concern than gracefully shutting down `PDPv0_ProvPeriod`.

### DIFF
--- a/harmony/harmonytask/harmonytask.go
+++ b/harmony/harmonytask/harmonytask.go
@@ -280,7 +280,7 @@ func (e *TaskEngine) GracefullyTerminate() {
 
 			if h.Name == "PDPv0_ProvPeriod" && h.Max.Active() > 0 {
 				timeout = time.Second
-				log.Infof("node shutdown deferred for %f seconds due to running PDPNextProvingPeriod task", timeout.Seconds())
+				log.Infof("node shutdown deferred for %f seconds due to running PDPv0_ProvPeriod task", timeout.Seconds())
 				continue
 			}
 


harmony/harmonytask/harmonytask.go
```diff
@@ -271,6 +271,19 @@ func (e *TaskEngine) GracefullyTerminate() {
...
+			// PDP tasks
+			if h.Name == "PDPv0_Prove" && h.Max.Active() > 0 {
+				timeout = time.Second
+				log.Infof("node shutdown deferred for %f seconds due to running PDPProve task", timeout.Seconds())
+				continue
+			}
+
+			if h.Name == "PDPv0_ProvPeriod" && h.Max.Active() > 0 {
+				timeout = time.Second
+				log.Infof("node shutdown deferred for %f seconds due to running PDPNextProvingPeriod task", timeout.Seconds())
```
rvagg:
```suggestion
				log.Infof("node shutdown deferred for %f seconds due to running PDPv0_ProvPeriod task", timeout.Seconds())
```

rvagg (approved):
looks like a good time to replace the contents of this loop with something more scalable

PR #947, opened 2026-02-05
Check for fully removed pieces if not found in deletion scheduling queue

ZenGround0:
Addressing #924 

I suspect that this will fix the problem but I could use your help seeing if this works @TippyFlitsUK

PR #964, opened 2026-02-11
fix: fallback on PieceLive in cases of no scheduled removals on-chain

redpanda-f:
`PieceDeleteWatcher` complains of https://github.com/filecoin-project/curio/blob/4be3802825fa3df858f83de2e324e41684ef1ec7/tasks/pdp/watch_piece_delete.go#L39.

Theory: 
- User-initiated deletion sets `rm_message_hash` for a piece after the delete tx is sent
- `processPendingPieceDeletes` selects `WHERE psp.rm_message_hash IS NOT NULL` (where rm_message_hash is something).
- Attempts to call PDP Verifier for `GetScheduledRemovals` and check if removal is contained there
    - Assumption here is that all pending deletes for curio local node view is also Scheduled removal on-chain, ergo, on-chain schedules are APPEND-ONLY
    - This is not the case: see https://github.com/FilOzone/filecoin-services/blob/cbf46e05062e53f0c7bb1b286a84852273736d62/service_contracts/src/FilecoinWarmStorageService.sol#L928C27-L928C36 for on-chain schedule addition
    - This removes it: https://github.com/FilOzone/filecoin-services/blob/cbf46e05062e53f0c7bb1b286a84852273736d62/service_contracts/src/FilecoinWarmStorageService.sol#L1319
- Attempted solution: rely on https://github.com/FilOzone/pdp/blob/12a37996ffcd3f9e92f42b64c843defa759fb4ae/src/PDPVerifier.sol#L241-L243 for accurate current live-ness data
- Kindly try finding "piece not live but also not scheduled for removal..." (Warning) to validate this works

### DIFF
--- a/tasks/pdp/watch_piece_delete.go
+++ b/tasks/pdp/watch_piece_delete.go
@@ -90,7 +90,7 @@ func processPendingPieceDeletes(ctx context.Context, db *harmonydb.DB, ethClient
 
 	for _, piece := range pendingDeletes {
 		if !piece.TxSuccess.Valid {
-			log.Debugf("for piece ($d:$d) tx %s not found in message_waits_eth", piece.DataSetID, piece.PieceID, piece.TxHash)
+			log.Debugf("for piece (%d:%d) tx %s not found in message_waits_eth", piece.DataSetID, piece.PieceID, piece.TxHash)
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


redpanda-f:
This looks like a copy of https://github.com/filecoin-project/curio/pull/947 :P

PR #971, opened 2026-02-11
feat: Remote Seal

magik6k:
Verified to work mostly


snadrus:
@magik6k this seems to need a test correction for its own test. can you look at that?

PR #1002, opened 2026-02-16
fix(gc): skip deleted miner actors in StorageGCMark

Reiers:
## Problem

When a miner actor is removed from the chain (e.g. a test/calibration miner that was killed), but its sector files still exist in storage paths, `StorageGCMark` calls `StateGetActor` and gets `"actor not found"`. This is treated as a fatal error, causing the task to fail and retry indefinitely (100% failure rate, every 9 minutes).

## Root Cause

`StorageGCMark.Do()` iterates all sectors in storage paths, loads miner actor state for each unique miner ID, then checks precommits/live/unproven sectors to decide what to GC. Two `StateGetActor` call sites had no handling for deleted actors:

1. **Stage 1** (line ~112): Loading miner state at chain head for liveness checks
2. **Stage 3** (line ~250): Loading miner state at finality height for snap sector-key cleanup

## Fix

Handle `"actor not found"` specifically at both sites:

- **Stage 1**: Skip loading miner state. Sectors stay in `toRemove` — they are orphaned since the miner has no on-chain precommits, live, or unproven sectors to subtract.
- **Stage 3**: Skip finality-tipset lookup. Snap sector-key cleanup is irrelevant for non-existent miners.

### Safety: Edge Cases Considered

| Scenario | Actor on-chain? | StateGetActor result | GC behavior |
|---|---|---|---|
| Miner removed from config but alive on-chain | ✅ | Success | **Protected** — normal live/unproven subtraction |
| Miner truly deleted / never existed | ❌ | `actor not found` | **GC orphaned files** ✅ |
| RPC timeout / network glitch | ✅ | Other error | **Task fails & retries** — no accidental GC |
| Mid-migration, miner healthy elsewhere | ✅ | Success | **Protected** — actor still exists |
| Miner faulted / missed deadlines | ✅ | Success | **Protected** — actor still exists |

Only the specific `"actor not found"` string triggers the skip. Any other `StateGetActor` error (transient RPC failures, timeouts) still fails the task, preventing accidental GC of sectors for healthy miners.

Uses string matching (`strings.Contains`) consistent with existing callers in `cmd/sptool/toolbox_deal_client.go`, since the typed `api.ErrActorNotFound` may not survive the JSON-RPC round-trip.

### DIFF
--- a/tasks/gc/storage_gc_mark.go
+++ b/tasks/gc/storage_gc_mark.go
@@ -61,6 +61,16 @@ func NewStorageGCMark(si paths.SectorIndex, remote *paths.Remote, db *harmonydb.
 func (s *StorageGCMark) Do(taskID harmonytask.TaskID, stillOwned func() bool) (done bool, err error) {
 	ctx := context.Background()
 
+	// Load configured miners from all harmony_config layers. Used to guard
+	// against marking sectors for miners that appear "not found" due to
+	// misconfiguration (wrong network build, missed upgrade) rather than
+	// actual deletion.
+	cfgMiners, err := configuredMiners(ctx, s.db)
+	if err != nil {
+		log.Warnw("failed to load configured miners for GC safety check, proceeding without", "error", err)
+		cfgMiners = make(map[address.Address]bool)
+	}
+
 	/*
 		CREATE TABLE storage_removal_marks (
 		    sp_id BIGINT NOT NULL,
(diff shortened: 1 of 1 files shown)


Reiers:
Before:

Skjermbilde 2026-02-16 kl  18 29 53


After:

Skjermbilde 2026-02-16 kl  19 48 22

LexLuthr:
There is a case when user might incorrect build for incorrect network or forgot to upgrade on NW upgrades. We will see similar errors there as well. How do we safeguard against that?

PR #1004, opened 2026-02-16
no more optimal (no fvm) build

snadrus:
Get rid of FFI_NO_FVM=1 and the code that triggered it CURIO_OPTIMAL_LIBFILCRYPTO as it blocks GPU detection by returning an empty vector. The resulting ffiselect reader does not provide sufficient gpu slots to the scheduler.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ setup-cgo-env:
 	fi
 
 build/.filecoin-install: $(FFI_PATH)
-		$(MAKE) curio-libfilecoin; \
+	$(MAKE) curio-libfilecoin
 	@touch $@
 
 MODULES+=$(FFI_PATH)


Makefile
```diff
@@ -28,11 +27,7 @@ setup-cgo-env:
 	fi
 
 build/.filecoin-install: $(FFI_PATH)
-	@if [ "$(CURIO_OPTIMAL_LIBFILCRYPTO)" = "1" ]; then \
 		$(MAKE) curio-libfilecoin; \
```
Reiers: This line has a trailing backslash (leftover from the old `if/else/fi` block) that shell-continues into the `@touch` below. Bash sees `@touch build/.filecoin-install` → `command not found` → Error 127 on every CI job.

```suggestion
	$(MAKE) curio-libfilecoin
```

PR #1006, opened 2026-02-17
fix(pdp): resolve FWSS StateView address via viewContractAddress()

rvagg:
Ref: https://filecoinproject.slack.com/archives/C09MJCQGLET/p1771256599302979

This probably worked when originally implemented, we just didn't catch it when we added the view contract indirection.

Actual fixes are in the tasks, but report was specifically on tasks/pdp/task_terminate_fwss.go: `error: failed to get data set 9966: message execution failed (exit=[33], revert reason=[message failed with backtrace: 00: f0176641 (method 3844450837) -- contract reverted at 75 (33) 01: f0176641 (method 6) -- contract reverted at 513 (33) (RetCode=33)], vm error=[none])`


ZenGround0 (approved):
Great refactor and great find

PR #1008, opened 2026-02-17
chore(deps): update IPLD & IPFS dependencies

rvagg:
I really just wanted the IPLD dependencies updated here but that pulls in boxo latest which then pulls in all the other cruft cause they're all in lockstep. The biggest jump in here is github.com/libp2p/go-libp2p from 0.44 to 0.47, but 0.44 was the trickiest one and that's already done. :crossed_fingers:


LexLuthr (approved):
That's the cleanest deps PR I have ever seen. They usually come with a bunch of file changes. 😆

PR #1013, opened 2026-02-17
update pdp contract bindings

LexLuthr:
This PR reduces the diff with pdpv0 branch. It brings in new contract bindings and other small changes.

### DIFF
--- a/cmd/curio/internal/translations/catalog.go
+++ b/cmd/curio/internal/translations/catalog.go
@@ -191,249 +191,237 @@ var messageKeyToIndex = map[string]int{
 	"List log systems":                                                                          148,
 	"List pipeline events":                                                                      149,
 	"Listen address for HTTP server":                                                            150,
-	"Location of the service provider":                                                          151,
-	"Lotus-Miner to Curio Migration.":                                                           152,
-	"Make sure to send FIL/tFIL to your 0x wallet address for PDP operations.":                  153,
-	"Manage batch sealing operations":                                                           154,
-	"Manage logging":                                                                            155,
-	"Manage node config by layers. The layer 'base' will always be applied at Curio start-up.": 156,
-	"Manage storage paths for this server.":                                                    157,
-	"Manage the sealing pipeline":                                                              158,
-	"Manage unsealed data":                                                                     159,
-	"Math Utils":                                                                               160,
-	"Maximum piece size":                                                                       161,
-	"Migrate from existing Lotus-Miner":                                                        162,
-	"Migrating lotus-miner config.toml to Curio in-database configuration.":                    163,
-	"Migrating metadata for %d sectors.":                                                       164,
-	"Miner %s created successfully":                                                            165,
-	"Miner creation error occurred: %s ":                                                       166,
-	"Minimum number of hugepages required":                                                     167,
-	"Minimum piece size":                                                                       168,
-	"Moves funds from the deal collateral wallet into escrow with the storage market actor": 169,
-	"Name of the Postgres database in Yugabyte cluster":                                     170,
-	"New Miner initialization complete.":                                                    171,
-	"New service actor address":                                                             172,
-	"Next steps:":                                                                           173,
-	"No address provided":                                                                   174,
-	"No domain provided, skipping HTTP configuration":                                       175,
-	"No host provided":                                                                      176,
-	"No path provided":                                                                      177,
-	"No path provided, abandoning migration ":                                               178,
-	"No private key provided":                                                               179,
-	"No value provided":                                                                     180,
-	"No, abort":                                                                             181,
-	"No, keep it":                                                                           182,
-	"Non-SP cluster configuration complete":                                                 183,
-	"Non-SP cluster configuration created successfully":                                     184,
-	"Non-SP cluster setup complete!":                                                        185,
-	"Note: This command is intended to be used to verify PoSt compute performance.\nIt will not send any messages to the chain. Since it can compute any deadline, output may be incorrectly timed for the chain.": 186,
-	"Number of 1GB hugepages to configure": 187,
-	"Number of sectors to start":           188,
-	"One database can serve multiple miner IDs: Run a migration for each lotus-miner.": 189,
-	"Optional setup steps (you can skip these and configure later):":                   190,
-	"Other":                              191,
-	"Output file path (default: stdout)": 192,
-	"Owner Wallet: %s":                   193,
-	"PDP":                                194,
-	"PDP (Proof of Data Possession) Configuration":                         195,
-	"PDP configuration layer created":                                      196,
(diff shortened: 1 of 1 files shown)


pdp/handlers.go
```diff
@@ -1197,51 +822,91 @@ func (p *PDPService) handleDeleteProofSetRoot(w http.ResponseWriter, r *http.Req
...
+	// Check if we have this piece or not
+	var found bool
+	err = p.db.QueryRow(ctx, `SELECT EXISTS(SELECT 1 FROM pdp_data_set_pieces WHERE data_set = $1 AND piece_id = $2)`, dataSetId, pieceID).Scan(&found)
+	if err != nil {
+		http.Error(w, "Failed to query piece existence: "+err.Error(), http.StatusInternalServerError)
+		return
+	}
+	if !found {
+		http.Error(w, "Piece not found", http.StatusNotFound)
+	}
```
Reiers: Bug: missing `return` after sending the 404. Execution falls through to the ABI packing below and sends an on-chain delete transaction for a non-existent piece (wastes gas).

```suggestion
		http.Error(w, "Piece not found", http.StatusNotFound)
		return
	}
```

cmd/pdptool/main.go
```diff
@@ -1484,11 +1502,265 @@ var removeRootsCmd = &cli.Command{
...
+		// Get the piece size
+		fi, err := file.Stat()
+		if err != nil {
+			return fmt.Errorf("failed to stat input file: %v", err)
+		}
+		raw_size := fi.Size()
+
+		client := &http.Client{}
+
+		req, err := http.NewRequest("GET", serviceURL+"/pdp/piece/uploads", nil)
```
Reiers: Bug: this sends GET but the route is `r.Post("/piece/uploads", ...)` and the handler explicitly rejects non-POST with 405. The streaming upload command will always fail.

```suggestion
		req, err := http.NewRequest("POST", serviceURL+"/pdp/piece/uploads", nil)
```

pdp/handlers_add.go
```diff
@@ -0,0 +1,504 @@
...
+				return nil, nil, nil, fmt.Errorf("subPieces must be in descending order of size, piece %d %s is larger than prev subPiece %s",
+					i, subPieceEntry.SubPieceCID, addPieceReq.SubPieces[i-1].SubPieceCID)
+			}
+
+			prevSubPieceSize = subPieceInfo.PaddedSize
+			totalSize += uint64(subPieceInfo.RawSize)
+		}
+		// sanity check that the rawSize in the CommPv2 matches the totalSize of the subPieces
+		if rawSize != totalSize {
+			return nil, nil, nil, fmt.Errorf("raw size miss-match: expected %d, got %d", totalSize, rawSize)
```
Reiers: Typo: "miss-match" → "mismatch" (also on line 242)

```suggestion
			return nil, nil, nil, fmt.Errorf("raw size mismatch: expected %d, got %d", totalSize, rawSize)
```

tasks/indexing/task_pdp_ipni.go
```diff
@@ -743,3 +743,48 @@ func (P *PDPIPNITask) Adder(taskFunc harmonytask.AddTaskFunc) {}
...
+
+		pid, err := peer.IDFromPublicKey(pk.GetPublic())
+		if err != nil {
+			return "", xerrors.Errorf("getting peer ID: %w", err)
+		}
+		peerID = pid.String()
+
+		n, err := tx.Exec(`INSERT INTO ipni_peerid (priv_key, peer_id, sp_id) VALUES ($1, $2, $3)`, privKey, peerID, -2)
+		if err != nil {
+			return "", xerrors.Errorf("failed to to insert the key into DB: %w", err)
```
Reiers: Typo: double "to"

```suggestion
			return "", xerrors.Errorf("failed to insert the key into DB: %w", err)
```

tasks/indexing/task_pdp_ipni.go
```diff
@@ -743,3 +743,48 @@ func (P *PDPIPNITask) Adder(taskFunc harmonytask.AddTaskFunc) {}
 
 var _ harmonytask.TaskInterface = &PDPIPNITask{}
 var _ = harmonytask.Reg(&PDPIPNITask{})
+
+// TODO: Remove this once pdpv0 is merged into main
```
Reiers: This TODO says "remove once pdpv0 is merged into main" — but this PR *is* the merge into main. Should this be resolved or reworded?

pdp/handlers.go
```diff
@@ -459,542 +456,179 @@ func (p *PDPService) handleGetProofSet(w http.ResponseWriter, r *http.Request) {
...
-	if err != nil {
-		log.Errorw("Failed to insert into database", "error", err, "txHash", txHashLower, "subroots", subrootInfoMap)
-		http.Error(w, "Internal server error", http.StatusInternalServerError)
-		return
-	}
-
-	// Step 10: Respond with 201 Created
-	w.Header().Set("Location", path.Join("/pdp/proof-sets", proofSetIDStr, "roots/added", txHashLower))
-	w.WriteHeader(http.StatusCreated)
+	http.Error(w, "todo", http.StatusBadRequest)
```
Reiers: Nit: 400 implies the client did something wrong. If deletion is not implemented yet, 501 is more accurate.

```suggestion
	http.Error(w, "Not implemented", http.StatusNotImplemented)
```

pdp/handlers.go
```diff
@@ -459,542 +456,179 @@ func (p *PDPService) handleGetProofSet(w http.ResponseWriter, r *http.Request) {
...
-                   pp.piece_padded_size
-            FROM pdp_piecerefs ppr
-            JOIN parked_piece_refs pprf ON pprf.ref_id = ppr.piece_ref
-            JOIN parked_pieces pp ON pp.id = pprf.piece_id
-            WHERE ppr.service = $1 AND ppr.piece_cid = ANY($2)
-        `, serviceLabel, subrootCIDsList)
-		if err != nil {
-			log.Errorf("Failed to retrieve pdp_piecerefs: %v", err)
-			return false, fmt.Errorf("failed to retrieve pdp_piecerefs")
+	showRemoved := chi.URLParam(r, "XXXshowRemoved") == "true"
```
Reiers: Is this intentionally dead? `chi.URLParam` reads route parameters (like `{dataSetId}`), not query string params. Since `XXXshowRemoved` is not in any route definition, this always returns `""`. If it should work, use `r.URL.Query().Get("showRemoved")` instead.

Reiers (changes requested):
Light review - found a couple of bugs and some small fixes. The big architectural stuff, I'll leave for the team to discuss.

Reiers:
All review suggestions moved to PR #1023 against `pdpv0`. Resolved the threads here.

PR #1020, opened 2026-02-18
fix: small correctness and quality improvements

Reiers:
Small one-liner fixes found during codebase audit cross-check.

### Bug fixes
- **`ClusterTaskHistory` double-scan** (`web/api/webrpc/cluster.go`): The SQL query selects both `id` (history row ID) and `task_id`, but both were scanned into `&t.TaskID` — the first value was silently overwritten by the second. Now discards the unused `id` column.

- **Duplicate alert for sealing tasks** (`alertmanager/alerts.go`): Sealing tasks (SDR, TreeD, etc.) with >5 failures produced the same alert line **twice** — once from the sealing-task check and once from the general >5 threshold check. Changed to `else if` so only one branch fires.

### Typo fixes
- **"failed to to insert"** → "failed to insert" in `task_ipni.go` and `task_pdp_ipni.go` error messages
- **"Command separated"** → "Comma-separated" in `--db-host` / `--db-host-cql` CLI help text + translation catalog

### Missing headers
- **`Content-Type: application/json`** added to `getSchema`, `getLayers` (config API), and `getSectors` (sector API) — other handlers in the same files already set this header.

### DIFF
--- a/cmd/curio/internal/translations/locales/zh/messages.gotext.json
+++ b/cmd/curio/internal/translations/locales/zh/messages.gotext.json
@@ -1440,7 +1440,7 @@
     },
     {
       "id": "Comma-separated list of hostnames for yugabyte cluster",
-      "translation": "Yugabyte 集群的主机名命令分隔列表",
+      "translation": "Yugabyte 集群的主机名逗号分隔列表",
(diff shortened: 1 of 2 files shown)


LexLuthr (changes requested):
Requires minor updates.

PR #1027, opened 2026-02-18
Harmonytask Notify (into integ/task)

snadrus:
HarmonyTask needs to outgrow the main poller:

- it'll mean faster task starts which is great for the PDP pipelines
- Lighter DB usage (this main poller is at the top of DB CPU hit lists) frees it up for more interesting work.
- Actual recording of tasks, owners, claims, and completions still happen in DB, preserving "anything can die" durability.

**Ugly downsides:**

NxN communication: But message-passing only goes to nodes that can do the task.
Messages are small and do not have an encoding (JSON, etc) so they are byte-matched for dispatch.
harmony_task in 30-60MiB RAM: only the not-running parts of tasks you could run.
later could be more efficient by storing the 64-bit values as 16-bit steps atop a 64-bit base.
Remaining:
Today's starvation risk (later tinys starving high-importance big tasks) will persist through this change, but will be resolved by the reservation system that will be built atop this change.
Design:

Peering is RPC-generic (uses Curio's HTTP with websockets)
Peering manages messages, connections, and informs the Scheduler of peer task-adds, peer task-takes
later: peer-task-reserves
The scheduler is also informed of
local task-adds (to try starting or to tell others via peering)
local task-ends as a trigger to start more work.
Delays:

Some delay should be added for pollers adding lists of tasks to be bundled into CanAccept
The scheduler will use polling on occasion, or frequently if some node lacks "modern" peering.
Note: plumbing added (but not enabled) for "reserved" capacity. This will be for an upcoming PR to avoid starvation.

### DIFF
--- a/harmony/harmonytask/scheduler.go
+++ b/harmony/harmonytask/scheduler.go
@@ -27,7 +27,6 @@ type schedulerEvent struct {
 	// This replaces the scheduler's in-memory available task map, ensuring
 	// stale entries (claimed by others, deleted) are garbage-collected.
 	DBTasks map[string][]task
-
 }
 
 // schedulerSource identifies the origin of a scheduler event. Each source
@@ -277,7 +276,7 @@ func (e *TaskEngine) startScheduler() {
 					if _, ok := availableTasks[event.TaskType]; ok {
 						availableTasks[event.TaskType].hasID[event.TaskID] = task{ID: event.TaskID, UpdateTime: time.Now(), Retries: event.Retries}
 						if h := e.taskMap[event.TaskType]; h != nil && h.TimeSensitive {
-							if err := e.pollerTryAllWork(ts, ee); err != nil {
+							if _, err := e.tryStartTask(event.TaskType, ts, ee); err != nil {
 								log.Errorw("failed tryAllWork", "error", err)
 							}
 						} else {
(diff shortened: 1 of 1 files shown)


harmony/harmonytask/peering.go
```diff
@@ -0,0 +1,255 @@
...
+		tasks: make(map[string]bool),
+	}
+	var sql struct {
+		ID    int64  `db:"id"`
+		Tasks string `db:"tasks"`
+	}
+	err = p.h.db.QueryRow(p.h.ctx,
+		`SELECT hm.id, hmd.tasks FROM harmony_machine_details hmd
+		 JOIN harmony_machines hm ON hm.id = hmd.machine_id
+		 WHERE hm.host_and_port = $1`, them.addr).Scan(&sql.ID, &sql.Tasks)
```
LexLuthr:
```suggestion
		 WHERE hm.host_and_port = $1`, them.addr).Scan(&sqlmachinedetails.ID, &sqlmachinedetails.Tasks)
```

harmony/harmonytask/scheduler.go
```diff
@@ -0,0 +1,343 @@
...
+					availableTasks[taskName].reservedTask = shouldReserve
+					e.peering.TellOthers(messageTypeReserve, taskName, shouldReserve)
+				}
+			case <-time.After(e.pollDuration.Load().(time.Duration)): // fast life & early-gather at Go_1.26
+				err := e.waterfall(taskSourceDb{e.db, availableTasks, e.peering, taskSourceLocal{availableTasks, e.peering}}, eventEmitter{e.schedulerChannel})
+				if err != nil {
+					log.Errorw("failed to full waterfall", "error", err)
+					continue
+				}
+			}
```
LexLuthr: Needs documents.

harmony/harmonytask/scheduler.go
```diff
@@ -0,0 +1,343 @@
...
+		stats.Record(context.Background(), TaskMeasures.GpuUsage.M(gpuUsage*100))
+	}
+
+	ramUsage := 1 - float64(availableResources.Ram)/float64(totalResources.Ram)
+	stats.Record(context.Background(), TaskMeasures.RamUsage.M(ramUsage*100))
+
+	return nil
+}
+
+type taskSourceLocal struct {
```
LexLuthr: needs doc

LexLuthr (approved):

PR #1041, opened 2026-02-20
pdp_data_set table robust to proving schedule changes 

### LINKED ISSUE #1031: pdpv0: Sanity check for changing challenge window
We are going to update the challenge window parameter on the mainnet contract.  We should ensure that the pdp_dataset table is getting refreshed with new challenge window values.  I suspect we need to add some code for this.  Probably the right place for this to be useful and not too burdensome is during nextProvingPeriodTask since this runs once a proving period.

ZenGround0:
Closes #1031 

Assuming we are increasing the challenge window it is necessary and sufficient to update the database's challenge window when running the nextProvingPeriod task.   If we run this earlier we risk invalidating the wake up timing logic for late proveTasks and also ensure late scheduling of nextProving period.  

The next challenge epoch/ prove at epoch is determined completely from calls to FWSS so nothing needs to change there.  The only thing we need to change is the db field which is used to determine how soon after prove_at epoch should we retry next_proving period.

I am refreshing both proving period and challenge window generally here but the logic here is only considering the case where these parameters increase.  If we are shrinking proving period or challenge window the behavior is less robust -- i.e. we are going to be late calling next_proving period and late prove tasks will attempt to prove past their deadline.

### DIFF
--- a/tasks/pdpv0/task_next_pp.go
+++ b/tasks/pdpv0/task_next_pp.go
@@ -379,6 +379,7 @@ func (n *NextProvingPeriodTask) processPendingPieceDeletes(ctx context.Context, 
 	return nil
 }
 
+// Note: this function needs revisiting if we are ever *shrinking* proving period or challenge window values
 func (n *NextProvingPeriodTask) refreshProvingPeriod(ctx context.Context, dataSetId int64, provingSchedule *contract.IPDPProvingSchedule) error {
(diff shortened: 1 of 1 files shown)


rjan90:
> If we are shrinking proving period or challenge window the behavior is less robust -- i.e. we are going to be late calling next_proving period and late prove tasks will attempt to prove past their deadline.

I do not think shrinking these params are a viable path at this point in time. So I do not think this is a concern

PR #1043, opened 2026-02-20
10x cheaper C2/PRU: CuZK Proving engine

magik6k:
### Summary

Integrate the `cuzk` persistent GPU SNARK proving daemon with Curio's task scheduler via gRPC. When enabled, Curio delegates PoRep C2, SnapDeals prove, and PSProve SNARK computations to the cuzk daemon instead of spawning per-proof child processes through ffiselect.
- Add gRPC client (`lib/cuzk/`) and SealCalls methods (`lib/ffi/cuzk_funcs.go`) for PoRep and SnapDeals
- Wire cuzk into PoRep, SnapDeals, and PSProve tasks with backpressure via `GetStatus`
- Add `make cuzk` build target (NOT in default BINS — CI unaffected)
- Vendor bellpepper-core and supraseal-c2 crate files so `git clone && make cuzk` works
- Add user-facing documentation under `documentation/en/experimental-features/`

### What is cuzk

cuzk is a persistent Rust daemon that keeps Groth16 SRS parameters (~47 GiB for 32 GiB PoRep) resident in CUDA-pinned host memory across proofs. The current ffiselect model spawns a fresh process per proof, loading the SRS from scratch each time (30-90s). cuzk eliminates this overhead entirely.
Beyond SRS residency, cuzk implements a 13-phase optimization pipeline that achieves **2.8x throughput** over the ffiselect baseline (37.7s/proof vs ~89s on RTX 5070 Ti). The key architectural contributions are pipelined partition synthesis, dual-worker GPU interlock, PCIe transfer optimization, and a split async GPU proving API.

### Architecture
```
Curio (Go)                          cuzk daemon (Rust/CUDA)
─────────────                       ──────────────────────
tasks/seal,snap,proofshare          persistent process
        │                                   │
    gRPC client ────── unix/TCP ──────► gRPC server
  (lib/cuzk/client.go)                     │
                                    ┌──────┴───────┐
                                    │   Scheduler   │
                                    │  (priority Q) │
                                    └──────┬───────┘
                                           │
                                    ┌──────┴───────┐
                                    │  GPU Workers  │
                                    │  (per device) │
                                    └──────────────┘
```
Vanilla proofs are generated locally in Curio (requires sector data on disk), then sent to cuzk for SNARK computation. The returned proof is verified locally before submission.
## Pipelining
The cuzk engine pipelines work at three levels:
### 1. Partition-Level Synthesis → GPU Pipeline (Phase 7)
Instead of synthesizing all 10 partitions of a sector as a batch before any GPU work begins (the ffiselect model), cuzk spawns `partition_workers` concurrent synthesis tasks. Each produces a single partition's ProvingAssignment (~13.6 GiB) and sends it through a bounded channel to the GPU worker:
```
Synthesis Workers:  Part 0Part 2 ...
                         │      │      │
                         ▼      ▼      ▼
GPU Channel:        ─── P0 P1 P2 ───►
                                        │
GPU Worker:                         Prove P0Prove P2 ...
```
The GPU processes partition N while workers synthesize partition N+1..N+k. With `partition_workers=10`, all 10 partitions synthesize concurrently and the GPU is continuously fed.
### 2. Dual-Worker GPU Interlock (Phase 8)
Inside `generate_groth16_proofs_c()`, each partition has ~1.3s of CPU preprocessing (pointer setup, bitmap population) before ~3.3s of CUDA kernels, followed by ~0.7s of CPU epilogue. Phase 8 narrows the C++ mutex to cover only the CUDA kernel region and runs two GPU workers per device:
```
Worker A: CPU prepepilogue══ CUDA ══
Worker B:          CPU prep──epilogue
GPU:               ████ A ████████ B ████
```
This achieves 100% GPU utilization — zero idle gaps between partitions in steady state.
### 3. Split Async GPU API (Phase 12)
Phase 12 splits the monolithic C++ prove call into `prove_start()` (GPU kernels) + `finalize()` (b_g2_msm CPU + proof assembly). The GPU worker releases the lock ~1.7s earlier per partition, immediately picking up the next synthesized partition while b_g2_msm runs in a spawned finalizer task:
```
GPU Worker: ══ CUDA ════ CUDA ══
Finalizer:        b_g2_msm  b_g2_msm  b_g2_msm
```
### SRS Residency
The daemon pre-populates `GROTH_PARAM_MEMORY_CACHE` at startup. Since the process is long-lived, the ~47 GiB SRS stays pinned in CUDA host memory across all proofs. No per-proof disk I/O.
### Per-Partition Working Set
Memory is proportional to `partition_workers`, not total partitions:
| Pipeline stage | Per-partition memory | Notes |
|---|---|---|
| During synthesis | ~16 GiB | 12 GiB a/b/c + 4 GiB aux |
| After prove_start | ~4 GiB | a/b/c freed immediately, only aux + density remain |
| Pending finalize | ~4 GiB | Held by finalizer task |
The formula: **Peak RSS ≈ 69 + (partition_workers × 20) GiB**
Validated configurations:
- 128 GiB system: pw=2, gw=1 → 110 GiB peak, 152s/proof
- 256 GiB system: pw=7, gw=1 → 208 GiB peak, 53s/proof
- 512 GiB system: pw=12, gw=2 → 400 GiB peak, 37.7s/proof
### Backpressure
Three mechanisms prevent OOM at high concurrency:
1. **Early a/b/c free:** `prove_start()` clears 12 GiB/partition immediately after GPU upload
2. **Channel capacity auto-scaling:** bounded to `max(synthesis_lookahead, partition_workers)`
3. **Partition semaphore held through send:** limits total in-flight synthesis outputs
## CPU Locking / GPU Mutex
The C++ `generate_groth16_proofs_c()` originally used a `static std::mutex` that serialized the entire function. cuzk introduces:
- **Heap-allocated mutex** (`create_gpu_mutex()` / `destroy_gpu_mutex()` FFI): one per physical GPU, managed by the engine. Passed through FFI as `*mut c_void`.
- **Narrowed scope:** acquired before per-GPU CUDA kernel launch, released after kernels complete but *before* `prep_msm_thread.join()` — b_g2_msm and proof assembly run outside the lock.
- **Backward compatible:** if `gpu_mtx` is null, falls back to the function-local static mutex (for non-engine callers).
The dual-worker interlock (2 workers per GPU) alternates lock acquisition so Worker B's CPU prep runs while Worker A holds the lock for CUDA kernels, and vice versa.
## Task Integration Details
When `[Cuzk] Address` is set in Curio config:
| Behavior | Change |
|---|---|
| `TypeDetails()` | GPU requirement zeroed, RAM set to 1 GiB (resource decisions delegated to cuzk) |
| `CanAccept()` | Queries `GetStatus` → rejects if `totalPending >= MaxPending` |
| `Do()` | Generates vanilla proof locally → sends to cuzk via `Prove` RPC → verifies returned proof locally |
When `Address` is empty (default), all tasks behave exactly as before. No behavioral change for existing deployments.
## Build
```bash
make cuzk          # builds extern/cuzk → ./cuzk binary (~1m51s from scratch)
make install-cuzk  # installs to /usr/local/bin
make clean         # includes cargo clean in extern/cuzk
make cuzk is NOT in the default BINS or BUILD_DEPS targets, so CI (which has no CUDA) is unaffected. Requires nvcc and cargo.
```
### Files Changed
New files:
- lib/cuzk/client.go — gRPC client wrapper (connect, Prove, GetStatus, HasCapacity)
- lib/cuzk/proving.pb.go, proving_grpc.pb.go — generated protobuf/gRPC stubs
- lib/ffi/cuzk_funcs.go — PoRepSnarkCuzk, ProveUpdateCuzk on SealCalls
- documentation/en/experimental-features/cuzk-proving-daemon.md — user guide
Modified files:
- deps/config/types.go — CuzkConfig struct + defaults
- cmd/curio/tasks/tasks.go — creates cuzk.Client, passes to task constructors
- tasks/seal/task_porep.go — cuzkClient field, Do/CanAccept/TypeDetails branches
- tasks/snap/task_prove.go — same pattern
- tasks/proofshare/task_prove.go — same + threaded through computeProof→computePoRep/computeSnap
- Makefile — cuzk build/install/clean targets
- .gitignore — /cuzk binary
Vendored crate files (for git clone && make cuzk):
- extern/bellpepper-core/ — 13 files (full crate: Cargo.toml, src/, licenses)
- extern/supraseal-c2/ — 8 files (Cargo.toml, build.rs, Cargo.lock, tests)


magik6k:
> Do you plan to do more bench mark on different hardware?

Definitely, but already benchmarked under some pipeline constrains which should be fairly representative.

Generally CuZK is shipped here as an entirely separate deamon which you connect to a Curio node as a 100% optional addon, so generally this integration is as safe as it gets.

I'll run this on more diverse HW once we get this PR in, but I expect it to be universally faster and cheaper:
* This uses less average RAM per proof execute
* It lowers minimum RAM required below even pre-supraseal-c2 requirements
* It optimizes some operations to the point that they are guaranteed to always be faster
* Heavier pipelining guarantees that bigger hardware is much better utilized
* Synth pipelines have memory-bandwidth optimizing semaphores ensuring really memory-bandwidth-expensive compute doesn't overlap

I'd be very surprised if there is any machine configuration that is worse with this.

Eventually over time we might figure out a good way to pull this deamon into the main Curio binary, but initially seems wise to gain confidence in maximally sandboxed way.

FWIW I just verified valid PoSt and PoRep proofs on Calibnet so that's already something.

snadrus (commented):
The Go integration works, but depends on the admin running CuZK ahead of time and on the same server (or else scheduling which still runs gets sideways). 
Can we limit addresses to localhost?

magik6k:
> Can we limit addresses to localhost?

Actually I find myself using non-localhost e.g. to share a GPU between a calib and non-calib node on my smaller setup. CuZK handles queuing/resources internally so that works ok.

snadrus (approved):

PR #1044, opened 2026-02-22
fix: batch timeout timezone bug and move evaluation to testable Go

snadrus:
**Summary**
Fix a timezone bug in the precommit and commit batching triggers that caused batches to take timeout + timezone_offset wall-clock time to fire instead of just timeout
Move the batch firing evaluation from opaque SQL PL/pgSQL into tested Go functions
Add a configurable MaxBatch setting so batches fire immediately when full for each of PreCommit & Commit batching. 

**Problem**
The 20260201-fix-batching-timeout migration fixed the comparison side of the batch timeout check (changing NOW() AT TIME ZONE 'UTC' to NOW() in the SQL batch functions), but left the write side untouched. The three trigger functions (set_precommit_ready_at, set_commit_ready_at, set_update_ready_at) still used CURRENT_TIMESTAMP AT TIME ZONE 'UTC', which returns TIMESTAMP WITHOUT TIME ZONE. When PostgreSQL stores that bare value into a TIMESTAMPTZ column, it reinterprets it using the session timezone. In non-UTC sessions this silently shifts the stored ready-at timestamp, making the effective timeout = configured_timeout + abs(tz_offset).
Before the 20260201 fix, both the write and read sides had the same timezone misuse so the errors canceled out. After 20260201 corrected only the read side, the asymmetry caused batches to take multiples longer than configured, especially visible with steady sector arrivals.

**Fix**
SQL triggers (20260222-fix-trigger-timestamps.sql): Replace CURRENT_TIMESTAMP AT TIME ZONE 'UTC' with NOW() in all three trigger functions. NOW() returns TIMESTAMPTZ directly, storing correctly regardless of session timezone. Also repairs any existing future-dated timestamps stuck by the bug.
Batch evaluation moved to Go (batch_timeout.go): The batch firing decision previously lived inside poll_start_batch_precommit_msgs / poll_start_batch_commit_msgs PL/pgSQL functions as an untestable FOR/IF loop. This PR moves the evaluation into three tested Go functions:
EvalBatchTimeout() — evaluates FULL, SLACK, and TIMEOUT conditions
GroupBatchRows() — aggregates SQL candidate rows into batch candidates
SlackEpochs() / ClampMaxBatch() — shared helpers for config translation
Both pollStartBatchPrecommitMsg and pollStartBatchCommitMsg now: query candidate rows from the DB, group and evaluate in Go, then execute a simple UPDATE ... SET task_id = $1 WHERE sector_number = ANY(...) for the winning batch. The SQL poll_start_batch_*_msgs functions are no longer called.
Configurable max batch size (config.types): New MaxBatch field on PreCommitBatchingConfig and CommitBatchingConfig. When set, the batch fires immediately upon reaching this size (FULL condition), without waiting for timeout or slack. Defaults to 0 (= protocol maximum). Clamped to protocol caps via the shared ClampMaxBatch() function.

**Test plan**
TestEvalBatchTimeout_FullCondition — FULL fires at/above max, not below, disabled when 0
TestEvalBatchTimeout_FullTakesPriority — FULL beats slack and timeout
TestEvalBatchTimeout_SlackCondition — epoch boundary arithmetic
TestEvalBatchTimeout_TimeoutCondition — fires at configured duration, not before
TestBuggyTriggerTimestamp — reproduces the timezone shift bug, proves correct timestamps fire on time
TestBatchTimeoutWithSteadySectorArrival — simulates the reported scenario; shows FULL overrides even a buggy timestamp
TestGroupBatchRows_* — single batch, multi-batch, multi-provider grouping
TestClampMaxBatch — zero/negative/below/at/above protocol cap
TestSlackEpochs — duration-to-epoch conversion with rounding
Verify existing TestPrecommit* and TestCommit* tests still pass

### DIFF
--- a/itests/batching_test.go
+++ b/itests/batching_test.go
@@ -26,12 +26,10 @@ func TestBatching(t *testing.T) {
 	require.NoError(t, err)
 
 	t.Run("PrecommitNoConcurrentEmptyBatch", func(t *testing.T) {
-		const (
-			spID       int64 = 1000
-			sealProof        = 8
-			numSectors       = 20
-			maxBatch         = 10
-		)
+		const spID int64 = 1000
+		const sealProof int64 = 8
+		const numSectors int64 = 20
+		const maxBatch int = 10
 
 		for i := int64(0); i < numSectors; i++ {
 			_, err := db.Exec(ctx, `INSERT INTO sectors_sdr_pipeline (


tasks/seal/batch_timeout.go
```diff
@@ -0,0 +1,134 @@
...
+	Reason     string
+}
+
+// EvalBatchTimeout decides whether a precommit/commit batch should fire.
+// Conditions checked in priority order:
+//
+//   - Full:    batchSize >= maxBatch (batch can't grow further)
+//   - Slack:   (minStartEpoch - slackEpochs) <= currentHeight
+//   - Timeout: now >= earliestReady + timeout
+func EvalBatchTimeout(
```
LexLuthr: The reason I did this completely in SQL was race conditions across the schedulers. Almost all nodes can do precommit and commit so they would end up scheduling phantom batches.

LexLuthr: Last fix was attempted at https://github.com/filecoin-project/curio/pull/394 which still produced empty batches before we moved to SQL batching and getting rid of phantom ones.

tasks/seal/batch_timeout_test.go
LexLuthr: Can we add a test where we have 20-30 sectors requiring batching and then try to batch in parallel with multiple goroutines. It would confirm if we have any regression.

Issue #1045, opened 2026-02-22
pdpv0: ResolveViewAddress can block indefinitely, hanging the chain scheduler


TippyFlitsUK:
## Problem

`ResolveViewAddress` calls `ViewContractAddress(nil)` — nil `CallOpts` means the underlying EthCall uses `context.Background()` with no timeout. If the go-ethereum ethClient's WebSocket connection enters a half-open state (TCP alive, no application data), the call blocks indefinitely.

This blocks `CurioChainSched.update()`, which runs all handlers sequentially. A single stuck handler prevents all subsequent handlers from executing and prevents the chain scheduler from processing new block notifications. ChainNotify messages pile up in the go-jsonrpc buffer (observed n=970+), causing a cascading backlog across all chain-dependent tasks including PDP proving.

I observed frequent recurring backlogs across multiple nodes, with the chain scheduler stalling for extended periods (up to 397 minutes in one instance) and requiring restarts to recover.

## Evidence

**pprof goroutine dump** from an affected node showed the exact stuck call stack:

```
goroutine 1729 [select, 397 minutes]:
  rpc.(*requestOp).wait()
    rpc.(*Client).CallContext()
      ethclient.(*Client).CallContract()
        bind/v2.(*BoundContract).call()
          contract.(*ContractWithViewCaller).ViewContractAddress()
            contract.ResolveViewAddress()
              pdpv0.processPendingTerminations()     <- watch_fwss_terminate.go:57
                chain_sched.(*CurioChainSched).update()
```

**Log analysis** confirmed that for the stuck block height, watchers[0-1] and callbacks[0-3] completed normally, but callback[7] (`TerminateServiceWatcher`) never returned — blocking callbacks[8] and all subsequent blocks.

**Observed on** both calibnet and mainnet nodes running the `pdpv0` branch, after deploying code containing commit `646852bf` which introduced `ResolveViewAddress` calls to the FWSS watchers.

## Root Cause

`ResolveViewAddress` was introduced in `646852bf` (PR #1006). It makes an EthCall to resolve the view contract address. The function passes `nil` as `CallOpts`, which provides no timeout protection. The go-ethereum `rpc.Client` WebSocket transport has no automatic reconnection — once the connection dies, pending calls block forever.

## Workaround

[...]

PR #1046, opened 2026-02-23
fix(pdp): add timeout to eth_call ops & cache ResolveViewAddress calls

rvagg:
Fixes: #1045

Unfortunately we don't inherit any context management from Harmony so we should be doing it ourselves. There might be an even better approach here at the top-level `Do()` for each task to cap a maximum timeout and make sure that everything _within_ the task inherits the context so at least have per-context cleanup.


redpanda-f (changes requested):
Nothing controversial here, apart from the fact that this does not already feature on main. Although I see no reason why it should not. Would love @LexLuthr buy-in here.

With anything Cache, I would like to see safety arguments though!

PR #1052, opened 2026-02-24
pdpNotification needs timeouts

snadrus:
Lets time out the HTTP interaction, and keep consistency by wrapping the 2 writes in a transaction.


rvagg (approved):
fwiw we're not actively using the notify call in the plain foc flow so I'd also be fine removing that entirely .. and renaming this task to "finalize_task" like it probably always should have been

PR #1058, opened 2026-02-26
feat(pdpv0): cap exponential backoff to a max limit for retries

redpanda-f:
Introduces a new helper function `RetryWaitExpWithMax` that caps exponential backoff to a max limit. This change makes sure no `pdpv0` task has unbounded exponential waits.

### DIFF
--- a/harmony/taskhelp/retry.go
+++ b/harmony/taskhelp/retry.go
@@ -20,3 +20,10 @@ func RetryWaitExp(initial time.Duration, factor float64) RetryWaitFunc {
 		return time.Duration(float64(initial) * math.Pow(factor, float64(retries)))
 	}
 }
+
+// RetryWaitExpWithMax returns a function that calculates an exponentially increasing duration based on retries, with a maximum cap.
+func RetryWaitExpWithMax(initial time.Duration, factor float64, max time.Duration) RetryWaitFunc {
+	return func(retries int) time.Duration {
+		return min(time.Duration(float64(initial)*math.Pow(factor, float64(retries))), max)
+	}
+}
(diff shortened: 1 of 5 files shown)


harmony/taskhelp/retry.go
```diff
@@ -20,3 +20,10 @@ func RetryWaitExp(initial time.Duration, factor float64) RetryWaitFunc {
 		return time.Duration(float64(initial) * math.Pow(factor, float64(retries)))
 	}
 }
+
+// RetryWaitExpWithMax returns a function that calculates an exponentially increasing duration based on retries, with a maximum cap.
+func RetryWaitExpWithMax(initial time.Duration, factor float64, max time.Duration) RetryWaitFunc {
+	return func(retries int) time.Duration {
+		return min(time.Duration(float64(initial)*math.Pow(factor, float64(retries))), max)
```
ZenGround0: Probably clearer if you replaced `time.Duration(float64(initial)*math.Pow(factor, float64(retries)))` with `RetryWaitExp`

tasks/pdpv0/task_next_pp.go
```diff
@@ -408,7 +408,7 @@ func (n *NextProvingPeriodTask) TypeDetails() harmonytask.TaskTypeDetails {
 			Ram: 1 << 20,
 		},
 		MaxFailures: 3, // Set retry limit to 3 attempts
-		RetryWait:   taskhelp.RetryWaitExp(5*time.Second, 2),
+		RetryWait:   taskhelp.RetryWaitExpWithMax(5*time.Second, 2, 5*time.Minute),
```
ZenGround0: Why are you adding this? 2^3*5 seconds is < 1 minutes so this is adding complexity for a noop.  Better to leave it off.

LexLuthr (commented):
Why are we doing this? Why does retry needs an explicit cap? Does harmony `MaxFailure` not working?

Issue #1062, opened 2026-02-26
pdpv0: Testing `Verify()` against `go-fil-commp-hashhash`


redpanda-f:
This task is an offshoot of https://github.com/filecoin-project/curio/issues/888 with the intention of providing tests surrounding `Verify()` function available here: https://github.com/filecoin-project/curio/blob/9901486555aeec1befa9a52e830e1de632e9e755/tasks/pdpv0/task_prove.go#L1035-L1058. Verification is needed since `savecache.go` lib clearly hints at it being a copy/paste of commp-hashash: https://github.com/filecoin-project/curio/blob/9901486555aeec1befa9a52e830e1de632e9e755/lib/savecache/savecache.go#L20-L21

The main difference that seems to feature in `savecache.go` implementation and commp-hashhash is the `hashSlab254(...)` function introducing a savecache. See: https://github.com/filecoin-project/curio/blob/9901486555aeec1befa9a52e830e1de632e9e755/lib/savecache/savecache.go#L370-L378

Per https://github.com/filecoin-project/curio/pull/997#issuecomment-3920190663, the understanding goes:
> commp-hashhash has a great set of fixtures that we could and should use, it builds trees, it has trustworthy roots for stable base data, the only gap is that hashhash gives us data and CommP roots but not merkle proofs. What we can do is, for various sizes and input data, build a set of fixtures that use the same base layer and commp root and a set of challenges with the merkle paths for those challenges across the base at interesting points so we can get good coverage.

commp-hashhash seems to be available here: https://github.com/filecoin-project/go-fil-commp-hashhash, with random test vectors here: https://github.com/filecoin-project/go-fil-commp-hashhash/blob/master/testdata/random.txt

## Done Criteria
- TODO

redpanda-f:
Will instead be closed by #1067 and #1068 as #1065 is superseded

PR #1065, opened 2026-02-27
tests(pdpv0): verify validity of `Verify()` against `commp-hashhash`

redpanda-f:
Parent issue: https://github.com/filecoin-project/curio/issues/1062

### DIFF
--- a/tasks/pdpv0/verify_test.go
+++ b/tasks/pdpv0/verify_test.go
@@ -17,11 +17,12 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/filecoin-project/go-state-types/abi"
-	"github.com/filecoin-project/lotus/lib/nullreader"
 
 	"github.com/filecoin-project/curio/lib/proof"
 	"github.com/filecoin-project/curio/lib/savecache"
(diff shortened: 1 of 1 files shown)


tasks/pdpv0/verify_test.go
```diff
@@ -0,0 +1,824 @@
...
+// pad payloads shorter than their declared unpadded piece size, matching how
+// Filecoin pads under-sized pieces with trailing zeros.
+type zeroReader struct{}
+
+func (z *zeroReader) Read(p []byte) (n int, err error) {
+	for i := range p {
+		p[i] = 0
+	}
+	return len(p), nil
+}
```
rvagg: replace all of this with `nullreader.NewNullReader`, already used through curio, including in the code we care about. I also added an improvement for it https://github.com/filecoin-project/lotus/pull/13529 so we'll pick that up too eventually.

redpanda-f: fixed in [c48d1c4](https://github.com/filecoin-project/curio/pull/1065/commits/c48d1c492a11f224cb291443b1086158879eb1a6)

PR #997, opened 2026-02-16
feat(pdpv0): introduce PDP Save Cache and Proving Task

redpanda-f:
Part of #888, to be applied post #996 

The TaskPDPSaveCache in `task_save_cache_pdpv0.go` was written for a different branch and relied on mk20:
- not available on pdpv0
- `pdp_pipeline` table - doesn't exist on `pdpv0`, we are not introducing this
- Piece CID v2 storage - pdpv0 stores v1 CIDs

This caused compilation failures and runtime errors when trying to use the save cache functionality on the pdpv0 branch.

- Depends on (introduce main alike indexstore in pdpv0): [https://github.com/filecoin-project/curio/pull/996](https://github.com/filecoin-project/curio/pull/996) as base layer, blocked by this PR
- Introduces two new column to `pdp_piecerefs`: 
  - `needs_save_cache BOOLEAN DEFAULT FALSE`, whether cache saving is needed for the piece?
  - `save_cache_task_id BIGINT DEFAULT NULL`, whether some save cache task_id is allocated
- New indexing: `WHERE save_cache_task_id IS NULL AND needs_save_cache = TRUE`
- Renamed task to "PDPv0_SaveCache" for clarity

Issue #888, opened 2026-01-15
PDP: cache merkle tree layer for proof generation to reduce RAM 100-1000x


rvagg:
Worth recording this, mentioned by @magik6k @ https://filecoinproject.slack.com/archives/C07CGTXHHT4/p1768466246974369?thread_ts=1767958899.957719&cid=C07CGTXHHT4

- Current: ProveTask builds full memtree in RAM (~3x piece size, up to 3 GiB for max pieces)
- Proposed: Adapt TreeD code to output a middle layer, cache in CQL (kilobytes), construct proofs from cached layer
- Benefit: Massive RAM reduction, enables higher parallelism, probably significantly quicker
- Reference: existing TreeD implementation that creates tree on disk

LexLuthr:
This already exists https://github.com/filecoin-project/curio/blob/main/tasks/pdp/task_save_cache.go and prove task in main using it.

BigLep:
There is good discussion from @rvagg in https://filecoinproject.slack.com/archives/C08TVNKJV7C/p1769558578248019?thread_ts=1769528497.089949&cid=C08TVNKJV7C about next steps.  It provides better and more specific information/ideas than my thoughts from 2026-01-23.  I'm quoting/summarizing it below as I understand it.

The original problem where this came up is because the task provisioning system was told these tasks only needed ~1.6MiB of ram (mistaken calculation), so it was overprovisioning them and causing OOM. Now it’s being told to account for ~2GiB so it won’t do as many in parallel. But that could potentially be a bottleneck on a system with a large number of data sets with challenge windows close to each other.

This isn’t an exercise in getting to perfect, we’re simply aiming for good. Perfect can come later cause we just need to launch this thing and get real people paying real money for it, we can amazify it later when we figure out what the biggest concerns are. We suspect generating proofs is not going to be a primary concern. What we have now on `pdpv0` is probably going to be a bottleneck, but SPs do have beefy resources to throw at this so it could even end up being something that could be ignored if we didn’t want to care about it.

Next steps:
1. Understand what the algorithm is in `main` and the tradeoffs it makes.  We don’t need high quality performance information, but understanding what the costs are and how those costs should be changed with different tunables or even potentially replacing the algorithm entirely. The costs to look are: CPU, RAM, I/O (including I/O patterns given that it’s designed to be read across a network), and database use.
2. Add some additional assurance via testing that `main` is doing the right thing and generating appropriate proofs for various cases that we can easily come up with. 
[...]

PR #1067, opened 2026-03-02
refactor(proof): extract cached proof pipeline behind testable interfaces (pdpv0)

rvagg:
_Pulls in some of #1065, now in lib/proof/verify_test.go_

Implements the testing assurance work outlined in #888 and the #997 review thread (https://github.com/filecoin-project/curio/pull/997#issuecomment-3920190663), specifically:

- Fixture-based testing using commp-hashhash test vectors as external trust anchor (11 sizes from 127 B to 512 MiB, validated against filecoin-ffi)
- `VerifyProof` moved to lib/proof/ (was `Verify` in tasks/pdpv0/task_prove.go), decoupled from contract types, independently testable
- `proveSubPieceCached` extracted into `GenerateCachedProof` behind `PieceReader` and `ProofCache` interfaces (following the handlers_pull.go pattern described in the #997 review thread for running proof logic without Harmony/DB/CQL dependencies)
- Production proof pipeline tested as a unit: cache lookup, section reader setup, sub-memtree construction, snapshot node matching, split proof combination

The `Verify` tests draw from #1065: manual tree construction and negative test structure. The main addition beyond #1065 is the interface extraction and `TestGenerateCachedProof` / `TestCacheSplitProof`, which exercise the actual production function with mocked dependencies rather than testing the underlying algorithms in isolation (also the main weakness of the itest).

Test summary (9 fixtures locally, 11 in CI -- large fixtures gated on `CI=true`):
- `TestMemtreeRootsMatchCommPFixtures`: 11 fixtures, external anchor
- `TestProofRoundTrip`: 11 fixtures x 5 positions + negative tests
- `TestCacheSplitProof`: 8 fixtures x 4 positions, step-by-step validation
- `TestGenerateCachedProof`: 8 fixtures x 4 positions, black-box production fn
- `TestManualEightLeafTree`: hand-built tree, independent of memtree
- `Verify` negative tests: wrong leaf, tampered sibling, truncated, swapped
- `TestSavecacheRootsMatchFixtures`: `savecache.Calc` vs fixture cross-check

The diff looks larger than it is, mostly this is moving things around. This also cleanly maps to `main` branch work where the same refactored code can be used by both pdpv0 and pdpv1 tasks just with different injected dependencies. Separate PR for that: https://github.com/filecoin-project/curio/pull/1068 - merging this back to main will be a trivial exercise since they match and now use the same code paths.


LexLuthr (approved):
Looks good. I just have a minor nit which should not cause any issues.

PR #1087, opened 2026-03-11
Mk20 updates

LexLuthr:
Pending MK20 TODO:

- [x] market/mk20/mk20.go:196 - TODO: Backpressure, client filter
- [x] market/mk20/ddo_v1.go:178 - TODO: Review if care about client at all
- [x] market/mk20/ddo_v1.go:213 - TODO: Guard against start duration here. Maybe use a config to allow minimum now+2days to max now+7days
- [x] tasks/storage-market/mk20.go:1097 - TODO: Use correct start duration
- [x] Make a decision on Auth and find a way to authenticate contracts as client (original client will Auth)
    Ideas:
        - ~~Make contracts have designated wallet~~
        - ~~Use registry~~
        - ~~Use zk proof in any reliable way with off chain verification~~
        - ~~Keep contracts unauthorised~~
        - ~~Event based deal creation (still uploads require auth)~~


harmony/harmonydb/downgrade/20260416-mk20-ddo-contracts.sql
```diff
@@ -0,0 +1,76 @@
+-- This function triggers a download for an offline piece.
```
Reiers, line 1: **Blocking — broken downgrade.** See body finding #1.

Add at the very top:
```sql
DROP FUNCTION IF EXISTS process_offline_download(TEXT, TEXT, TEXT, BIGINT, BIGINT, TEXT);
DROP TABLE IF EXISTS ddo_contracts;
```

Without these, rollback leaves (a) an orphan `ddo_contracts` table and (b) both the old 5-arg and new 6-arg `process_offline_download` functions installed simultaneously.

harmony/harmonydb/sql/20260416-mk20-ddo-contracts.sql
```diff
@@ -0,0 +1,89 @@
+CREATE TABLE IF NOT EXISTS ddo_contracts (
+    address TEXT PRIMARY KEY,
+    allowed BOOLEAN NOT NULL DEFAULT FALSE
+);
+
+-- Delete the older process_offline_download
+DROP FUNCTION IF EXISTS process_offline_download(TEXT, TEXT, TEXT, BIGINT, TEXT);
```
Reiers, line 7: `process_offline_download` now has a 6-arg signature (`_raw_size BIGINT` added). Please verify every Go call site before merge:

```
grep -rn "process_offline_download" --include='*.go'
```

Any 5-arg caller will fail at runtime with `function process_offline_download(...) does not exist`.

market/mk20/filters.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+func (m *MK20) applyAllowList(ctx context.Context, client string) (bool, error) {
+	var allowed sql.NullBool
+	err := m.DB.QueryRow(ctx, `SELECT status FROM market_allow_list WHERE wallet = $1`, client).Scan(&allowed)
+	if err != nil {
+		if !errors.Is(err, pgx.ErrNoRows) {
+			return false, xerrors.Errorf("failed to query the allow list status from DB: %w", err)
+		}
+		return !m.cfg.Market.StorageMarketConfig.MK20.DenyUnknownClients, nil
```
Reiers, line 19: **Inconsistent with MK12.** `market/mk12/mk12.go:815-835` does `StateLookupID` first, then queries with both the as-passed address **and** the ID form:

```go
err = m.db.QueryRow(ctx,
    `SELECT status FROM market_allow_list WHERE wallet = $1 OR wallet = $2`,
    originalClient.String(), idClient.String()).Scan(&allowed)
```

MK20 only checks one form. Since `market_allow_list` is shared with MK12, an operator allowlisting `f1abc...` will see MK20 deals with `f0123` silently rejected while MK12 deals from the same wallet pass. Real footgun.

Suggest plumbing the `MK20API` into this struct and matching MK12's pattern.

market/mk20/auth.go
```diff
@@ -0,0 +1,154 @@
...
+		return "", nil, nil, fmt.Errorf("invalid signature")
+	}
+
+	return keyType, pubKey, sig, nil
+}
+
+func verifySignature(db *harmonydb.DB, keyType string, pubKey, signature []byte, requestMethod string, requestPath string, cfg *config.CurioConfig) (bool, string, error) {
+	now := time.Now().UTC().Truncate(time.Minute)
+	timeStamps := []time.Time{now, now.Add(-time.Minute)}
+	var msgs [][32]byte
```
Reiers: **Clock-skew bug.** Only tolerates server clock ahead of client. If the client clock is 30s ahead, the client signs `15:01`, the server sees `15:00` and `14:59`, signature fails.

```go
timeStamps := []time.Time{now, now.Add(-time.Minute), now.Add(time.Minute)}
```

Also: 2-minute replay window is generous for a state-mutating protocol. Consider 1 minute + nonce store.

market/mk20/auth.go
```diff
@@ -0,0 +1,154 @@
...
+	return false, "", errors.New("invalid signature")
+}
+
+func AuthenticateClient(db *harmonydb.DB, id, client string) (bool, error) {
+	var allowed bool
+	err := db.QueryRow(context.Background(), `SELECT EXISTS (SELECT 1 FROM market_mk20_deal WHERE id = $1 AND client = $2)`, id, client).Scan(&allowed)
+	if err != nil {
+		return false, xerrors.Errorf("querying client: %w", err)
+	}
+	return allowed, nil
```
Reiers, line 124: **Raw string address comparison.** If the deal stored `client = f0123` and auth derives `f1abc...` (same wallet), this fails.

Either normalize both sides to ID form when storing + querying, or do `StateLookupID` here and query `WHERE id = $1 AND (client = $2 OR client = $3)` with both forms.

Same issue in `market/mk20/utils.go:84` (`validateClient`). Silent failure mode where users can't access their own deals if the address form drifted.

market/mk20/auth.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+func clientAllowed(ctx context.Context, db *harmonydb.DB, client string, cfg *config.CurioConfig) (bool, error) {
+	var allowed bool
+	err := db.QueryRow(ctx, `SELECT allowed FROM market_mk20_clients WHERE client = $1`, client).Scan(&allowed)
+	if err != nil {
+		if errors.Is(err, pgx.ErrNoRows) {
```
Reiers, line 131: **Second, overlapping allowlist.** `market_mk20_clients` here vs `market_allow_list` in `filters.go`. They can disagree: client allowed in one but not the other produces confusing rejections.

Suggest dropping `market_mk20_clients` and using `market_allow_list` for both gates (matches MK12). Also uses `context.Background()` — should accept the request context.

market/mk20/ddo_v1.go
```diff
@@ -22,6 +21,7 @@ import (
 	"github.com/filecoin-project/curio/deps/config"
 	"github.com/filecoin-project/curio/harmony/harmonydb"
```
Reiers, line 22: Typo: `ErrUnknowContract` → `ErrUnknownContract`. Exported, so will be awkward to rename later.

market/mk20/ddo_v1.go
```diff
@@ -98,88 +97,154 @@ func (d *DDOV1) Validate(db *harmonydb.DB, cfg *config.MK20Config) (DealCode, er
...
-		return v.Int64(), Ok, nil
+	if deal.Data == nil {
+		return ErrBadProposal, xerrors.Errorf("deal data is required for market verification")
 	}
 
-	var abiStr string
-	err := db.QueryRow(ctx, `SELECT abi FROM ddo_contracts WHERE address = $1`, d.ContractAddress).Scan(&abiStr)
+	var allowed bool
+	err := db.QueryRow(ctx, `SELECT allowed FROM ddo_contracts WHERE address = $1`, d.MarketAddress).Scan(&allowed)
 	if err != nil {
```
Reiers, line 144: **EVM address case sensitivity.** `ddo_contracts.address` is looked up with exact-string match. If an operator inserts `0xAbC...` (checksummed) and a deal arrives with `0xabc...` (lowercase), lookup returns `ErrNoRows` → `ErrUnknowContract`. Silent.

```go
err := db.QueryRow(ctx,
    `SELECT allowed FROM ddo_contracts WHERE lower(address) = lower($1)`,
    d.MarketAddress).Scan(&allowed)
```

Also lowercase on insert. Same applies to `PDPV1.RecordKeeper` in `pdp_v1.go`.

Reiers (commented):
Comprehensive review pass. The PR description marks the TODOs as complete, but several resolutions are incomplete, inconsistent with adjacent MK12 code, or silently buggy. I'm putting the high-severity findings in the body here and the more localized ones as inline comments, because GitHub only allows inline on lines that appear as additions in this diff.

---

## Blocking for merge

### 1. Broken SQL downgrade — `harmony/harmonydb/downgrade/20260416-mk20-ddo-contracts.sql`

The downgrade file does **not** undo the schema additions in the forward migration:

1. `ddo_contracts` table is never dropped. After rollback, an orphan table with allowlist rows persists, referencing code paths that no longer exist.
2. `process_offline_download` has a new 6-arg signature (`_raw_size BIGINT` added) in the forward migration. Postgres treats different signatures as distinct functions. The downgrade creates the 5-arg version **alongside** the new 6-arg version rather than replacing it. Both coexist silently.

Compare against the pattern in `20260410-ipni-head-cas.sql` which correctly does `DROP FUNCTION IF EXISTS ...` with the full new signature first.

**Suggested fix at the top of the downgrade file:**
```sql
DROP FUNCTION IF EXISTS process_offline_download(TEXT, TEXT, TEXT, BIGINT, BIGINT, TEXT);
DROP TABLE IF EXISTS ddo_contracts;
```

### 2. `Content-Type` set after `WriteHeader` in 4 handlers — `market/mk20/http/http.go`

Lines 312-313 (pattern repeats at 348, 389-390, 423-424):
```go
w.WriteHeader(http.StatusOK)              // commits headers
w.Header().Set("Content-Type", "application/json")  // silently ignored
```

Go's `http.ResponseWriter` commits all headers on `WriteHeader`. Subsequent `Header().Set()` calls are discarded; clients receive `text/plain`. Swap the order in all four handlers (`mk20supportedContracts`, `supportedProducts`, `supportedDataSources`, and the write path of `mk20status`).

### 3. `process_offline_download` signature change — verify all Go callers

The forward migration adds a `_raw_size BIGINT` parameter. Every Go call site invoking this function needs to pass the new argument or the call will fail at runtime with `function ... does not exist`. Grep and audit:
```
grep -rn "process_offline_download" --include='*.go'
```

### 4. Cross-PR break with #1163

Multiple `ON CONFLICT (piece_cid, piece_padded_size, long_term, cleanup_task_id) DO NOTHING` clauses exist in this PR:
- `market/mk20/mk20.go:752, 852`
- `market/mk20/mk20_upload.go:361, 767`
- `tasks/storage-market/mk20.go:380, 502`

#1163 (`fix: make parked_piece entries truly unique`, approved and on deck for the same release) replaces the 4-column unique constraint with a partial unique index `parked_pieces_active_piece_key` on `(piece_cid, piece_padded_size, long_term) WHERE cleanup_task_id IS NULL`. Once #1163 lands, Postgres errors with `there is no unique or exclusion constraint matching the ON CONFLICT specification`.

**Fix for all six call sites once #1163 merges:**
```sql
ON CONFLICT (piece_cid, piece_padded_size, long_term) WHERE cleanup_task_id IS NULL DO NOTHING
```

Coordinate merge order — whichever lands second needs a follow-up commit.

---

## Serious (functional / security)

### 5. `applyAllowList` missing `StateLookupID` — diverges from MK12

`market/mk20/filters.go:12-27` only queries `market_allow_list` with the wallet as-passed. Compare `market/mk12/mk12.go:815-835` which looks up both the as-passed address and the ID form:

```go
// MK12 (correct):
client, err := m.api.StateLookupID(ctx, deal.ClientDealProposal.Proposal.Client, types.EmptyTSK)
err = m.db.QueryRow(ctx,
    `SELECT status FROM market_allow_list WHERE wallet = $1 OR wallet = $2`,
    originalClient.String(), client.String()).Scan(&allowed)
```

`market_allow_list` is shared between MK12 and MK20. An operator who allowlists `f1abc...` will see MK20 deals submitted with `f0...` silently rejected while MK12 deals from the same wallet pass. Real operator footgun.

**Fix:** plumb the `MK20API` into `MK20` (already there) and do the lookup here.

### 6. Two overlapping client allowlists

This PR introduces `market_mk20_clients` (checked in `auth.go:146-149` during HTTP auth) while `market_allow_list` (checked in `filters.go:19` during deal sanitization) still exists. The two can disagree:

- Client allowed in `market_mk20_clients` but not `market_allow_list` → auth passes, deal rejected at sanitization with confusing error.
- Client allowed in `market_allow_list` but not `market_mk20_clients` → cannot even authenticate.

Pick one model. My suggestion: drop `market_mk20_clients` and use `market_allow_list` for both gates (matching MK12). If operators need an MK20-only denylist, extend the schema rather than fork the table.

### 7. Auth replay window + clock-skew asymmetry — `auth.go:70-75`

```go
timeStamps := []time.Time{now, now.Add(-time.Minute)}
```

Two issues:
1. The 2-minute replay window is generous for a protocol that mutates state (creates deals, modifies pipelines). Consider 1 minute, or add a server-side seen-nonce store.
2. Only tolerates server clock ahead of client. A client whose clock is 30s **ahead** fails auth (client signs `15:01`, server sees `15:00` and `14:59`). Add `now.Add(time.Minute)` to the slice to tolerate both directions.

### 8. Asymmetric start-epoch validation — `mk20.go:354-366`

```go
if *deal.Products.DDOV1.StartEpoch > head.Height()+sealDuration {
    return rejectionInfo(ErrBadProposal, "Start epoch is greater than...")
}
```

Only the upper bound is enforced. Client can submit `StartEpoch = head.Height()+10` and pass ingestion. The sealing pipeline in `tasks/storage-market/mk20.go:1072-1075` uses the client-provided start directly, so the deal enters sealing with a start that's already past by the time PoRep completes. Result: guaranteed PreCommit failure.

The TODO in the PR description explicitly said "minimum now+2days to max now+7days". Only the max is live.

**Fix:** add a minimum:
```go
if *deal.Products.DDOV1.StartEpoch < head.Height()+minStartOffset {
    return rejectionInfo(ErrBadProposal, "Start epoch too soon to seal in time")
}
```
With `minStartOffset` at 2 days (like the default fallback in the task) or config-driven.

### 9. `validateClient` / `AuthenticateClient` — raw string address comparison

Same class of bug as #5. `market/mk20/utils.go:84` compares `client != addr.String()` and `market/mk20/auth.go:139` queries `client = $2`. Both raw string, no `StateLookupID` normalization. A user who creates a deal with their `f0` ID will not be able to authenticate with their `f1` pubkey-derived address, and vice versa.

Normalize both sides to ID form when storing and comparing, **or** query both forms explicitly.

### 10. EVM address case sensitivity — `ddo_contracts` and friends

`ddo_v1.go:144` and `pdp_v1.go:82` compare EVM addresses as strings. An operator storing `0xAbC...` in `ddo_contracts.address` will fail to match a deal submitted with `0xabc...`. Normalize to lowercase on both insert and lookup:

```go
err := db.QueryRow(ctx,
    `SELECT allowed FROM ddo_contracts WHERE lower(address) = lower($1)`,
    d.MarketAddress).Scan(&allowed)
```

Same for `PDPV1.RecordKeeper`.

### 11. Backpressure incomplete — `market/backpressure/backpressure.go:203`

The TODO marked as complete in the PR description is only partially done:
- ✅ running pipelines, downloading pending, commp pending
- ❌ aggregation pending (`aggPending` is computed in the query, never thresholded)
- ❌ upload queue (HTTP PUT chunk uploads)
- ❌ offline queue

Clients can flood `/uploads/*` without any backpressure gate. Either complete the thresholds or explicitly note which pieces are deferred.

### 12. Unpinned `go-car` in `Dockerfile`

This PR reverts `go-car` from the pinned `v2.16.0` (pinned in #1106 during the Go 1.25→1.26 transition) back to `@latest`. No commit message explanation. Concerns: reproducibility, supply chain, and Go 1.26 compatibility assumptions. If the pin is no longer needed, please document why.

### 13. Misleading error messages — `utils.go:521-555`

`IsDataSourceEnabled` and `IsProductEnabled`:
1. Return `http.StatusInternalServerError` (literal 500) cast to `DealCode` on DB errors. Happens to numerically equal `ErrServerInternalError` but fragile.
2. On real DB errors, report `"data source %s is not enabled"` — if Postgres is down, the user is told the feature is disabled.
3. `IsProductEnabled` has a copy-paste typo: reports `"data source %s is not enabled"` when checking **products**.

---

## Cleanup / minor (see inline comments where possible)

- `unknowClient` field (`mk20.go:64`) set but never read. Also name is a typo.
- `ErrUnknowContract` (`ddo_v1.go:22`) — typo, exported.
- `UpdateDealWithTx` and `UpdateDeal` in `db_utils.go` are byte-identical. Remove `UpdateDealWithTx` (it has no callers).
- `disabledMiners` on `MK20DealHandler` (`http.go:47`) loaded from `MK12.DisabledMiners` (wrong config path) and never read. Remove or wire correctly.
- `dealRateLimitMiddleware` mounted twice — once on `Router`, once on `APIRouter`. Either redundant or effectively halves the allowed rate per API route.
- 50/sec/IP rate limit is global, including the high-frequency `PUT /uploads/{id}/{chunkNum}` chunk endpoint. Use per-route limits.
- `requestTimeout = 10s` on `/deal` is aggressive given chain lookups + eth RPC + DB writes. 30s recommended.
- `PDPV1.Validate` (`pdp_v1.go:109-124, 132-148, 150-169`) checks dataset existence without `client` filter. `sanitizePDPDeal` re-checks **with** client filter. Wastes a DB hit and leaks existence (different error messages between the two checks).
- `eth_keys WHERE role='pdp'` (`pdp_v1.go:67`) runs on every PDP deal submit. Cache it.
- `context.Background()` used in `clientAllowed`, `IsDataSourceEnabled`, `IsProductEnabled`. Should accept/use the request context.
- `localClientIDBytes` (`ddo_v1.go:187-193`) silently falls back to raw string bytes if address parsing fails, leading to opaque "rejected by market" errors downstream. Propagate the parse error.
- `MarketAddress` validation in `ddo_v1.go:100-110`: the `len < 2` and `[0:2] != "0x"` checks are redundant — `common.IsHexAddress` already handles both correctly.

---

## Test coverage

Only change to `market/mk20/types_test.go` is a type fix for an existing test. No new tests for:
- Market contract verification flow (`VerifyMarketDeal` + `ddo_contracts` lookup + case sensitivity)
- Two-table client allowlist interaction
- Clock-skew auth
- New `/contracts` endpoint
- The new 6-arg `process_offline_download` signature
- `StartEpoch` lower-bound validation (none, per #8)

For a 4500+ line PR with security-critical paths, this is thin. At minimum recommend a regression test for the DDO + MarketAddress + AllocationId end-to-end path.

---

## Recommendation

Not merge-ready as-is. Items #1, #2, #3, #4 are unambiguous bugs that will bite in production (downgrade, Content-Type, cross-PR constraint, signature change). Items #5, #6, #9 (address normalization + dual allowlist) need a design decision before this ships — operators will hit them on day one. Happy to pair on any of the fixes.

~ @Reiers

market/mk20/ddo_v1.go
```diff
@@ -98,88 +97,154 @@ func (d *DDOV1) Validate(db *harmonydb.DB, cfg *config.MK20Config) (DealCode, er
...
+	if d.StartEpoch != nil {
+		startEpoch = new(big.Int).SetUint64(uint64(*d.StartEpoch))
 	}
 
-	// Call contract
-	output, err := eth.CallContract(ctx, msg, nil)
+	mdeal := mk20contract.ICurioDealViewV1CurioDealView{
+		DealId:          new(big.Int).SetUint64(*d.MarketDealID),
+		State:           mk20contract.DealStatusOpen,
+		ProviderActorId: new(big.Int).SetUint64(localProviderID),
```
Reiers, line 192: If `client` fails to parse, this silently returns raw string bytes. Downstream `VerifyDeal` gets arbitrary bytes in `clientId`, returns `false`, and the user gets `ErrDealRejectedByMarket` with no context about the actual bug (malformed client).

Propagate the parse error instead.

market/mk20/ddo_v1.go
```diff
@@ -98,88 +97,154 @@ func (d *DDOV1) Validate(db *harmonydb.DB, cfg *config.MK20Config) (DealCode, er
 		}
 	}
 
-	if d.ContractAddress == "" {
-		return ErrProductValidationFailed, xerrors.Errorf("contract address is not set")
+	if d.MarketAddress != "" {
+		if len(d.MarketAddress) < 2 {
+			return ErrProductValidationFailed, xerrors.Errorf("market address too short")
+		}
+		if d.MarketAddress[0:2] != "0x" {
```
Reiers, line 104: The `len < 2` and `[0:2] != "0x"` checks are redundant. `common.IsHexAddress` below handles `0x`-prefix + 40-char hex + case-insensitive correctly.

```go
if d.MarketAddress != "" {
    if !common.IsHexAddress(d.MarketAddress) {
        return ErrProductValidationFailed, xerrors.Errorf("invalid market address")
    }
}
```

market/mk20/db_utils.go
```diff
@@ -0,0 +1,277 @@
...
+	if err != nil {
+		return xerrors.Errorf("update deal: %w", err)
+	}
+	if n != 1 {
+		return xerrors.Errorf("update deal: expected 1 row affected, got %d", n)
+	}
+	return nil
+}
+
+func (d *Deal) UpdateDeal(tx *harmonydb.Tx) error {
```
Reiers, line 155: **Dead code.** `UpdateDealWithTx` (lines 124-148) and this function (`UpdateDeal`, 149-177) are byte-for-byte identical. `UpdateDealWithTx` has no callers in the repo. Suggest removing it.

market/mk20/http/http.go
```diff
@@ -117,20 +118,22 @@ func Router(mdh *MK20DealHandler, domainName string) http.Handler {
 // @securityDefinitions.apikey CurioAuth
 // @in header
 // @name Authorization
-// @description Use the format: `CurioAuth PublicKeyType:PublicKey:Signature`
+// @description Use the format: `CurioAuth KeyType:AddressBytes:Signature`
```
Reiers, line 117: `dealRateLimitMiddleware()` is also applied on `APIRouter` at line 136. All API requests go through both, effectively halving the allowed rate on `/deal`, `/status`, `/uploads`, etc.

Also: 50/sec/IP global includes `PUT /uploads/{id}/{chunkNum}` which is the parallel chunk upload endpoint. Legitimate large-aggregate uploads from one IP will hit this immediately. Suggest per-route limits.

market/mk20/http/http.go
```diff
@@ -69,10 +70,10 @@ func AuthMiddleware(db *harmonydb.DB, cfg *config.CurioConfig) func(http.Handler
...
 			}
 
-			allowed, client, err := mk20.Auth(authHeader, db, cfg)
+			allowed, client, err := mk20.Auth(authHeader, r.Method, r.URL.EscapedPath(), db, cfg)
 			if err != nil {
 				log.Errorw("failed to authenticate request", "err", err)
-				http.Error(w, "Error during authentication", http.StatusInternalServerError)
+				http.Error(w, "Unauthorized", http.StatusUnauthorized)
 				return
 			}
```
Reiers, line 75: All auth failures return the same generic `Unauthorized` with no log of which check failed (missing header / bad signature / client not in allowlist / client not authorized for this deal id).

Painful to debug in production. Suggest `log.Infow` with structured rejection reason, keeping the response body generic.

Dockerfile
```diff
@@ -62,14 +62,8 @@ RUN --mount=type=cache,target=/go/pkg/mod \
...
 
-#RUN go install github.com/ipld/go-car/cmd/car@latest \
-# && cp $GOPATH/bin/car /usr/local/bin/
-RUN git clone https://github.com/ipld/go-car.git /tmp/go-car && \
-    cd /tmp/go-car && \
-    git checkout v2.16.0 && \
-    cd cmd/car && go build -o car && \
-    chmod +x car && \
-    cp car /usr/local/bin/
+RUN go install github.com/ipld/go-car/cmd/car@latest \
```
Reiers, line 65: **Un-pinning `go-car` from `v2.16.0` → `@latest`.**

The pin was introduced in #1106 during the Go 1.25→1.26 transition. No commit message explanation here for why it's being reverted.

Concerns: reproducibility (each build fetches a different binary), supply chain (breaking or malicious upstream releases), and an implicit assumption that current `@latest` still builds on Go 1.26.

If the pin is no longer needed, please document. Otherwise keep pinned and bump explicitly.

market/backpressure/backpressure.go
```diff
@@ -0,0 +1,322 @@
...
+	if err == nil {
+		return pressure.(bool), nil
+	}
+	p, err := c.checkMK20Backpressure(ctx, cfg, db)
+	if err != nil {
+		return false, err
+	}
+	_ = c.cache.SetWithTTL(mk20BackpressureKey, p, time.Minute*2)
+	return p, nil
+}
```
Reiers, line 322: The PR description marks the backpressure TODO complete, but the existing `TODO: add aggregation check, upload queue, download queue and offline queue` at the end of `checkMK20Backpressure` is still open.

Actual state:
- ✅ running pipelines, downloading pending, commp pending
- ❌ aggregation pending — `aggPending` is computed in the SQL but never thresholded
- ❌ upload queue
- ❌ offline queue

A client can flood `/uploads/*` without any gate. Either complete the thresholds here or explicitly note in the PR description which pieces are deferred.

Reiers:
**Self-correction on finding #3:** I went and ran the grep myself. The only Go caller of `process_offline_download` is `tasks/storage-market/mk20.go:688`, and it already passes all 6 arguments matching the new signature:

```go
tx.QueryRow(`SELECT process_offline_download($1, $2, $3, $4, $5, $6)`,
    piece.ID, piece.PieceCIDV2, piece.PieceCID, piece.PieceSize, piece.RawSize, mk20.ProductNameDDOV1)
```

So finding #3 is a non-issue — the caller was updated alongside the signature change. Striking it from the blocking list. Forward migration correctly `DROP`s the old 5-arg signature first, so that path is clean.

Findings #1 (broken downgrade), #2 (Content-Type ordering), and #4 (cross-PR ON CONFLICT break with #1163) remain blocking as written.

~ @Reiers

Reiers:
Sharpening finding #4 (cross-PR ON CONFLICT break with #1163).

Looked at the PR's own SQL migration: the new `process_offline_download` function body already uses the **new** partial-index form:

```sql
ON CONFLICT (piece_cid, piece_padded_size, long_term)
    WHERE cleanup_task_id IS NULL
DO UPDATE ...
```

So the SQL function is ready for the post-#1163 world. But the Go-embedded SQL strings in the same PR still use the old 4-column form at:

- `market/mk20/mk20.go:752, 852`
- `market/mk20/mk20_upload.go:361, 767`
- `tasks/storage-market/mk20.go:380, 502`

The transition to the partial index was partially applied — the new SQL function got the right syntax, but these six Go call sites were missed.

**Merge order matters because of Postgres ON CONFLICT inference rules** (inference requires an exact match of `(cols)` + `WHERE` predicate to an existing unique index/constraint):

- Before #1163: the unique constraint on `(piece_cid, piece_padded_size, long_term, cleanup_task_id)` (4 cols, no predicate) is the only candidate. Only `ON CONFLICT (piece_cid, piece_padded_size, long_term, cleanup_task_id)` matches.
- After #1163: the partial index `parked_pieces_active_piece_key` on `(piece_cid, piece_padded_size, long_term) WHERE cleanup_task_id IS NULL` is the only match. Only the new-syntax clause works.
- The two forms are **mutually exclusive** under PG inference.

**So three possible outcomes:**
1. #1163 merges first, then #1087: #1087 breaks at runtime because the 4-col constraint no longer exists. Must update the six Go sites before merging.
2. #1087 merges first, then #1163: #1163's Go-side refactor already updates its own files; it needs to be extended to also patch the six Go sites introduced by #1087.
3. Coordinated merge: whichever PR goes second applies a same-commit update to the six Go sites to flip syntax from old → new.

Cleanest is option 3. Whoever merges second owns the six-site update and runs the full itest suite. ~ @Reiers

Reiers:
Two more findings from a pass on `web/api/webrpc/market_2.go` and the settings UI:

### 14. `AddMarketContract` does not canonicalize the address, still has open TODO

`web/api/webrpc/market_2.go:553-575`:
- Line 569: `// TODO: Verify that contract address exists on chain and if it supports CurioABI` is still open. Operators can add garbage / non-existent / wrong-interface addresses, and every deal routed to them fails with an opaque error.
- Line 572: `contract` is inserted as-is with no lowercasing. **This concretely reproduces finding #10** — an operator adds `0xAbC...` via the UI, a deal arrives with `0xabc...`, the `lower()`-less lookup in `ddo_v1.go:144` returns `ErrNoRows` → `ErrUnknowContract`. The two ends disagree on casing.
- Fix at the webRPC layer: lowercase on insert + lowercase on lookup in `ddo_v1.go:144`. Or add a checksum-canonicalization step.
- Minimal validation could also call `market.Version(&bind.CallOpts{Context: ctx})` on the contract at add time — if it reverts or returns != 1, reject the add.

The UI also passes `allowed=true` on add (`settings.mjs:addContract`), overriding the SQL `DEFAULT FALSE`. Probably intentional but worth a one-line note in the UI ("Adding a contract enables it immediately").

### 15. WebRPC surface has no authentication (pre-existing, but worth flagging)

`AddMarketContract` / `UpdateMarketContract` / `RemoveMarketContract` are all unauthenticated. Anyone who can reach the Curio web port can manipulate the DDO contract allowlist.

This is a pre-existing property of Curio's web UI, **not introduced by this PR**. But this PR makes the surface security-relevant by turning `ddo_contracts` into a policy table that gates real deals. Suggest at least an inline comment or docs call-out that the web UI must be on a trusted network.

~ @Reiers

snadrus:
Cursor review has some interesting finds: 

### Needs doc?:

getDealState is not used by Curio
The Markdown describes the full ICurioDealViewV1 surface, including getDealState, and tells providers to reason about Open / Active / Finalized. There are no non-generated call sites of GetDealState in the repo (only market/mk20/contract/CurioDealViewV1.go). So the docs are accurate as an interface contract for integrators, but Curio does not perform on-chain deal-state housekeeping via getDealState today. If readers infer from the Solidity comment (“used by Curio while sealing”) that sealing polls the contract, that is not implemented. The Markdown is softer (“for deal-state queries”) but could explicitly say intake uses verifyDeal / version only, unless you plan to wire getDealState later.

### Probably worth fixing: 

Allowlist key is a raw string match
VerifyMarketDeal looks up ddo_contracts with d.MarketAddress exactly as submitted. AddMarketContract stores mAddr.String() (canonical hex from go-ethereum). Mixed-case or all-lower client payloads could fail the lookup even for the same contract. Worth normalizing (e.g. always common.HexToAddress(...).Hex()) on both insert and lookup for safety aligned with “address must exist in ddo_contracts”.

### Makes debugging more difficult:

HTTP semantics for “market missing” vs “rejected”
Both missing deal (DealNotFound) and verifyDeal == false map to ErrDealRejectedByMarket (426). The doc’s “market-missing” is narrative, not a separate status code—fine, but operators should know it is the same bucket as other market rejections.

snadrus (approved):

LexLuthr:
> Cursor review has some interesting finds:
> 
> ### Needs doc?:
> getDealState is not used by Curio The Markdown describes the full ICurioDealViewV1 surface, including getDealState, and tells providers to reason about Open / Active / Finalized. There are no non-generated call sites of GetDealState in the repo (only market/mk20/contract/CurioDealViewV1.go). So the docs are accurate as an interface contract for integrators, but Curio does not perform on-chain deal-state housekeeping via getDealState today. If readers infer from the Solidity comment (“used by Curio while sealing”) that sealing polls the contract, that is not implemented. The Markdown is softer (“for deal-state queries”) but could explicitly say intake uses verifyDeal / version only, unless you plan to wire getDealState later.
> 
> ### Probably worth fixing:
> Allowlist key is a raw string match VerifyMarketDeal looks up ddo_contracts with d.MarketAddress exactly as submitted. AddMarketContract stores mAddr.String() (canonical hex from go-ethereum). Mixed-case or all-lower client payloads could fail the lookup even for the same contract. Worth normalizing (e.g. always common.HexToAddress(...).Hex()) on both insert and lookup for safety aligned with “address must exist in ddo_contracts”.
> 
> ### Makes debugging more difficult:
> HTTP semantics for “market missing” vs “rejected” Both missing deal (DealNotFound) and verifyDeal == false map to ErrDealRejectedByMarket (426). The doc’s “market-missing” is narrative, not a separate status code—fine, but operators should know it is the same bucket as other market rejections.

All AIs seem to be stuck in the same loop. They can't see beyond code or understand intent.

PR #1094, opened 2026-03-13
itests should try pg / scylla

snadrus:
Have tests run against Postgres and (where applicable for indexstore) Scylla. 
Durability is unimportant in test running, but speed is. 
Also, instances are shared (not possible before with YB low limits): This enables considerable memory & start-time savings.

Test Grouping
- Some small tests were combined.
- itests got restructured so it's *impossible* not to run: simply add another /itests/itestgroup$N/ folder for various $N and it'll become a new job. This will keep CI from having edits every new test. 

Results:
CI resource reductions: most test jobs run 1-2 minutes faster. 
Should reduce the CI load.

### DIFF
--- a/documentation/en/developer/testing.md
+++ b/documentation/en/developer/testing.md
@@ -14,7 +14,10 @@ make test-dbs-up
 # 2. Build FFI dependencies (needed once, or after submodule updates)
 FFI_USE_OPENCL=1 make deps
 
-# 3. Run tests (make test-dbs-up prints this invocation for copy-paste)
+# 3. Run all integration tests (DB env vars are set by the Makefile)
+make test
+
+# Or run one shard / one test (same env as make test; tags may differ from CI)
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -700,6 +719,16 @@ jobs:
           echo "" >> $GITHUB_STEP_SUMMARY
           echo "### 📊 [View Coverage Report](https://artifact.ci/${{ github.repository }}/actions/runs/${{ github.run_id }}/coverage-merged/coverage.html)" >> $GITHUB_STEP_SUMMARY
 
+  stop-databases:
+    runs-on: [self-hosted, docker]
+    needs: [test]
+    if: always() && needs.test.result != 'cancelled'
```
rvagg: why exclude cancelled? shouldn't we just always stop?

PR #1136, opened 2026-04-01
Debug/chain use

snadrus:
When Curio is started with: CURIO_CHAIN_RPC_STATS=1
Every 10 seconds the full set of callsites calling Lotus will be logged in an aggregated way. 

This PR changes only logging to understand why main uses more chain than pdpv0 branch. 
Not intended for production.


LexLuthr:
Can we close this and delete the branch?

LexLuthr:
I am closing it as all SP have moved to `main`

PR #1146, opened 2026-04-07
Feat/harmonytask oldest first

snadrus:
harmonytask "pollerTryAllTasks" order:

was:
- user-given and should balance time-sensitive & pipeline end-first

new:
- prefer the oldest task, BUT if it's got a pipeline after it, do those first.

Effect:
- This allows multiple pipelines (PoREP, PDP) to have equal worth
- It directly solves "age starvation" in the case where there's enough resources to start the task & they aren't taken by later ones.

Note: "too big" tasks remain in danger of resource starvation by younger, smaller tasks. Note, this risk is now reduced and only applies when the SP is considerably overwhelmed with work, or especially-poorly laid-out (asking GPU machines to do piddly non-GPU work). Fixing it enters the realm of NP-complete challenges, so heuristics are as good as it can be fixed.


tasks/pdp/task_prove.go
```diff
@@ -538,7 +538,8 @@ func (p *ProveTask) CanAccept(ids []harmonytask.TaskID, engine *harmonytask.Task
 
 func (p *ProveTask) TypeDetails() harmonytask.TaskTypeDetails {
 	return harmonytask.TaskTypeDetails{
-		Name: "PDPProve",
+		Name:      "PDPProve",
+		MayFollow: []string{"PDPInitPP", "AggregatePDPDeal", "PDPProvingPeriod"},
```
LexLuthr: We probably should not have AggregatePDPDeal here. We should keep 2 pipelines with PDP as well. 1 is for data onboarding and other is proving. They should each have a priority/follow graph. Proving over all wins of onboarding.

tasks/pdp/task_next_pp.go
```diff
@@ -224,7 +224,8 @@ func (n *NextProvingPeriodTask) CanAccept(ids []harmonytask.TaskID, engine *harm
 
 func (n *NextProvingPeriodTask) TypeDetails() harmonytask.TaskTypeDetails {
 	return harmonytask.TaskTypeDetails{
-		Name: "PDPProvingPeriod",
+		Name:      "PDPProvingPeriod",
+		MayFollow: []string{"PDPProve"},
```
LexLuthr, line 228: Wouldn't this create a cyclic chain in  `FindPreferredTaskRunOrder` with PDPProve?

harmony/harmonytask/scheduler.go
LexLuthr, line 1: pollAllTaskTypes() does not load posted time from DB

LexLuthr, line 1: EmitTaskNew also loses postedTime when we emit NewTask event

harmony/harmonytask/treehelper/treehelper_test.go
LexLuthr, line 1: Can we put a test to catch cyclic order? I think it is very easy to mess that up.

LexLuthr (commented):
There are few places where PostedTime is not being done correctly and I am worried about cyclic dependency of follow.

tasks/pdp/task_prove.go
snadrus: Pipeline 1 (onboarding): PDPAddDataSet → PDPCommP → … → AggregatePDPDeal.
Pipeline 2 (proving): AggregatePDPDeal → PDPInitPP → PDPProve → PDPProvingPeriod (with PDPSync still following PDPProve as before).

Same for Pdpv0 now

tasks/pdp/task_next_pp.go
snadrus: fixed

harmony/harmonytask/treehelper/treehelper_test.go
snadrus: 2 tests:
1. Ensure harmonytask.Registry is comprehensive.
2. Ensure all registry tasks' MayFollow DAGs are acyclic.

harmony/harmonytask/scheduler.go
snadrus: fixed. it hints to CanAccept the preferred order.

snadrus: In my mind, new means new: the postedDate is now(). Anything more precise would suffer clock-drift between machines. WDYT?

LexLuthr: We should probably make NTP or chrony a dep. Problem with now() is that important task like wdPost go back to queue as new. So imagine if we have 2 tasks in the same time period, now we don't have a deterministic way to say which one get's priority. Even if older one should be retried immediately. This basically breaks the "respect the posted time" after 1 failure.

harmony/harmonytask/scheduler.go
```diff
@@ -439,14 +448,15 @@ func (e *TaskEngine) pollAllTaskTypes() map[string][]task {
...
+		PostedTime time.Time `db:"posted_time"`
 		Retries    int       `db:"retries"`
 	}
 	names := make([]string, len(e.handlers))
 	for i, h := range e.handlers {
 		names[i] = h.Name
 	}
 	err := e.db.Select(context.Background(), &rows,
-		`SELECT id, name, update_time, retries FROM harmony_task WHERE owner_id IS NULL AND name = ANY($1)`, names)
+		`SELECT id, name, update_time, posted_time, retries FROM harmony_task WHERE owner_id IS NULL AND name = ANY($1)`, names)
```
LexLuthr, line 459: Let's add order by posted_time here. Saves us sorting compute later.

harmony/harmonytask/scheduler.go
```diff
@@ -146,6 +146,11 @@ func (e *TaskEngine) startScheduler() {
 						if len(tasks) == 0 {
 							continue
 						}
+
+						sort.Slice(tasks, func(i, j int) bool {
```
LexLuthr, line 150: We should probably use below for this to match canAccept() with peer ordering. 

```
// taskLessByPostedTime defines FIFO order for the same task type: older posted_time first; unknown
// posted_time (zero) is treated as newest so DB-backed ordering wins once polled.
func taskLessByPostedTime(a, b task) bool {
	aUnk := a.PostedTime.IsZero()
	bUnk := b.PostedTime.IsZero()
	switch {
	case aUnk && bUnk:
		return a.ID < b.ID
	case aUnk:
		return false
	case bUnk:
		return true
	case a.PostedTime.Equal(b.PostedTime):
		return a.ID < b.ID
	default:
		return a.PostedTime.Before(b.PostedTime)
	}
}
```

LexLuthr (approved):
Looks good now. Couple of nits.

harmony/harmonytask/scheduler.go
snadrus: It's less compute to group into buckets before sorting each bucket. And we are free to offload this sort away from the database (unlike sort-with-limit).

PR #1151, opened 2026-04-09
chore: update frisbii, use fast-path Has() for head requests (main)

### LINKED ISSUE #732: Retrieval: avoid full block loads for HEAD /ipfs/ requests
Background [here](https://filecoinproject.slack.com/archives/C06GD1SS56Y/p1761238820280559).

A `HEAD /ipfs/bafy...` shouldn't trigger a full block load from the blockstore. It should effectively be a `Has()` operation. We need a way to plumb this through the linksystem layer in between so that we don't waste byte movements for something we're not going to serve.

rvagg:
Closes: https://github.com/filecoin-project/curio/issues/732

NOTE: This upgrades to Go 1.25, unfortunately, but the core ipfs tooling is being very eager to upgrade their Go versions. I don't think this one's a big deal though and there's a couple of lint fixes in here to deal with it.


LexLuthr (commented):
This will break CI and a bunch of other things as Go is only upgraded here. I am about to do a PR for Go upgrade today. Let's rebase this on `main` after that and it should be fine. Please do not merge before that

LexLuthr:
https://github.com/filecoin-project/curio/pull/1159 supersedes this. Closing it.

PR #1159, opened 2026-04-14
Switch to Postgres template itests

LexLuthr:


harmony/harmonydb/sql/20251015-pdp-v0-piece-adds-datasetid-nullable.sql
```diff
@@ -1,7 +1,7 @@
 -- Changes the data_set column to be nullable in the pdp_data_set_piece_adds table to faciliate create-and-add workflow.
 -- Combined migration: make `data_set` nullable and adjust PK
--- New primary key: (add_message_hash HASH, add_message_index ASC)
--- Old primary key: (data_set HASH, add_message_hash ASC, add_message_index ASC)
+-- YugabyteDB primary key: (add_message_hash HASH, add_message_index ASC)
```
snadrus: The hash adds nothing. Let's just use the pg way

LexLuthr: We can't change existing Schema if you are saying to just dump the YB format. This just allows us to create new key PG in manner for testing.

snadrus: We can add a blanket conversion away from hash in a new SQL. We can then change the old ones. Already-upgraded people will get the unification away from hash, and tests (and luddites) get the code that sets it right already. 

Old SQL:
  -hash
  +asc
  
New "fix it" SQL:
   if HASH, set to ASC
   
This would simplify the hard-to-test upgrader.

LexLuthr: Fixed

PR #1163, opened 2026-04-17
fix: make parked_piece entries truly unique

LexLuthr:
# Summary

Fix duplicate active rows in parked_pieces and safely enforce the intended uniqueness rule.

## Problem

A previous uniqueness constraint allowed multiple active rows for the same piece key due to NULL behavior on cleanup_task_id.

```
yugabyte=# SELECT
  NULL = NULL           AS eq_null,
  NULL <> NULL          AS neq_null,
  NULL IS NULL          AS is_null,
  NULL IS NOT DISTINCT FROM NULL AS same_null;
 eq_null | neq_null | is_null | same_null 
---------+----------+---------+-----------
         |          | t       | t
(1 row)
```

This caused duplicate active records and blocked creation of the correct active-row unique index.

## Solution

1. Update migration logic to create the index only when duplicates are absent.
2. Add a Go repair job to clean duplicate active groups on live systems.
3. Select a keeper row that still has the backing file on disk.
4. Reattach refs to the keeper row.
5. Retire duplicate rows through normal cleanup flow.
6. Create the unique partial index after cleanup completes.

## Result

Only one active row can exist per logical piece key, while preserving file safety and live-system compatibility.

### DIFF
--- a/cmd/curio/tasks/tasks.go
+++ b/cmd/curio/tasks/tasks.go
@@ -246,7 +246,8 @@ func StartTasks(ctx context.Context, dependencies *deps.Deps, shutdownChan chan 
 			}
 			cleanupPieceTask := piece2.NewCleanupPieceTask(db, must.One(slrLazy.Val()), 0)
 			aggregateChunksTask := piece2.NewAggregateChunksTask(db, stor, must.One(slrLazy.Val()))
-			activeTasks = append(activeTasks, parkPieceTask, cleanupPieceTask, aggregateChunksTask)
+			pfix := piece2.NewFixParkPieceTask(db, must.One(slrLazy.Val()))
+			activeTasks = append(activeTasks, parkPieceTask, cleanupPieceTask, aggregateChunksTask, pfix)
 		}
 	}
 
--- a/harmony/harmonydb/sql/20260410-ipni-head-cas.sql
+++ b/harmony/harmonydb/sql/20260410-ipni-head-cas.sql
@@ -185,9 +185,27 @@ BEGIN
 END;
 $$;
 
-CREATE UNIQUE INDEX IF NOT EXISTS parked_pieces_active_piece_key
-    ON parked_pieces (piece_cid, piece_padded_size, long_term)
-    WHERE cleanup_task_id IS NULL;
+-- Create a unique index for the parked_pieces table if duplicates do not exist.
+-- This is to avoid duplicate entries in the parked_pieces table.
+-- This is a no-op if the index already exists.
+-- If duplicates exist, then a Go task will remove them correctly and create the index
+DO $$
+BEGIN
+  IF NOT EXISTS (
+    SELECT 1
+    FROM (
+      SELECT piece_cid, piece_padded_size, long_term
+      FROM parked_pieces
+      WHERE cleanup_task_id IS NULL
+      GROUP BY 1,2,3
+      HAVING count(*) > 1
+    ) dupes
+  ) THEN
+    CREATE UNIQUE INDEX IF NOT EXISTS parked_pieces_active_piece_key
+      ON parked_pieces (piece_cid, piece_padded_size, long_term)
+      WHERE cleanup_task_id IS NULL;
+  END IF;
+END $$;
 
 -- This function triggers a download for an offline piece.
 -- It is different from MK1.2 PoRep pipeline as it downloads the offline pieces
--- a/pdp/handlers_upload.go
+++ b/pdp/handlers_upload.go
@@ -301,7 +301,8 @@ func (p *PDPService) handlePieceUpload(w http.ResponseWriter, r *http.Request) {
 		var parkedPieceID int64
 		err := tx.QueryRow(`
             INSERT INTO parked_pieces (piece_cid, piece_padded_size, piece_raw_size, long_term)
-            VALUES ($1, $2, $3, TRUE) RETURNING id
+            VALUES ($1, $2, $3, TRUE) ON CONFLICT (piece_cid, piece_padded_size, long_term, cleanup_task_id) DO NOTHING
+			RETURNING id
         `, pieceCIDComputed.String(), paddedPieceSize, readSize).Scan(&parkedPieceID)
 		if err != nil {
 			return false, fmt.Errorf("failed to create parked_pieces entry: %w", err)
@@ -525,7 +526,8 @@ func (p *PDPService) handleStreamingUpload(w http.ResponseWriter, r *http.Reques
 		var parkedPieceID int64
 		err := tx.QueryRow(`
             INSERT INTO parked_pieces (piece_cid, piece_padded_size, piece_raw_size, long_term)
-            VALUES ($1, $2, $3, TRUE) RETURNING id
+            VALUES ($1, $2, $3, TRUE) ON CONFLICT (piece_cid, piece_padded_size, long_term, cleanup_task_id) DO NOTHING
+			RETURNING id
         `, pcid.String(), paddedPieceSize, readSize).Scan(&parkedPieceID)
 		if err != nil {
 			return false, fmt.Errorf("failed to create parked_pieces entry: %w", err)
--- a/pdp/handlers_upload.go
+++ b/pdp/handlers_upload.go
@@ -300,10 +300,24 @@ func (p *PDPService) handlePieceUpload(w http.ResponseWriter, r *http.Request) {
 		// 1. Create a long-term parked piece entry
 		var parkedPieceID int64
 		err := tx.QueryRow(`
-            INSERT INTO parked_pieces (piece_cid, piece_padded_size, piece_raw_size, long_term)
-            VALUES ($1, $2, $3, TRUE) ON CONFLICT (piece_cid, piece_padded_size, long_term, cleanup_task_id) DO NOTHING
-			RETURNING id
-        `, pieceCIDComputed.String(), paddedPieceSize, readSize).Scan(&parkedPieceID)
+							WITH existing_piece AS (
+							  SELECT id
+							  FROM parked_pieces
+							  WHERE piece_cid = $1
+								AND piece_padded_size = $2
+								AND long_term = TRUE
+								AND cleanup_task_id IS NULL
+							),
+							inserted_piece AS (
+							  INSERT INTO parked_pieces (piece_cid, piece_padded_size, piece_raw_size, long_term)
+							  SELECT $1, $2, $3, TRUE
+							  WHERE NOT EXISTS (SELECT 1 FROM existing_piece)
+							  RETURNING id
+							)
+							SELECT id FROM existing_piece
+							UNION ALL
+							SELECT id FROM inserted_piece
+							LIMIT 1;`, pieceCIDComputed.String(), paddedPieceSize, readSize).Scan(&parkedPieceID)
 		if err != nil {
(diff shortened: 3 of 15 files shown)


tasks/piece/task_fix_park_piece_unique.go
```diff
@@ -0,0 +1,200 @@
...
+		_, err = f.db.Exec(ctx, `CREATE UNIQUE INDEX parked_pieces_active_piece_key ON parked_pieces (piece_cid, piece_padded_size, long_term) WHERE cleanup_task_id IS NULL;`)
+		if err != nil {
+			if strings.Contains(err.Error(), "SQLSTATE 23505") {
+				// Let's fail this so we can retry few seconds later, which is likely to succeed
+				return false, xerrors.Errorf("duplicate rows exist while creating index: %w", err)
+			}
+			return false, xerrors.Errorf("creating index: %w", err)
+		}
+		return true, nil
+	}
```
snadrus, line 81: Deleting this has the same result, correct? The for loop has nothing to do and the same code is below

tasks/piece/task_fix_park_piece_unique.go
```diff
@@ -0,0 +1,200 @@
...
+												  AND complete = true
+												GROUP BY 1,2,3
+												HAVING count(*) > 1;`)
+
+		if err != nil {
+			return false, xerrors.Errorf("querying parked piece: %w", err)
+		}
+
+		if len(reduplicates) > 0 {
+			return false, nil
```
snadrus, line 138: This should go back to the top (line 50 or so) and fix duplicates again, right?

tasks/piece/task_fix_park_piece_unique.go
```diff
@@ -0,0 +1,200 @@
...
+		PieceCid  string  `db:"piece_cid"`
+		PieceSize int64   `db:"piece_padded_size"`
+		LongTerm  bool    `db:"long_term"`
+		RowCount  int64   `db:"row_count"`
+		IDs       []int64 `db:"ids"`
+	}
+
+	var duplicates []duplicate
+
+	err = f.db.Select(ctx, &duplicates, `SELECT piece_cid, piece_padded_size, long_term,
```
rvagg, line 82: this looks like a for{} loop to me rather than trying to do both things twice: for{ any duplicates? yes = deal with them & continue; no = make index and break }.

or if it's just an iambored then run it once and wait for next time and maybe it'll be fixed, but I think jus doing it in a hungry loop to squash them down would be the rigth approach no?

LexLuthr: IAmBored. Run once, try to fix as many as you can. If all are fixed then create index and finish. If could not complete for any reason, retry. If all retry exhausted, we will try again later.

Quick retry is to work around transient DB issues. We can remove the task maybe a few months when we know this is fixed on all SP.

Once index is created then next run will simply check that and mark finished.

cmd/curio/tasks/tasks.go
```diff
@@ -246,7 +246,8 @@ func StartTasks(ctx context.Context, dependencies *deps.Deps, shutdownChan chan
 			}
 			cleanupPieceTask := piece2.NewCleanupPieceTask(db, must.One(slrLazy.Val()), 0)
 			aggregateChunksTask := piece2.NewAggregateChunksTask(db, stor, must.One(slrLazy.Val()))
-			activeTasks = append(activeTasks, parkPieceTask, cleanupPieceTask, aggregateChunksTask)
+			pfix := piece2.NewFixParkPieceTask(db, must.One(slrLazy.Val()))
```
ZenGround0, line 249: What is the general curio deprecation plan for such fixup tasks?  I propose we remove this task after some period of time or some number of releases.  Arbitrarily I'd say remove this after 1 or 2 major releases.  For PDP users that seems sufficient. I could use your help understanding what the typical release adoption pattern is for curio PoRep.

harmony/harmonydb/sql/20260410-ipni-head-cas.sql
```diff
@@ -185,9 +185,27 @@ BEGIN
 END;
 $$;
 
-CREATE UNIQUE INDEX IF NOT EXISTS parked_pieces_active_piece_key
-    ON parked_pieces (piece_cid, piece_padded_size, long_term)
-    WHERE cleanup_task_id IS NULL;
+-- Create a unique index for the parked_pieces table if duplicates do not exist.
+-- This is to avoid duplicate entries in the parked_pieces table.
+-- This is a no-op if the index already exists.
+-- If duplicates exist, then a Go task will remove them correctly and create the index
```
ZenGround0, line 191: Do we want the redundancy of two index creating paths in this PR when the go task can always be trusted to create this index?   

I'm imagining a different approach where we rely on the temporary task exclusively for now and then after a few releases we forgo the uniqueness check on parked pieces and just assume it, i.e. removing the "IF NOT EXISTS" clause here.

tasks/piece/task_fix_park_piece_unique.go
```diff
@@ -0,0 +1,138 @@
...
+					// If piece does not exist then we will park it otherwise fail here
+					if !errors.Is(err, storiface.ErrSectorNotFound) {
+						continue
+					}
+				}
+				defer func() {
+					_ = pr.Close()
+				}()
+
+				_, err = f.db.Exec(ctx, `UPDATE parked_piece_refs SET piece_id = $1 WHERE piece_id = ANY($2::bigint[]);`, id)
```
ZenGround0: I see how this is resolving the `parked_piece_ref` problem of referencing duplicate content addressed `parked_pieces` entries but I don't see us actually removing the duplicates from `parked_pieces`.  It seems like we'd need to do that to successfully create the unique index on `parked_pieces` and make progress.  Otherwise it seems like this task will keep doing all of the deduplication work all the time without realizing it has made progress.

Is this correct?  If not what am I missing?

tasks/piece/task_fix_park_piece_unique.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	for _, dup := range duplicates {
+		if stillOwned() {
+			for _, id := range dup.IDs {
+				// Verify that the piece is still parked
+				// If piece exists then check if we can access the data
+				pr, err := f.sc.PieceReader(ctx, storiface.PieceNumber(id))
+				if err != nil {
+					// If piece does not exist then we will park it otherwise fail here
```
ZenGround0: This comment is not reflecting the code since our actual behavior is 
1. If `ErrSectorNotFound` we continue trying to deduplicate all `parked_piece_refs` to this piece
2. Otherwise we don't attempt to process this ID and try the next one

Either comment or behavior need to change

ZenGround0: The second behavior makes sense to me but I don't understand why we want to dedup piece refs over to a missing piece even if there are possibly non-missing pieces later on in the ids list.  What is the reasoning for that?

ZenGround0 (commented):
Not quite ready to approve but potentially just because I am confused about some basic things.

cmd/curio/tasks/tasks.go
LexLuthr: I do want to remove these tasks later. Preferably after 2 releases or maybe a little more(I am still trying to understand the impact on PoRep users). I also want to change how these tasks are scheduled in the first place. Once new harmony changes land, it would be best of create a new task type which can safely be marked "off" after completing it's original intent so new tasks are not created for it.

harmony/harmonydb/sql/20260410-ipni-head-cas.sql
LexLuthr: SQL schema changes should never ever be changed in Go side. This is the SQL which will anchor the schema change once Go code is gone. I am breaking the Curio's cardinal rule here because I could not find a reliable way to do this purely on SQL side.

tasks/piece/task_fix_park_piece_unique.go
LexLuthr: Nope. I messed that up.

tasks/piece/task_fix_park_piece_unique.go
LexLuthr: This was again my mistake. I have done too many iteration of this and did a poor job cleaning it up.

LexLuthr:
I changed the inserts to CTEs as a rollout bridge: before the new partial unique index exists, the new partial ON CONFLICT target is not valid, but after the index exists the old 4-column conflict target will not catch conflicts from the new index. The CTE first selects an existing active parked_pieces row and only inserts when none exists, returning an id in both cases, so the code works across both schema states until a later PR can simplify it to the new partial ON CONFLICT.

A follow up PR which removes this task should also replace these complex CTEs with ON CONFLICT() correctly post index creation.

snadrus:
This we cannot remove this task until we pass 2 network boundaries to ensure everyone runs it. We should notate that network version.

pdp/handlers_upload.go
```diff
@@ -300,10 +300,24 @@ func (p *PDPService) handlePieceUpload(w http.ResponseWriter, r *http.Request) {
 		// 1. Create a long-term parked piece entry
 		var parkedPieceID int64
 		err := tx.QueryRow(`
-            INSERT INTO parked_pieces (piece_cid, piece_padded_size, piece_raw_size, long_term)
-            VALUES ($1, $2, $3, TRUE) ON CONFLICT (piece_cid, piece_padded_size, long_term, cleanup_task_id) DO NOTHING
-			RETURNING id
-        `, pieceCIDComputed.String(), paddedPieceSize, readSize).Scan(&parkedPieceID)
+							WITH existing_piece AS (
```
ZenGround0, line 303: Since this query is super verbose and repeatedly used you should extract it out somewhere and share it.

tasks/piece/task_fix_park_piece_unique.go
```diff
@@ -0,0 +1,151 @@
...
+			for _, id := range dup.IDs {
+				// Verify that the piece is still parked
+				// If piece exists then check if we can access the data
+				pr, err := f.sc.PieceReader(ctx, storiface.PieceNumber(id))
+				if err != nil {
+					// If piece does not exist then we check the next one
+					continue
+				}
+				_ = pr.Close()
+				_, err = f.db.Exec(ctx, `UPDATE parked_piece_refs SET piece_id = $1 WHERE piece_id = ANY($2::bigint[]);`, id, dup.IDs)
```
ZenGround0, line 106: I might be missing it but this still seems to have the problem that `parked_pieces` table is not cleaning up all duplicates.

ZenGround0: Yeah I was missing the fundamental thing where we have `task_cleanup_piece` checking for parked piece entries with no refs and doing cleanup automatically.

ZenGround0 (approved):

LexLuthr:
> This we cannot remove this task until we pass 2 network boundaries to ensure everyone runs it. We should notate that network version.

@snadrus Added a TODO in the task file with NV30 target and CTE cleanup.

snadrus (approved):
Is there any test that can be written? 
This code is quite unusual (though I see no direct problem in it).

beck-8:
https://filecoinproject.slack.com/archives/C06GD1SS56Y/p1776951197970209?thread_ts=1776412973.488759&cid=C06GD1SS56Y

Verified after several rounds of testing. No issues seen so far.

PR #1165, opened 2026-04-20
Safter HarmonyTask

snadrus:
With all the new developments, there are varying rules to using many of the new variables. 
This PR introduces helpers that contain those special variables so that their usage is contained in accessors or labelled groups.
No functionality is changed in this PR. 

This includes [feat/harmonytask_oldest_first](https://github.com/filecoin-project/curio/pull/1146) and that PR should get merged first to make this easy to review.


lib/cachedreader/cachedreader.go
```diff
@@ -265,13 +265,16 @@ func (cpr *CachedPieceReader) getPieceReaderFromMarketPieceDeal(ctx context.Cont
...
-				}
+			unpadded := uint64(dl.Length.Unpadded())
+			dealRawSize := unpadded
+			if dl.RawSize > 0 {
+				dealRawSize = uint64(dl.RawSize)
+			} else if rawSize > 0 {
+				// Metadata can disagree with this deal row; prefer it when deal raw_size is unset.
+				dealRawSize = rawSize
+			}
+			if dealRawSize > unpadded {
```
LexLuthr: It should never happen. We should not process this but raise an error.

itests/retrievals_test.go
```diff
@@ -233,6 +233,9 @@ func buildRetrievalFixtures(t *testing.T, dir string) retrievalFixtures {
 }
 
 func buildRetrievalSeedPlan(fixtures retrievalFixtures) []retrievalFixtureSeed {
+	// raw_size=0 is repaired asynchronously by FixRawSize; use an impossible value (> unpadded) so
+	// the retrieval assertion does not race while still exercising "invalid raw_size → full piece" behavior.
+	invalidRawSize := int64(fixtures.rawSizeZero.PieceSize.Unpadded()) + 1
```
LexLuthr: This was fixed by bumping the fix raw size scheduler interval to 1 hour which was actually reasonable. This has not been racy in any tests on main. Are you sure this branch and rebase/merged with main and you still see the race condition?

LexLuthr (commented):
This review is focused on a single commit. PR level review coming soon

LexLuthr (approved):
I am proving this as changes are required at in #1146. This seems good once we disregard those issues.

itests/retrievals_test.go
snadrus: @LexLuthr can you point me to this? it's possible that this scheduling PR & the poller changes have touched that somehow.

LexLuthr: https://github.com/filecoin-project/curio/pull/1162/changes/d671c3f410126c3130f0cbb8843dca3d43a61bd6 commit. I found the race when I was testing the correct runner size. https://github.com/filecoin-project/curio/pull/1162 which got merged into https://github.com/filecoin-project/curio/pull/1159 which got merged into `main`. 30 minutes bump was enough to not have race. It is also reasonable for a background task. It was too aggressive before.

snadrus:
> #1146

This one now showed the same problem. It's flaky too.

itests/retrievals_test.go
snadrus: reverted

PR #1166, opened 2026-04-22
PDPv0: switch to AWS CI

LexLuthr:
This PR actually has very few changes. All are related to CI and using templates.
This PR also updated Go version which is the cause of huge diff.


web/srv.go
```diff
@@ -169,20 +169,19 @@ func setupDevModeProxy(mx *mux.Router) error {
...
-		originalDirector(req)
-		req.URL.Path = path.Join(target.Path, req.URL.Path)
-
-		if !strings.HasPrefix(req.URL.Path, "/") {
-			req.URL.Path = "/" + req.URL.Path
-		}
-
-		req.URL.Scheme = target.Scheme
-		req.URL.Host = target.Host
+	proxy := &httputil.ReverseProxy{
```
rvagg, line 172: was this a lint thing, a deprecation thing, or just an improvement you spotted?

rvagg (approved):

web/srv.go
LexLuthr: lint because of deprecation. New linter forced it. `main` also has this change.

PR #1167, opened 2026-04-22
replace savecache with commp

LexLuthr:
This PR drops the local savecache library and uses the update commp-hash-hash library. The local lib was a modified version the commp-hash-hash and these changes are not available in upstream lib.


rvagg (approved):
nice

tasks/pdp/task_save_cache.go
```diff
@@ -82,7 +82,8 @@ func (t *TaskPDPSaveCache) Do(taskID harmonytask.TaskID, stillOwned func() bool)
 		}
 
 		if !has {
-			cp := savecache.NewCommPWithSize(pi.RawSize)
+			cp := commp.NewCalcWithSnapshot(commp.SnapshotLayerIndex(4 << 20))
+			defer cp.Reset()
```
Reiers, line 87: While this file is open, three small things (pre-existing, but this PR is the right moment to fix):

1. **Inversion bug at line 110** (unchanged by this PR): `if pcid2.Equals(pcid)` errors on match. Should be `if !pcid2.Equals(pcid)`. Compare `tasks/pdpv0/task_save_cache.go:136` which does it correctly — the two tasks currently disagree.
2. **Missing `n != pi.RawSize` check** after `io.CopyBuffer`. Compare `tasks/pdpv0/task_save_cache.go:119-121` which has it.
3. **Inline `4 << 20`** vs the named `PaddedReadSize` const in pdpv0 — consistency fix.

tasks/pdp/task_save_cache.go
```diff
@@ -96,7 +97,7 @@ func (t *TaskPDPSaveCache) Do(taskID harmonytask.TaskID, stillOwned func() bool)
 				return false, xerrors.Errorf("failed to copy piece data to commP: %w", err)
 			}
 
-			digest, _, lidx, _, snap, err := cp.DigestWithSnapShot()
+			digest, _, snap, err := cp.DigestWithSnapshot()
```
Reiers, line 105: Per upstream contract, `DigestWithSnapshot` returns `snapshot == nil` (with `err == nil`) when `layerIdx > treeHeight`. Current config is safe (`MinSizeForCache = 100 MiB` ensures `treeHeight ≥ 22 > 17`), but a defensive nil check here would be cheap insurance against future threshold changes or upstream behavioral drift:

```go
digest, _, snap, err := cp.DigestWithSnapshot()
if err != nil { ... }
if snap == nil {
    return false, xerrors.Errorf("snapshot layer unavailable for this piece size")
}
```

Same applies to `tasks/pdpv0/task_save_cache.go:124`.

tasks/pdpv0/task_save_cache.go
```diff
@@ -25,6 +25,7 @@ import (
 // cost of building/saving the layer. The 32 MiB threshold value was derived from conversations
 // here: https://github.com/filecoin-project/curio/pull/997#issuecomment-3960996974
 const MinSizeForCache = uint64(32 * 1024 * 1024)
+const PaddedReadSize = 4 << 20
```
Reiers, line 28: Nice to see this as a named constant. Suggest exporting it (or pulling to a shared package) so `tasks/pdp/task_save_cache.go` can use the same name instead of inlining `4 << 20` there. Small thing, but the magic number is now in two places with two different styles.

Reiers (commented):
Clean, focused refactor. The net −481 LOC comes from dropping the `lib/savecache` fork now that upstream `go-fil-commp-hashhash` v0.4.0 exposes `NewCalcWithSnapshot` / `DigestWithSnapshot` / `SnapshotLayerIndex`. Call-site translation looks correct: same layer target (~4 MiB padded per node → layer 17), same output wiring into `indexstore.NodeDigest`.

Bonus from the upstream bump: `MaxLayers` went from 35 (1 TiB cap) to 50 (~32 PiB cap). Silent capability increase. Worth noting in release notes.

## Findings

### 1. Pre-existing bug in `tasks/pdp/task_save_cache.go` — inverted CID equality

Not introduced by this PR, but this PR touches this exact function and leaves the bug in place:

```go
// tasks/pdp/task_save_cache.go:110 (unchanged)
if pcid2.Equals(pcid) {
    return false, xerrors.Errorf("commP cid does not match piece cid: %s != %s", pcid2.String(), pcid.String())
}
```

The error message says "commP cid does not match" but the condition triggers when they **do** match. Should be `if !pcid2.Equals(pcid)`.

Compare `tasks/pdpv0/task_save_cache.go:136` (correct, both in main and in this PR):
```go
if !computedV2.Equals(pcidV2) {
    return false, xerrors.Errorf("commP cid does not match piece cid: ...")
}
```

The two tasks disagree. Either the v1 path has never hit this branch in production (it's the newer PDPv1 task wired in #1106), or every valid piece is silently erroring out and retrying. Given `MaxFailures: 3`, the task would eventually give up and never mark `after_save_cache = TRUE`, stalling the pipeline for any piece > 100 MiB.

Trivial one-character fix and this PR is already in the file. Would be great to include.

### 2. Pre-existing: `pdp` task missing `n == RawSize` sanity check that `pdpv0` has

Also in `tasks/pdp/task_save_cache.go`, after `io.CopyBuffer`:

```go
n, err := io.CopyBuffer(cp, reader, make([]byte, 4<<20))
if err != nil { ... }
// No size check here — goes straight to DigestWithSnapshot
```

`tasks/pdpv0/task_save_cache.go:119-121` (this PR) has:
```go
if uint64(n) != task.RawSize {
    return false, xerrors.Errorf("copied size does not match expected piece size: %d != %d", n, task.RawSize)
}
```

Same asymmetry as #1. Truncated reads / early EOF will silently compute a CID against a shorter payload in v1 but not v0. Again trivial to add while you're here.

### 3. `snap` can be nil (per upstream contract), not checked

`DigestWithSnapshot` returns `snapshot == nil` (with `err == nil`) when the requested `layerIdx > treeHeight`:

> If the requested layer index exceeds the tree height for the data written, the CommP and padded size are still returned but snapshot will be nil (err == nil).

Current code:
```go
digest, _, snap, err := cp.DigestWithSnapshot()
if err != nil { ... }
// ...
leafs := make([]indexstore.NodeDigest, len(snap.Nodes))   // nil deref if snap == nil
for i, s := range snap.Nodes { ... }
```

**Currently safe:** `MinSizeForCache` is 100 MiB (v1) / 32 MiB (v0), both well above the 4 MiB threshold where `layerIdx=17 > treeHeight`. So in the current config, `snap` is always non-nil inside the `if pi.RawSize > MinSizeForCache` branch.

**But fragile:**
- If either threshold is ever lowered below ~4 MiB, nil panic.
- If upstream ever changes the "return nil" behavior (e.g. to zero-pad instead), silent drift.

Suggest a defensive check:
```go
if snap == nil {
    return false, xerrors.Errorf("snapshot layer unavailable (piece too small for layer %d)", commp.SnapshotLayerIndex(4<<20))
}
```

### 4. Inconsistent constant: `PaddedReadSize` vs inlined `4 << 20`

`tasks/pdpv0/task_save_cache.go` introduces `const PaddedReadSize = 4 << 20` and uses it for both the snapshot target and the copy buffer.

`tasks/pdp/task_save_cache.go` inlines `4 << 20` in both places without a constant.

Minor, but the magic number is now in two places with two different styles. Suggest pulling into a shared package constant (maybe `pdp/shared.go`) or at least matching the pdpv0 pattern in the pdp task.

### 5. Empty PR description

This is a −481 LOC refactor that removes a locally-vendored merkle tree implementation and replaces it with the upstream library. Worth a one-paragraph PR description noting:
- What `savecache` was and why the fork existed
- Which upstream version (`v0.4.0`) introduces the replacement APIs
- The silent `MaxLayers` capability increase
- Any behavioral differences verified equivalent (the test suite change is reassuring but worth calling out)

Helps reviewers and release notes.

---

## Recommendation

Good change overall, approve-leaning. I'd like to see:
- The inversion bug (#1) fixed in this PR — trivial and the file is already open.
- The `n != RawSize` check (#2) added to the pdp task — matches pdpv0.
- Nil-check on `snap` (#3) — defensive one-liner.
- PR description filled in.

#4 is cosmetic, can land here or follow-up. Otherwise the refactor itself is clean, CI is green, rvagg already approved, and the test updates look correct.

~ @Reiers

Reiers (approved):

PR #1171, opened 2026-04-23
Fix apt cache for CI tests

LexLuthr:


.github/actions/install-deps/action.yml
```diff
@@ -4,15 +4,14 @@ description: 'Install common dependencies'
...
-    description: "skips the apt update"
+    description: "when true, only apt-get update (refresh indexes) then git/submodules; skips apt install and GCC alternatives for pre-provisioned runners"
 
 runs:
   using: 'composite'
   steps:
     - name: Update apt cache
       run: sudo apt-get update
       shell: bash
-      if: ${{ inputs.skip == 'false' }}
```
rvagg, line 15: do we even need this step if we're not using the index anywhere below?

rvagg (approved):
could also do boxo 0.39 while you're in there

LexLuthr:
> could also do boxo 0.39 while you're in there

That requires pubsub upgrade which breaks Lotus and build fails.

.github/actions/install-deps/action.yml
LexLuthr: We use it in itest job.

PR #1176, opened 2026-04-24
Feat: Task polling performance

snadrus:
Lets poll a little less & treat the DB more gently:

Better indexes
Care about task-ends to inform when polls should happen for PoREP pipeline
Have market nudge poller too.
A <task>.Wake() concept so some tasks can be awakened by others (calls task adder).


tasks/storage-market/storage_market.go
```diff
@@ -57,7 +58,11 @@ const (
 	numPollers
 )
 
-const dealPollerInterval = 3 * time.Second
+// dealPollerIdleInterval is the gap after each poll() before the next idle (timer-driven) poll.
+const dealPollerIdleInterval = 30 * time.Second
```
LexLuthr, line 62: This will slow down the whole pipeline. MK12 pipeline has a lot more steps than MK20. And this poller performs some steps outside of task system.

tasks/seal/poller.go
```diff
@@ -39,7 +40,12 @@ const (
 	numPollers
 )
 
-const sealPollerInterval = 10 * time.Second
+const sealPollerInterval = 60 * time.Second
```
LexLuthr, line 43: This will make SDR tasks wait 60 seconds before they enter the pipeline.

cmd/curio/tasks/tasks.go
```diff
@@ -388,6 +400,106 @@ func StartTasks(ctx context.Context, dependencies *deps.Deps, shutdownChan chan
 	if err != nil {
 		return nil, err
 	}
+	if sealPoller != nil {
+		// Sector-targeted SignalNext: tasks that always relate to a single sector
+		// publish (spID, secNum) via the seal.PipelineRef context. Use SignalNext
+		// when known to advance the pipeline as soon as their work is done.
+		for _, taskName := range []string{"SDR", "TreeD", "TreeRC", "SyntheticProofs", "PoRep", "Finalize", "MoveStorage"} {
```
LexLuthr: This will quickly become a problem. We should avoid raw strings and use a a var which comes from the tasks themselves. One type and things don't work and we will waste time trying to figure out why.

cmd/curio/tasks/tasks.go
```diff
@@ -397,6 +509,24 @@ func StartTasks(ctx context.Context, dependencies *deps.Deps, shutdownChan chan
...
+		}
+		if dealMarket != nil {
+			watcher.AddOnLanded(dealMarket.WakeDealPoller)
+		}
+		// Snap pipeline: prove-message landing arms updateLanded (which flips
+		// sealed=TRUE for snap deals) but the snap submit task only checks
+		// from its IAmBored cycle. Wake it directly so the deal becomes
+		// index-eligible without waiting up to ~10s for the next bored tick.
+		if snapSubmit != nil {
+			watcher.AddOnLanded(snapSubmit.Wake)
```
LexLuthr: This carried a serious risk as it explodes into a scheduler. Scheduler in turn calls Lotus API which can get stuck some times which will basically hold the watcher. Nothing on watcher should be blocking. Can we disconnect them via a channel?

cmd/curio/tasks/tasks.go
```diff
@@ -388,6 +400,106 @@ func StartTasks(ctx context.Context, dependencies *deps.Deps, shutdownChan chan
...
+			taskName := taskName
+			ht.OnTaskComplete(taskName, func(ctx context.Context, _ harmonytask.TaskID, success bool) {
+				if !success {
+					return
+				}
+				spID, secNum, ok := seal.PipelineRef(ctx)
+				if !ok {
+					return
+				}
+				sealPoller.SignalNext(ctx, spID, secNum)
```
LexLuthr: Similar concern as snap scheduler. This works fine on small cluster but we might see delays and issues on large clusters. Might be a good idea to use channels to signal things instead of direct callbacks.

itests/deal_e2e_test.go
```diff
@@ -366,7 +366,8 @@ func seedVariantPipelines(
 					Offline:       false,
 					URL:           v.pieceRefURL,
 					Announce:      v.shouldIndex,
-					FastRetrieval: v.shouldIndex,
+					FastRetrieval: true,
+					ShouldIndex:   v.shouldIndex,
```
LexLuthr: Why are we bringing this back?

itests/deal_e2e_test.go
```diff
@@ -460,10 +462,20 @@ func waitForAllVariantCompletion(t *testing.T, ctx context.Context, db *harmonyd
...
+		// assertions run. With fast task wakes, deal-pipeline complete=TRUE
+		// can be set before MoveStorage finishes, so gate on that too.
+		var notMoved int
+		if err := db.QueryRow(ctx,
+			`SELECT COUNT(*) FROM sectors_sdr_pipeline WHERE after_move_storage = FALSE AND failed = FALSE`,
+		).Scan(&notMoved); err != nil {
+			return false, xerrors.Errorf("query move_storage: %w", err)
 		}
-		return false, nil
+		return notMoved == 0, nil
```
LexLuthr: Wouldn't this also impact production? Why are we adjusting the test behaviour without addressing the early complete concern?

tasks/seal/poller_precommit_msg.go
```diff
@@ -83,34 +89,53 @@ func (s *SealPoller) pollStartBatchPrecommitMsg(ctx context.Context) {
...
+		var firingResult BatchEvalResult
+		batches := GroupBatchRows(rows)
+		for i := range batches {
+			b := &batches[i]
+			result := EvalBatchTimeout(b.EarliestReady, timeout, b.MinStartEpoch, slackEpochs, currentHeight, now, len(b.SectorNums), maxBatch)
+			if result.ShouldFire {
+				if firing == nil {
+					firing = b
+					firingResult = result
+				}
```
LexLuthr, line 104: So we do only 1 batch? Then we sit till next trigger starts this? With poller that was quick but new design would mean we sit there for a lot longer. Or did I misunderstand it?
Can we add some more batching test? I would like us to cover failure and multiple batches ready etc.

tasks/seal/poller_commit_msg.go
```diff
@@ -127,42 +132,63 @@ func (s *SealPoller) pollStartBatchCommitMsg(ctx context.Context) {
...
+		var firingResult BatchEvalResult
+		batches := GroupBatchRows(rows)
+		for i := range batches {
+			b := &batches[i]
+			result := EvalBatchTimeout(b.EarliestReady, timeout, b.MinStartEpoch, slackEpochs, currentHeight, now, len(b.SectorNums), maxBatch)
+			if result.ShouldFire {
+				if firing == nil {
+					firing = b
+					firingResult = result
+				}
```
LexLuthr, line 145: Same concern as precommit

tasks/seal/poller.go
```diff
@@ -242,6 +399,83 @@ func (s *SealPoller) poll(ctx context.Context) error {
...
+												p.after_commit_msg_success,
+												p.failed, 
+												p.failed_reason,
+												p.start_epoch
+											FROM 
+												sectors_sdr_pipeline p
+											WHERE 
+												p.sp_id = $1 AND p.sector_number = $2
+												AND (p.after_commit_msg_success != TRUE 
+												OR p.after_move_storage != TRUE);`, spID, sectorNum)
```
LexLuthr, line 446: This missed using idx_sdr_pipeline_active. I am assuming that's what it is intended for?

LexLuthr (commented):
We need to audit the pipeline once and ensure this system is not adding more time to some of them.

tasks/storage-market/storage_market.go
snadrus: All  `INSERT...harmony_task` calls are accounted-for I think

tasks/seal/poller.go
snadrus: By what trigger? 
`(TaskEngine).AddTaskByName()` adds and shares an event of the add so that it can be processed immediately. Any task creation should use it.

itests/deal_e2e_test.go
snadrus: FastRetrieval didn't cover all the cases that ShouldIndex, it just got lucky due to timing to fit in the previous one that had it set.

cmd/curio/tasks/tasks.go
snadrus: solved with a unified tasknames package (from the other PR you reviewed)

itests/deal_e2e_test.go
snadrus: fixed. It had been too slow to experience, now it's an explicit check.

tasks/storage-market/storage_market.go
snadrus: TODO: Review poller mk12 & mk20 in /tasks/storage-market/

cmd/curio/tasks/tasks.go
snadrus: todo: these need batching and chan/goroutine thread separation

cmd/curio/tasks/tasks.go
snadrus: go fn() // should solve that

itests/deal_e2e_test.go
snadrus: fixed. AI lies!

tasks/seal/poller.go
snadrus: TODO solution: we need the CLI that sets-up things also run or call the poller portion. 
Simplest: find SDR task runners (from harmony_machine_details) and nudge them to run

tasks/seal/poller.go
snadrus: it has became a primary-key look-up, which is indexed

tasks/seal/poller_precommit_msg.go
snadrus: I think you missed that earliestDeadline will start the poller immediately after this one. It's not as fast as grabbing all batches, but there shouldn't be all that many at one time.

tasks/seal/poller_commit_msg.go
snadrus: same answer

cmd/curio/tasks/tasks.go
snadrus: eval: this gets 1 row by ID at a time, so batching is illogical. adding a goroutine call to callers.

tasks/storage-market/storage_market.go
snadrus: I used SignalNext() to signal a single deal to run through the progression. 
Web causes a poller wake.
CLI calls a bunch of http endpoints to cause a poller wake. 

Wake has a 10s debounce.

tasks/seal/poller.go
snadrus: CLI calls all RPC endpoints' market poller wake endpoint.

itests/deal_e2e_test.go
```diff
@@ -167,6 +167,19 @@ func TestDealPipelineFullPath(t *testing.T) {
...
+	// path. With the fast-indexing hooks (AddOnSealed → indexingTask.Wake), the
+	// deal pipeline can reach complete=TRUE before MoveStorage finishes. Wait
+	// for it so retrieval assertions don't race against file moves.
+	require.NoError(t, helpers.WaitForCondition(ctx, 60*time.Second, 2*time.Second, func() (bool, error) {
+		var notMoved int
+		if err := db.QueryRow(ctx, `SELECT COUNT(*) FROM sectors_sdr_pipeline WHERE after_move_storage = FALSE`).Scan(&notMoved); err != nil {
+			return false, err
+		}
+		return notMoved == 0, nil
+	}))
```
LexLuthr, line 181: If this is working similarly in production then we have a problem. We explicitly want indexing to happen after move storage on storage nodes and not from seal storage. Unsealed files are dropped as part of finalize and after move storage, we know what remains. If we index before that and then drop the unsealed copy, then it is all useless data bloating DB and IPNI announcements we can never serve.

LexLuthr:
Can we add pipeline timing tests here? Ideally, I would prefer itests and not just unit tests for this. We should spin up more than 1 scheduler, backed by the same DB and test for all kind of triggers and conditions between the 2. Debugging timings/racedelay issues created from this redesign will take much longer as code path and trigger are fairly complicated now.

itests/deal_e2e_test.go
LexLuthr: So, it was a timing bug. Just not how you or I expected. Earlier all deals were going in same sector due to 30 seconds poller so /piece assert was working because unsealed copy existed.
```
// Assert retrievals only if indexing was requested. Otherwise, there is a chance that the piece will not be available
		// if the should_index = false deals are all in the same sector resulting in no unsealed copy
		if v.shouldIndex {
			assertVariantPieceRetrievals(t, baseURL, v)
			assertVariantIPFSBlockRetrieval(t, baseURL, v)
		}
```

tasks/snap/task_submit.go
```diff
@@ -835,27 +869,88 @@ func (s *SubmitTask) updateLanded(ctx context.Context, tx *harmonydb.Tx, spId, s
...
 
 func (s *SubmitTask) Adder(taskFunc harmonytask.AddTaskFunc) {
+	s.adder.Set(taskFunc)
+}
+
+// Wake triggers an immediate schedule() pass without waiting for the next
+// IAmBored tick. Use this when an upstream event (e.g. prove message landing
+// on chain) makes it likely that updateLanded would now find work — calling
+// Wake collapses the worst-case ~10s IAmBored gap.
+func (s *SubmitTask) Wake() {
```
LexLuthr: This should be on moveStorage. Otherwise, we are not really waking up poller for any useful task and in the end we might end up creating indexing task from IAmBored poller only.

LexLuthr: Same is true in normal SDR sealing pipeline

tasks/indexing/task_pdp_v0_indexing.go
```diff
@@ -190,8 +191,8 @@ func (P *PDPIndexingV0Task) TypeDetails() harmonytask.TaskTypeDetails {
 	const indexingTaskRAM = 128 << 20 // 128 MiB
 
 	return harmonytask.TaskTypeDetails{
-		Name:      "PDPv0_Indexing",
-		MayFollow: []string{"PDPv0_PullPiece", "PDPv0_Prove"},
+		Name:      tasknames.PDPv0_Indexing,
+		MayFollow: []string{tasknames.PDPv0_PullPiece, tasknames.PDPv0_Prove},
```
LexLuthr, line 195: This was in parallel earlier. Since, chain msg can still take 30 seconds, we should still have this in parallel. So, we try to index and announce as soon as we have the piece on disk and don't wait for add msg to be successful.

snadrus: I'm unclear here. MayFollow expresses importance during extreme resource contention. If not extremely constrained, they will be parallel

BigLep:
@snadrus : I assume you'll re-request reviews when this is ready for reviewers to look at it again.

itests/deal_e2e_test.go
snadrus: is this resolved then?

snadrus:
> Can we add pipeline timing tests here? Ideally, I would prefer itests and not just unit tests for this. We should spin up more than 1 scheduler, backed by the same DB and test for all kind of triggers and conditions between the 2. Debugging timings/racedelay issues created from this redesign will take much longer as code path and trigger are fairly complicated now.

The theme is that everything works the same, but pollers poll less often and are triggered now. The only risk would be slowdown from a case that needs a trigger added. This would already be evident in longer time windows between Posted and Started times in harmony_task. It's likely that third-party DB-fiddling would see additional delay, but I cannot test nor inform that.

tasks/snap/task_submit.go
snadrus: done

harmony/harmonytask/completion_meta.go
```diff
@@ -0,0 +1,91 @@
...
+	fns := e.completionCallbacks[taskType]
+	e.completionMu.RUnlock()
+	if len(fns) == 0 {
+		return
+	}
+	cbCtx := buildCompletionContext(meta)
+	for _, fn := range fns {
+		fn(cbCtx, tid, true)
+	}
+}
```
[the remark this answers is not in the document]
snadrus: resolved by only doing callbacks on success

tasks/message/watch.go
```diff
@@ -55,6 +61,23 @@ func NewMessageWatcher(db *harmonydb.DB, ht *harmonytask.TaskEngine, pcs *chains
...
+// fast and non-blocking (a debounced wake signal is the intended use).
+//
+// Must be called before the first message lands; not safe to call concurrently
+// with itself or with watcher updates.
+func (mw *MessageWatcher) AddOnLanded(fn func()) {
+	if fn == nil {
+		return
+	}
+	mw.onLanded = append(mw.onLanded, fn)
+}
```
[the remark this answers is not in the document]
snadrus: comment fixed. we don't need to police that much trust

cmd/curio/tasks/tasks.go
```diff
@@ -400,6 +416,128 @@ func StartTasks(ctx context.Context, dependencies *deps.Deps, shutdownChan chan
...
+				ipniTask.Wake()
+			}
+		})
+	}
+	if snapMoveStorage != nil {
+		ht.OnTaskComplete(tasknames.UpdateEncode, func(_ context.Context, _ harmonytask.TaskID, success bool) {
+			if success {
+				snapMoveStorage.Wake()
+			}
+		})
```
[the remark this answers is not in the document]
snadrus: nah, it's not that big of a deal to do a few DB calls.

LexLuthr (approved):

PR #1178, opened 2026-04-28
Feat/cuzk-vulkan backend

snadrus:
It would be nice if CuZK didn't leave Vulkan systems behind. Lets add that to the current list of backends which includes Cuda. This expands the run scope to all modern cards including AMD, Apple M2+ and even Broadcom for those proving-servers on cell-phone hardware. 

The first version is about correct testing and uses Bellperson simplicity. 
After that, a full CuZK workalike based on Supraseal kernels will be implemented.


LexLuthr:
@snadrus Are we good to close this?

snadrus:
> @snadrus Are we good to close this?

Let's discuss today

PR #1182, opened 2026-04-29
fix/flaky tests due to fixRawSize

snadrus:
Lets pause fixRawSize for an hour for itests so we can do ugly itest tricks


LexLuthr (changes requested):
Why not just change the IAmBored time to 1 hour?

LexLuthr (approved):

Issue #1039, opened 2026-02-19
pdpv0: piece deletions lead to proving failures


ZenGround0:
In the past an issue was observed where actually processing the delete of a pdp dataset piece leads to `subpiece not found` proving errors.  This is probably going to start showing up now that we've merged @TippyFlitsUK's patch #1033.  #1033 is necessary because we are now being careful about not removing things that are still getting processed in the scheduled removals queue which involves hitting the eth rpc.  If we just keep these pieces around without deleting them ever then these rpc queries build up.

I merged #1033 because we should not consider curio pdp ready until we are at least trying to actually delete our deleted pieces.  Let's use newly arising bugs as motivation to make piece deletion actually work E2E.

This issue can track any such bugs that show up

BigLep:
2026-04-09 conversation: 
- Rod's foc-observer makes it  appear to be affecting old datasets.
- Need to look to see if we're seeing this anymore
  - If we only see it for old datasets then that gives confidence we can remove soft delete
  - If we are seeing it for new datasets, then back to drawing board as to what is going on.
- If we do remove soft delete, the need to test it with Jame on calibnet and this not endorsed SP on mainnet. Also need make sure these SPs have backups.
- This is a GA requirement because we need to make sure SPs have a way to truly remove content.
- An estimate of 3 dev days remaining was set

BigLep:
Investigation is happening in https://www.notion.so/filecoindev/Curio-Bug-Deletes-cause-proving-failures-34adc41950c1806aa7b7f3e768b1b127

LexLuthr:
I am not sure if I am behind the curve on this. But, it was proven beyond doubt that this code was not the one deleting the piece. And this converting it to comment actually did nothing to save pieces. Real issue was a different code path in `proving` task which was deleting the piece. Mongo's SP might simply need cleaning up for old deleted pieces.

ZenGround0:
@LexLuthr 

I've been digging into this and finding related evidence that removed pieces are not related to scheduling deletion.  But I haven't found any evidence of the proving task deleting pieces.  Let's try to meet up about this today.

ZenGround0:
Some relevant timing information: @beck-8 is experiencing this issue on proofsets created in early February 2026.

BigLep:
@ZenGround0 : what are the next steps for this issue?

ZenGround0:
Moving this to fast follows.  Since it is a DB issue and not data loss I am ok deferring.

[...]

PR #1186, opened 2026-04-30
Abort failed add requst

ZenGround0:
I don't quite think this is the root cause of #1039 but I found it while looking around the dataset piece table construction process.


rvagg:
Oh nice, I'm tracing something different that sounds related. I described it in https://github.com/filecoin-project/curio/pull/1188 where I fixed this exact line and a few other callsites. I'm going to close this one and ask that #1188 be reviewed instead.

PR #1187, opened 2026-05-01
ping-health

### LINKED ISSUE #851: pdpv0:  add some simple status checks for /pdp/ping
add some simple status checks for /pdp/ping

- db check
- height diff check
- other ?

Will doing some simple, time-consuming checks help troubleshoot PDP node problems faster?
can also choose a better node for Synapse SDK

snadrus:
Fixes https://github.com/filecoin-project/curio/issues/851
Caches the relevant alert results for the ping. 

Refreshes every 5 minutes.

### DIFF
--- a/alertmanager/alerts.go
+++ b/alertmanager/alerts.go
@@ -26,6 +26,7 @@ import (
 	"github.com/filecoin-project/curio/build"
 	"github.com/filecoin-project/curio/deps/config"
 	"github.com/filecoin-project/curio/harmony/harmonydb"
+	"github.com/filecoin-project/curio/tasks/tasknames"
 
 	"github.com/filecoin-project/lotus/blockstore"
 	"github.com/filecoin-project/lotus/chain/actors/adt"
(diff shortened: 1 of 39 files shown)


snadrus:
@beck-8

LexLuthr (commented):
Approach is what I wanted to do but 30 minutes window is too big practically for the use case. We probably want to have a much lower window like 5 minutes for these 3 checks. Should we have one more task with PDP specific checks? We don't want to keep saying we are unhealthy for 30 minutes when there was a transient sync issue. 

AJ:  👍

alertmanager/task_alert.go
```diff
@@ -52,6 +53,9 @@ type AlertTask struct {
 	db      *harmonydb.DB
 	plugins []plugin.Plugin
 	al      *curioalerting.AlertingSystem
+
+	pingMu       sync.Mutex
+	pingProblems bool
```
LexLuthr, line 61: Can we extend this to also point what's the cause of unhealthy node from those 3? I think it would be helpful to know what's the issue if this is frequent. Maybe a struct with 3 bools which have the details. We can log the reason before we sent the unhealthy response on HTTP side.

snadrus: I don't want to leak this to consumers. My change warns if any of these fail when it happens. It'll provide more background to the failure behavior.

snadrus:
@LexLuthr what's a PDP-specific check? 
We could check if PDP tasks are seeing "lots of failures"

LexLuthr (approved):
Looks good. Can we add alert timing tests?

alertmanager/task_alert.go
```diff
@@ -29,7 +32,8 @@ import (
 	"github.com/filecoin-project/lotus/storage/ctladdr"
 )
 
-const AlertMangerInterval = 30 * time.Minute
+const AlertMangerInterval = 5 * time.Minute
```
LexLuthr:
```suggestion
const AlertManagerInterval = 5 * time.Minute
```

LexLuthr (commented):
We should probably update the docs to new timing changes. It is no longer every 1 hour but every top of the hour.

alertmanager/task_alert.go
snadrus: done

snadrus:
> Looks good. Can we add alert timing tests?

what is that?

LexLuthr:
> > Looks good. Can we add alert timing tests?
> 
> what is that?

Ignore that please. I had a misunderstanding there.

alertmanager/alerts.go
```diff
@@ -151,11 +151,25 @@ func balanceCheck(al *alerts) {
 	}
 }
 
-// taskFailureCheck retrieves the task failure counts from the database for a specific time period.
-// It then checks for specific sealing tasks and tasks with more than 5 failures to generate alerts.
-func taskFailureCheck(al *alerts) {
-	Name := "TaskFailures"
-	al.alertMap[Name] = &alertOut{}
+// sealingTasks are sealing-pipeline task names; any single failure triggers an alert.
+var sealingTasks = []string{"SDR", "TreeD", "TreeRC", "PreCommitSubmit", "PoRep", "Finalize", "MoveStorage", "CommitSubmit", "WdPost", "ParkPiece"}
```
LexLuthr: Can we avoid raw strings and use variables?

LexLuthr: This is not a blocker for merging but vars would make this less brittle.

LexLuthr (approved):

PR #1189, opened 2026-05-01
Add pdp spec doc from pdpv0

rvagg:
Brings across the spec work @ZenGround0 did on pdpv0 to main.

### DIFF
--- a/documentation/en/experimental-features/PDPCURIOSPEC.md
+++ b/documentation/en/experimental-features/PDPCURIOSPEC.md
@@ -64,3 +64,101 @@ TODO: focus on transitions between this and next proving period not the proving 
 ## TODO: more categories
 
 # Payment Settlement
+
+# PDP Harmony Tasks
+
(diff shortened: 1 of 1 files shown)


LexLuthr (commented):
Can we finish up the TODOs? I think AI is perfect to add the first draft for remaining content from code and we can then edit/review it. Since we are in GA run now, it would not look great with half done docs.

rvagg:
@LexLuthr can we land this as-is first. @ZenGround0's original mission in this doc was to human-write it to ensure accuracy, so I'd like to land it in that state before we go expanding it out. There's more to do, but there's so much nuance in all of these points that it's a time-consuming effort to human-verify everything. AI never does a perfect job at describing the protocol even if it is pretty good, and it always needs verification anyway so that's just more human-time.

LexLuthr (approved):
Still requires the link in SUMMARY.md for the page to show up on the doc site.

PR #1190, opened 2026-05-01
fix(pay): detect rail finalization from getRail revert

rvagg:
FilecoinPay zeroes the rail struct atomically with the settle that brings settledUpTo to endEpoch, so post-confirm getRail reverts and the existing EndEpoch == SettledUpTo observation never fires. deletion_allowed is never set and DeleteDataSetTask never runs.

Treat a getRail revert as finalization when pdp_delete_data_set shows after_terminate_service = TRUE and service_termination_epoch <= current. With both met, finalization is the only explanation for the revert.

---

This comes from observation; we have **zero** properly deleted data sets in PDPVerifier on mainnet out of out of 482 and 1 out of 6716 on calibnet and it's not FWSS mediated ([ref](https://filecoinproject.slack.com/archives/C08TVNKJV7C/p1777564051017669)).

* `settleRailInternal` has a special case for terminated rails that are not zeroed (I can't find a path into this branch, maybe it's just an invariant guard) https://github.com/FilOzone/filecoin-pay/blob/2f4c9a3486f8c61d04fff127e6479f7b0483e2e0/src/FilecoinPayV1.sol#L1220-L1221
* `settleRailInternal` in the normal last-settlement pass should hit `checkAndFinalizeTerminatedRail` https://github.com/FilOzone/filecoin-pay/blob/2f4c9a3486f8c61d04fff127e6479f7b0483e2e0/src/FilecoinPayV1.sol#L1270
* `checkAndFinalizeTerminatedRail` will finalize if this was the final settlement: https://github.com/FilOzone/filecoin-pay/blob/2f4c9a3486f8c61d04fff127e6479f7b0483e2e0/src/FilecoinPayV1.sol#L1287-L1288
* `finalizeTerminatedRail` will `_zeroOutRail` https://github.com/FilOzone/filecoin-pay/blob/2f4c9a3486f8c61d04fff127e6479f7b0483e2e0/src/FilecoinPayV1.sol#L1295
* `_zeroOutRail` is pretty brutal and sets everything to their uninitialised values: https://github.com/FilOzone/filecoin-pay/blob/2f4c9a3486f8c61d04fff127e6479f7b0483e2e0/src/FilecoinPayV1.sol#L1618-L1628

The reason this is a problem for us here is that the `payment.GetRail` call we're making in tasks/pay/watcher.go is gated by an address != 0 check: https://github.com/FilOzone/filecoin-pay/blob/2f4c9a3486f8c61d04fff127e6479f7b0483e2e0/src/FilecoinPayV1.sol#L1576 so it reverts and we get an `err` and don't continue.

### DIFF
--- a/tasks/pay/watcher.go
+++ b/tasks/pay/watcher.go
@@ -146,13 +146,17 @@ func verifySettle(ctx context.Context, db *harmonydb.DB, ethClient ethchain.EthC
 			continue
 		}
 
-		// FilecoinPay zeroes the rail struct atomically with the final settle, so
-		// post-confirm GetRail reverts and the readable EndEpoch==SettledUpTo path
-		// below cannot fire. Treat the revert as finalization when FWSS still maps
-		// the rail and the lockup has elapsed.
+		// FilecoinPay zeroes the rail atomically with the final settle, so
+		// post-confirm GetRail reverts with RailInactiveOrSettled and the readable
+		// EndEpoch==SettledUpTo path below cannot fire. Treat that specific revert
+		// as finalization when FWSS still maps the rail and the lockup has elapsed.
 		if getRailErr != nil {
-			if dataSet.Int64() == 0 {
+			if !IsRailInactiveOrSettledError(getRailErr) {
 				log.Errorw("failed to get rail, skipping termination checks", "railId", railId, "error", getRailErr)
+				continue
+			}
+			if dataSet.Int64() == 0 {
+				log.Errorw("rail finalized but no FWSS dataset mapping, skipping", "railId", railId)
 				continue
 			}
 			finalized, finErr := pipelineFinalizationElapsed(ctx, db, dataSet.Int64(), current)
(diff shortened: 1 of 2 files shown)


LexLuthr (approved):

tasks/pay/watcher.go
```diff
@@ -138,15 +138,36 @@ func verifySettle(ctx context.Context, db *harmonydb.DB, ethClient ethchain.EthC
...
 		}
 
-		dataSet, err := fwssv.RailToDataSet(contract.EthCallOpts(ctx), big.NewInt(railId))
-		if err != nil {
-			log.Errorw("failed to get rail to data set, skipping termination checks", "railId", railId, "error", err)
+		// FilecoinPay zeroes the rail struct atomically with the final settle, so
+		// post-confirm GetRail reverts and the readable EndEpoch==SettledUpTo path
+		// below cannot fire. Treat the revert as finalization when FWSS still maps
+		// the rail and the lockup has elapsed.
+		if getRailErr != nil {
```
Kubuxu, line 153: Is there a way we can be more specific with this error?
I'm thinking about a case where `GetRail` fails due to a network issue, then `RailToDataSet` succeeds.

Kubuxu (approved):
One comment, but I'm fine merging as is

tasks/pay/watcher.go
rvagg: yeah .. this is an important point; I'll see if I can wrangle something

rvagg: k, added a 4-byte error detector like the one we have for fwss in pdpv0; I also had AI write me a program that ran against real RPC and known good/bad rails to check that we're getting the outcome we need from it and it's :ok_hand:

PR #1192, opened 2026-05-01
feat: robusthttp ssrf config

magik6k:
I have a dev cluster on calib where data is fetched from a service on the same machine, the protections make that setup really annoying


LexLuthr (approved):
I was expecting this PR. I went too production mode. 😉

Issue #1193, opened 2026-05-01
Show Machine version number


snadrus:
### Checklist

- [x] This is **not** a question or a support request. If you have any Curio related questions, please ask in the [discussion forum](https://github.com/filecoin-project/curio/discussions).
- [x] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/curio/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [x] I **have** searched on the [issue tracker](https://github.com/filecoin-project/curio/issues) and the [discussion forum](https://github.com/filecoin-project/curio/discussions), and there is no existing related issue or discussion.
- [x] I am running the [`Latest release`](https://github.com/filecoin-project/curio/releases), or the most recent RC(release candidate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.

### Curio component

- [ ] PoRep Pipeline
- [ ] Snap Deals
- [ ] Market
- [ ] IPNI provider
- [x] Curio UI
- [ ] Other

### Improvement Suggestion

The Cluster Machines list should have the version number shown per machine

LexLuthr:
Complete by #1194

PR #1194, opened 2026-05-01
GUI shows per-machine build Git commit hash

snadrus:
The Cluster Machines GUI would be more helpful if it included the per-machine Curio git hash. That way during "upgrade day" it would be clear what's done & what's not. 
Completes: https://github.com/filecoin-project/curio/issues/1193

### DIFF
--- a/build/version.go
+++ b/build/version.go
@@ -82,3 +82,12 @@ func CommitIDPrefix() string {
 	}
 	return rest
 }
+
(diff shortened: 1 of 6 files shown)


LexLuthr (commented):
Can we do quick screenshot for this? With long string we are likely to see a bad UI. Maybe we should only display the commit?

LexLuthr (approved):

Reiers:
Screenshot 2026-05-05 at 14 01 56

PR #1195, opened 2026-05-04
fix(pdp): race-safe parked_pieces upsert

rvagg:
From an observed (somewhere in the pdpv0 path, error logged from handlers.go:40):

    failed to create parked_pieces entry: ERROR: duplicate key value violates unique constraint "parked_pieces_active_piece_key" (SQLSTATE 23505)

Assuming this is from an update race for an identical piece. Low probability, but possible with the current check-then-execute pattern. I'm not sure how else this could be explained. Replacing with a conflict-safe upsert as a utility.

Caveats:

* market/mk20/mk20.go:887 has one of these as part of a larger batch that I've left alone
* tasks/storage-market/mk20.go:522 and tasks/storage-market/mk20.go:799 has variants as part of a larger construct that I've left alone

@LexLuthr this one's for you to consider.

### DIFF
--- a/itests/park_piece_test.go
+++ b/itests/park_piece_test.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"os"
 	"path/filepath"
+	"sync"
 	"testing"
 	"time"
 
@@ -15,6 +16,7 @@ import (
 	"github.com/filecoin-project/curio/harmony/harmonydb"
 	"github.com/filecoin-project/curio/harmony/harmonytask"
 	"github.com/filecoin-project/curio/lib/ffi"
+	"github.com/filecoin-project/curio/lib/parkpiece"
 	"github.com/filecoin-project/curio/lib/paths"
 	"github.com/filecoin-project/curio/lib/storiface"
 	"github.com/filecoin-project/curio/tasks/piece"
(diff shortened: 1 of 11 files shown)


LexLuthr:
Can we also replace those big MK20 batch inserts as well? I can add a commit if you wish. Those probably won't be able to use utility func and would require direct code changes. 
Can we also add tests to this lib? I think this is a good time to put in test so we don't break park piece unique in future.

rvagg:
I'll change the SQL to be the upsert form. But can you explain how we'd add tests here for sql? Is this an itest you want? Can you describe the basic shape of it?

LexLuthr:
> I'll change the SQL to be the upsert form. But can you explain how we'd add tests here for sql? Is this an itest you want? Can you describe the basic shape of it?

Basically just a test using these 2 funcs and more to the point a test which fail if there are ever 2 similar rows. We landed in this situation due to SQL semantics. So, I am hoping our tests try to insert using all tricks AI can think of and each must fails. Long term, this test prevents someone from removing the index without understanding the consequences.

LexLuthr:
## Recommended Tests
1. TestParkPieceUpsertRejectsDuplicateActiveRows
Use parkpiece.Upsert and parkpiece.UpsertSkip repeatedly with the same (piece_cid, piece_padded_size, long_term) and assert:
• both calls return the same id
• COUNT(*) WHERE cleanup_task_id IS NULL for that key is always 1
• changing piece_raw_size, complete, skip, or task_id cannot create a second active row
• long_term=true and long_term=false are allowed as separate rows
• rows with cleanup_task_id IS NOT NULL are allowed and do not count as active duplicates

2. TestParkPieceUpsertConcurrentWriters
Start 32 or 64 goroutines on a barrier. Each goroutine opens a transaction with harmonydb.OptionRetry() and calls either:
• parkpiece.Upsert(tx, cid, size, rawSize, true)
• parkpiece.UpsertSkip(tx, cid, size, rawSize, true, skip)
Then assert:
• no call returns 23505
• all returned ids are identical
• active row count is exactly 1
• duplicate-group query returns zero rows:
    ```
    SELECT piece_cid, piece_padded_size, long_term, count(*)
    FROM parked_pieces
    WHERE cleanup_task_id IS NULL
    GROUP BY 1,2,3
    HAVING count(*) > 1;
    ```

3. TestParkedPiecesActiveUniqueIndexBlocksAllBypassAttempts
Seed one active row, then try every bypass that previously could have worked or might be reintroduced:
• plain duplicate INSERT
• duplicate INSERT with explicit cleanup_task_id NULL
• duplicate with different piece_raw_size
• duplicate with complete=true
• duplicate with skip=true
• duplicate with stale/different task_id
• old broken form:
  `ON CONFLICT (piece_cid, piece_padded_size, long_term, cleanup_task_id) DO NOTHING`
• untargeted ON CONFLICT DO NOTHING
• INSERT ... SELECT
• insert retired duplicate with non-null cleanup_task_id, then attempt UPDATE cleanup_task_id = NULL
For every case, either require an error or require the active duplicate count to remain 1. The important assertion is “no second active row exists,” because DO NOTHING may succeed without creating a row.

LexLuthr (approved):
Nice! Thank you. We can now rest easy around these inserts.

PR #1196, opened 2026-05-04
ipni provider metrics

LexLuthr:
These metrics should help us understand choke points if any in the IPNI provider subsystem. It would also provide insight into Indexer<>Curio sync patterns.

## Pending

- [x] Merge https://github.com/filecoin-project/curio/pull/1187
- [x] Add IPNI sync to health check and also raise alert if IPNI sync is falling behind. Covered in https://github.com/filecoin-project/curio/pull/1201

### DIFF
--- a/documentation/en/configuration/metrics-reference.md
+++ b/documentation/en/configuration/metrics-reference.md
@@ -184,6 +184,16 @@ This document lists all Prometheus metrics exported by Curio. All metrics use th
 | `curio_http/request_count` | gauge/counter | Counter of HTTP requests |
 | `curio_http/response_bytes_count` | gauge/counter | Sum of HTTP response content-length |
 | `curio_http/response_status_count` | gauge/counter | Counter of HTTP response status codes |
+| `curio_ipni_announce_attempts_total` | gauge/counter | Total number of IPNI direct announce attempts. |
+| `curio_ipni_announce_http_roundtrip_seconds` | gauge | Duration of outbound IPNI announce HTTP round trips in seconds. |
+| `curio_ipni_entry_cache_hit_wait_seconds` | gauge | Duration callers wait after hitting the IPNI entry cache. |
+| `curio_ipni_entry_cache_hits_total` | gauge/counter | Total number of IPNI entry cache hits by request type, cache origin, and readiness state. |
+| `curio_ipni_entry_cache_lookups_total` | gauge/counter | Total number of IPNI entry cache lookups. |
+| `curio_ipni_entry_reconstruction_seconds` | gauge | Duration of IPNI entry reconstruction after cache miss. |
+| `curio_ipni_entry_requests_total` | gauge/counter | Total number of IPNI entry requests handled by the serve chunker. |
+| `curio_ipni_entry_speculative_unused_total` | gauge/counter | Total number of speculative IPNI entry cache fills evicted without being consumed by a demand request. |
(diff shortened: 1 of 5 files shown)


market/ipni/chunker/serve-chunker.go
```diff
@@ -109,8 +159,26 @@ func (p *ServeChunker) getEntry(rctx context.Context, block cid.Cid, speculated
...
+		if request == entryRequestDemand {
+			cacheEntry.markConsumedByDemand()
+		}
+		recordEntryCacheLookup(request, entryCacheResultHit)
+		recordEntryCacheHit(request, origin, state)
+
+		waitStart := time.Now()
+		v := cacheEntry.prom.Val(rctx)
+		observeEntryCacheHitWait(request, origin, state, time.Since(waitStart))
 		if v.Error == nil {
```
ZenGround0, line 182: Something seems a little off with recording promise error cases here as cache hits.  But this looks like a pretty rare case since I see 0 promise errors in the betterstack logs.  So this is probably good enough.

ZenGround0 (approved):
This LGTM.  I'd suggest merging this as is and then adding additional IPNI sync health check in a follow on.  But I'll leave that up to you.

BigLep:
@LexLuthr : let me know if I should look at the metric definition markdown now or wait until further updates.

LexLuthr:
> @LexLuthr : let me know if I should look at the metric definition markdown now or wait until further updates.

This is good to start. But, the file will change once I fix some assumptions in generator and move everything to [OpenCensus](https://opencensus.io/) and create dashboard YAML. That PR should be up today.

market/ipni/chunker/serve-chunker.go
LexLuthr: This not about success or error. It's simply about how long the process itself took. When there is a promise, we wait for it. We don't care if it give us the data or fails, what we care about is the wait times. If those are high then we need to change how we do speculative entry building i.e. make it parallel instead of at the end trigger. This metrics is more for dev decision making rather than anything user can do. But is is important on busy system.

documentation/en/configuration/metrics-reference.md
```diff
@@ -184,6 +184,16 @@ This document lists all Prometheus metrics exported by Curio. All metrics use th
 | `curio_http/request_count` | gauge/counter | Counter of HTTP requests |
 | `curio_http/response_bytes_count` | gauge/counter | Sum of HTTP response content-length |
 | `curio_http/response_status_count` | gauge/counter | Counter of HTTP response status codes |
+| `curio_ipni_announce_attempts_total` | gauge/counter | Total number of IPNI direct announce attempts. |
+| `curio_ipni_announce_http_roundtrip_seconds` | gauge | Duration of outbound IPNI announce HTTP round trips in seconds. |
+| `curio_ipni_entry_cache_hit_wait_seconds` | gauge | Duration callers wait after hitting the IPNI entry cache. |
+| `curio_ipni_entry_cache_hits_total` | gauge/counter | Total number of IPNI entry cache hits by request type, cache origin, and readiness state. |
+| `curio_ipni_entry_cache_lookups_total` | gauge/counter | Total number of IPNI entry cache lookups. |
+| `curio_ipni_entry_reconstruction_seconds` | gauge | Duration of IPNI entry reconstruction after cache miss. |
```
magik6k, line 192: Should be ms, it's usually really fast from CQL

LexLuthr: Thanks! I will update these.

magik6k (approved):

Issue #1180, opened 2026-04-29
pdp: no subpiece found proving error


ZenGround0:
We observe a continuous stream of unrecoverable prove_task errors that log as 

```
"failed to generate proofs: failed to prove piece 1 (<x>, <y>, <z>): no subpiece found",
```

It is traced back to this line of code: 

```golang
...
	// find first subpiece with subpiece_offset >= pieceChallengeOffset
	challSubPiece, challSubPieceIdx, ok := lo.FindLastIndexOf(subPieces, func(subPiece subPieceMeta) bool {
		return subPiece.SubPieceOffset <= pieceChallengeOffset
	})
	if !ok {
		return contract.IPDPTypesProof{}, xerrors.New("no subpiece found")
	}
...
```
This code scans through all matching subpieces in the curio `pdp_data_set_pieces` table that match a particular dataset and piece id and finds the subpiece where the challenged leaf resides.  

Something is getting mismatched between `pdp_data_set_pieces` and the onchain state of the dataset.  Either  no good offset exists or these rows are getting deleted.

ZenGround0:
This may eventually trace back to #1039 but I want to handle it separately as it is unclear, and looking at the evidence actually unexpected, that deletions would actually cause this.

beck-8:
I've also seen this issue on machines with slow IO. Testnet ID 23's database and storage are both on a RAID 5 array. A few nights ago, the array experienced high IO latency during a check, which then led to a large number of these problems.


Task | Successful | Failed
-- | -- | --
StorageGCMark | 0 | 4718 (100.00%)
SendTransaction | 397 | 0 (0%)
PDPv0_Prove | 137 | 215 (61.08%)
PDPv0_ProvPeriod | 180 | 3 (1.64%)

ZenGround0:
> I've also seen this issue on machines with slow IO

Very interesting.  Did these correspond to `"failed to generate proofs: failed to prove piece 1 (<x>, <y>, <z>): no subpiece found",
` error message?  I've been looking through logs and see many failures to retrieve pieces or query the chain api based on io conditions but I haven't seen evidence that the missing subpiece row problem is caused by io issues on the database.

beck-8:
Yes, that's the error. Perhaps it's an extreme case. I can't pinpoint exactly when the problem first occurred. But I'm certain it was caused by my disk I/O (I'm not sure whether it affected the database or the data write-to-disk process).
```
[...]

PR #1198, opened 2026-05-05
Fix: IPNI cleanup avoids piecerefs belonging to in flight adds 

ZenGround0:
This resolves an existing data race that is *possibly* behind #1180 which is characterized by a missing pdp_data_set_pieces row

 The idea is as follows:
* a piece add is triggered by handlers add sending an add message on chain and creating `pdp_data_set_piece_adds` row
* The IPNI cleanup subsection of the delete piece watcher sees a pdp_piecerefs with refcount = 0 and gets ready to delete
* The add message lands and a pdp_data_set_pieces row is inserted into the DB.  pdp_piecerefs refcount is incremented by trigger
* The IPNI cleanup process deletes pdp_piecerefs which cascades to deleting pdp_data_set_pieces

This is technically possible and hence a race we should fix anyway.  For a variety of reasons I am skeptical this is actually causing our observed issue.  Primarily the IPNI cleanup task waits 24 hours after creation and add latency is probably a lot smaller that that.  And additionally the above sequence is so pathological it is hard to see it happening so frequently.  But we should ship a fix to the issue and then see if it changes our observations.

### DIFF
--- a/tasks/pdpv0/watch_piece_delete.go
+++ b/tasks/pdpv0/watch_piece_delete.go
@@ -202,7 +202,7 @@ func processIndexingAndIPNICleanup(ctx context.Context, db *harmonydb.DB, cfg *c
 
 				if n == 0 {
 					skipLoop = true
-					log.Debugf("Piece %s with pdp_piecerefs ID %d became referenced or in-flight; skipping cleanup", piece.PieceCID, piece.ID)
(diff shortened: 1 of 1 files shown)


LexLuthr (approved):
Good catch! But, I agree with your observation that this is less likely to be the path which caused those errors. 24 hour wait time was to avoid this which makes this very unlikely.

PR #1201, opened 2026-05-06
IPNI sync alert

LexLuthr:
- Raise IPNI sync alert for all providers
- For PDP provider not syncing fail health check

### DIFF
--- a/alertmanager/alerts.go
+++ b/alertmanager/alerts.go
@@ -3,8 +3,12 @@ package alertmanager
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
+	"io"
 	"math"
+	"net/http"
+	"reflect"
 	"slices"
 	"strings"
 	"time"
@@ -26,6 +30,7 @@ import (
 	"github.com/filecoin-project/curio/build"
 	"github.com/filecoin-project/curio/deps/config"
 	"github.com/filecoin-project/curio/harmony/harmonydb"
+	"github.com/filecoin-project/curio/lib/lists"
 	"github.com/filecoin-project/curio/tasks/tasknames"
 
 	"github.com/filecoin-project/lotus/blockstore"
@@ -374,6 +379,8 @@ func (al *alerts) getAddresses() ([]address.Address, []address.Address, error) {
 			}
 		}
 	}
+
+	// TODO:  Use forEachConfig here
 
 	addrMap := make(map[string]struct{})
 	minerMap := make(map[string]struct{})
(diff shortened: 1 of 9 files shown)


alertmanager/alerts.go
```diff
@@ -375,6 +380,8 @@ func (al *alerts) getAddresses() ([]address.Address, []address.Address, error) {
 		}
 	}
 
+	// TODO:  Use forEachConfig here
```
ZenGround0: Is this TODO meant to be addressed in this PR?  If not I've had success keeping track of things like this in the past by writing up a small github issue and then linking to that issue from the TODO text.

alertmanager/alerts.go
```diff
@@ -937,3 +916,219 @@ func pendingMessagesCheck(al *alerts) {
...
+
+			if parsed.LastAdvertisementTime.Before(time.Now().Add(-1 * time.Hour)) {
+				al.alertMap[Name].alertString += fmt.Sprintf("IPNI service %s is not synced for %s provider in last 1 hour: LastAdvertisementTime is too old\n", service, d.Miner)
+				continue
+			}
+		}
+	}
+}
+
+func forEachConfig[T any](ctx context.Context, db *harmonydb.DB, cb func(name string, v T) error) error {
```
ZenGround0: nit: Since they are perfectly generic and unrelated to alerts it seems like these functions might belong better inside deps/config or perhaps a new config library

alertmanager/alerts.go
```diff
@@ -937,3 +916,219 @@ func pendingMessagesCheck(al *alerts) {
...
+				for i := range preAllocated {
+					preAllocated[i] = config.CurioAddresses{
+						PreCommitControl:      []string{},
+						CommitControl:         []string{},
+						DealPublishControl:    []string{},
+						TerminateControl:      []string{},
+						DisableOwnerFallback:  false,
+						DisableWorkerFallback: false,
+						MinerAddresses:        []string{},
+						BalanceManager:        config.DefaultBalanceManager(),
```
ZenGround0: If I'm not mistaken this is the one thing that is causing all of the Addresses checking logic?  Or is there some reason we need slices to be []string{} and not nil?

ZenGround0: I do see we have `ensureEmptyArrays` which sets nil values to empty slice values so maybe this is semantically important?

ZenGround0: What do you think about implementing all of this logic in a custom toml unmarshaller for CurioAddresses?   Here is a claude code stub which I think might be able to hide away all of this logic.  It does an extra encode step to keep things simple which is a bit annoying. But the existing code does an extra decode step to count lengths so overall I kind of like this better.

```golang
func (c *CurioAddresses) SetDefaults() {
    if c.BalanceManager == (BalanceManagerConfig{}) {
        c.BalanceManager = DefaultBalanceManager()
    }
}

func (c *CurioAddresses) UnmarshalTOML(data any) error {
    // 1. Seed defaults so anything the TOML omits stays defaulted
    c.SetDefaults()

    // 2. Round-trip through the encoder to reuse struct-tag decoding.
    //    The type alias `addr` shares CurioAddresses' field layout but
    //    doesn't inherit its methods — so toml.Unmarshal won't recurse
    //    back into UnmarshalTOML.
    var buf bytes.Buffer
    if err := toml.NewEncoder(&buf).Encode(data); err != nil {
        return xerrors.Errorf("re-encoding CurioAddresses: %w", err)
    }
    type addr CurioAddresses
    return toml.Unmarshal(buf.Bytes(), (*addr)(c))
}
```

ZenGround0: If you don't like putting this machinery to the Unmarshaller that's ok but I'd at least like to see this CurioAddresses stuff special cased behind another method.  The general function is very generic and nice and I don't like to see it crowded out by this one special case.

alertmanager/alerts.go
```diff
@@ -937,3 +916,219 @@ func pendingMessagesCheck(al *alerts) {
...
+
+		if err := cb(name, info); err != nil {
+			return xerrors.Errorf("cb: %w", err)
+		}
+	}
+
+	return nil
+}
+
+func loadConfigs(ctx context.Context, db *harmonydb.DB) (map[string]string, error) {
```
ZenGround0: nit: "// returns a map from config names to the toml strings of config bodies"

alertmanager/alerts.go
```diff
@@ -937,3 +916,219 @@ func pendingMessagesCheck(al *alerts) {
...
+			}
+
+			if parsed.LastAdvertisementTime.Before(time.Now().Add(-1 * time.Hour)) {
+				al.alertMap[Name].alertString += fmt.Sprintf("IPNI service %s is not synced for %s provider in last 1 hour: LastAdvertisementTime is too old\n", service, d.Miner)
+				continue
+			}
+		}
+	}
+}
+
```
ZenGround0: I think I get why this works to slurp up any config type.  I believe it is relying on toml unmarshal being very permissive.  If there are not no matching fields in the config the unmarshalled value will happily keep its zero value defaults.  And the unmarshalled value will happily ignore all unmatching fields.  

But there is a tricky thing here with how the callback functions that feels like a footgun. All non matching layers contribute a zero value to the callback and those zero values need to be handled.  We are fine in this PR because we do noop on nil slice value with the `..` syntax.  But this whole pattern of cb being called exactly once on all layers should be explained in a function comment

ZenGround0 (commented):
Overall this looks good.  I left some cleanup suggestions that would improve code quality substantially.

alertmanager/alerts.go
LexLuthr: I considered the custom UnmarshalTOML, but I think it is more invasive than this helper needs. The issue here is just that BurntSushi grows []CurioAddresses through reflection and creates zero-value address structs; those have nested types.FIL values under BalanceManager with nil internals, so we need to seed defaults before decode. A custom unmarshaller could do that, but then CurioAddresses takes over its own TOML decoding, needs the re-encode/alias trick, and changes BurntSushi metadata behavior by marking the whole table decoded. For this PR I’d rather keep the generic ForEachConfig flow and move the CurioAddresses default/preallocation bit behind a small private helper with a test.

alertmanager/alerts.go
LexLuthr: Moved. I should have out the PR in draft till I was done.

alertmanager/alerts.go
LexLuthr: Sorry! PR was supposed to be draft. It is now complete

alertmanager/alerts.go
LexLuthr: done

alertmanager/alerts.go
LexLuthr: Done

deps/config/utils_test.go
```diff
@@ -0,0 +1,95 @@
+package config
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestForEachConfigDecodesAddressesAndNestedConfig(t *testing.T) {
```
ZenGround0, line 10: Nice thank you!

ZenGround0 (approved):
Thanks a lot for the cleanup

Reiers (approved):

PR #1206, opened 2026-05-07
fix: synthetic proof finalize

LexLuthr:
We were acquiring the sector for synthetic proof incorrectly. This leads to below error:

` ensure one copy: not all paths are set`

### DIFF
--- a/lib/ffi/sdr_funcs.go
+++ b/lib/ffi/sdr_funcs.go
@@ -640,8 +640,17 @@ func TruncateAndMoveUnsealed(tempUnsealed, unsealed string, ssize abi.SectorSize
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


magik6k (approved):
Looks good but really unsure what this has to do with synthetic porep

LexLuthr:
> Looks good but really unsure what this has to do with synthetic porep

That's the funny part. This only happens because synthetic PoRep pathway creates unsealed copy well before finalize. Then when we run finalize we only care about cache and cleaning it up. So, for normal case we create unsealed copy and set the path correctly inside that func but for Synthetic that never happens and we see this error.

Shekelme:
With this commit, I was able to finalize several sectors on the cluster where this problem was found.
image

`curio version 1.28.0+mainnet+git_8cd6bedb_2026-05-07T16:52:45+04:00`

PR #1209, opened 2026-05-07
fix(workflow): broaden PDP label detection for fs-wg

BigLep:
## Summary
- expand PDP matching from only `pdpv0` to any `pdp` mention in issue/PR titles
- also check issue/PR body text for `pdp` mentions
- for pull requests, inspect changed file paths and label when any path contains `pdp`

## Test plan
(I'll do this if folks thinks it's necessary.)
- [ ] Open/edit an issue with `pdp` in the title and verify `team/fs-wg` is added
- [ ] Open/edit an issue with `pdp` only in the body and verify `team/fs-wg` is added
- [ ] Open/edit a PR with no `pdp` in title/body but with a changed file path containing `pdp`; verify label is added
- [ ] Open/edit an issue/PR with no PDP mention and no PDP file paths; verify no label is added

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/workflows/label-for-foc-wg.yml
+++ b/.github/workflows/label-for-foc-wg.yml
@@ -1,4 +1,4 @@
-name: Label PDPv0 Issues and PRs
+name: Label PDP Issues and PRs
 
 on:
   issues:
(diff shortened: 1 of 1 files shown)


ZenGround0 (approved):

rjan90 (approved):
Looks good to me as-is. Small optional follow-up: if we want PDP file-path detection to catch files added after the PR is opened, we should include `synchronize` in the `pull_request_target` events. Not blocking from my side.

BigLep (commented):
Thanks @rjan90! Good call — added `synchronize` to `pull_request_target` events in 74e4f917 so file-path detection re-runs when new commits are pushed to an open PR.

rjan90:
@snadrus I think this needs a green ✅ from a curio code owner before it will get merged. Should @ZenGround0 get added to that list?

LexLuthr (approved):

PR #1212, opened 2026-05-08
refactor metrics and metrics docs generator

LexLuthr:
This PR does a big refactor of how metrics are generated and exported in Curio.

- Added targeted Curio metrics for IPNI provider performance, deterministic node identity, HarmonyTask scheduling delay/runtime, wallet message landing, proof service/share, and winning compute time.
- Standardized metrics export through OpenCensus while preserving Prometheus-facing names, labels, metric types, and histogram buckets where applicable.
- Improved metrics documentation generation so exported names, labels, and types are derived from registered views/collectors, and regenerated the metrics reference.

**Disclaimer: Some of the old metric name have been updated to remove the incorrect prefix or double `curio` prefix. This PR will definitely impact the existing dashboards if any.**

### DIFF
--- a/cmd/curio/run.go
+++ b/cmd/curio/run.go
@@ -123,7 +123,7 @@ var runCmd = &cli.Command{
 			return err
 		}
 		if err := curiometrics.RecordNodeInfo(ctx, dependencies.Name, dependencies.ListenAddr); err != nil {
-			return xerrors.Errorf("recording node info: %w", err)
+			log.Errorf("recording node info, metrics will no longer have correct node label: %s", err)
 		}
 
 		go ffiSelfTest() // Panics on failure
--- a/cmd/curio/run.go
+++ b/cmd/curio/run.go
@@ -16,6 +16,7 @@ import (
 	"github.com/filecoin-project/curio/cmd/curio/rpc"
 	"github.com/filecoin-project/curio/cmd/curio/tasks"
 	"github.com/filecoin-project/curio/deps"
+	curiometrics "github.com/filecoin-project/curio/lib/metrics"
 	"github.com/filecoin-project/curio/lib/shutdown"
 
 	"github.com/filecoin-project/lotus/lib/ulimit"
@@ -120,6 +121,9 @@ var runCmd = &cli.Command{
 		err = dependencies.PopulateRemainingDeps(ctx, cctx, true)
 		if err != nil {
 			return err
+		}
+		if err := curiometrics.RecordNodeInfo(ctx, dependencies.Name, dependencies.ListenAddr); err != nil {
+			return xerrors.Errorf("recording node info: %w", err)
 		}
 
 		go ffiSelfTest() // Panics on failure
--- a/deps/stats/wallet_exporter.go
+++ b/deps/stats/wallet_exporter.go
@@ -7,7 +7,6 @@ import (
 	"time"
 
 	logging "github.com/ipfs/go-log/v2"
-	promclient "github.com/prometheus/client_golang/prometheus"
 	"go.opencensus.io/stats"
 	"go.opencensus.io/stats/view"
 	"go.opencensus.io/tag"
@@ -42,7 +41,7 @@ var WalletExporterMeasures = struct {
 	SentNFil    *stats.Int64Measure // landed messages
 	GasPaidNFil *stats.Int64Measure // landed messages (gasUsed * (basefee + tip))
 
-	MessageLandDuration promclient.Histogram
+	MessageLandDuration *stats.Float64Measure
 }{
 	BalanceNFil: stats.Int64(pre+"balance_nfil", "Balance in NanoFIL", stats.UnitDimensionless),
 	Power:       stats.Int64(pre+"power", "Power in Bytes", stats.UnitBytes),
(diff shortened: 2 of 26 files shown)


deps/stats/wallet_exporter.go
```diff
@@ -42,7 +41,7 @@ var WalletExporterMeasures = struct {
 	SentNFil    *stats.Int64Measure // landed messages
 	GasPaidNFil *stats.Int64Measure // landed messages (gasUsed * (basefee + tip))
 
-	MessageLandDuration promclient.Histogram
```
ZenGround0, line 45: Purely out of curiosity why were we using two histogram recording libraries?

scripts/metricsdocgen/metricsdocgen.go
```diff
@@ -68,6 +70,29 @@ type fileResult struct {
...
 }
 
+type viewDef struct {
+	Name        string
+	MeasureExpr string
+	Type        string
+	Labels      []string
+}
+
+var registeredLotusMetricViews = map[string]Metric{
```
ZenGround0, line 80: Are you adding these for the first time? Or were they already registered somewhere else?

documentation/en/configuration/metrics-reference.md
```diff
@@ -4,207 +4,232 @@
...
 
-This document lists all Prometheus metrics exported by Curio. All metrics use the `curio_` namespace prefix.
+This document lists Prometheus metrics exported by Curio using the exact metric names exposed on the scrape endpoint.
 
 > **Note**: This file is auto-generated from source code. Run `make docsgen-metrics` to update.
+## Node Metrics
+
+| Metric | Type | Labels | Description |
+|--------|------|--------|-------------|
+| `curio_node_info` | gauge | `node`, `version` | Curio node identity and version. Value is always 1. |
```
ZenGround0, line 14: I don't quite understand the purpose of the "Labels" column.  I guess labels are extra metadata of a particular format (specified below) that is added to metrics?

cmd/curio/run.go
```diff
@@ -121,6 +122,9 @@ var runCmd = &cli.Command{
 		if err != nil {
 			return err
 		}
+		if err := curiometrics.RecordNodeInfo(ctx, dependencies.Name, dependencies.ListenAddr); err != nil {
+			return xerrors.Errorf("recording node info: %w", err)
```
ZenGround0: You know better than me but are we sure we want to fail startup on metrics failure?  This could be a log loudly and keep going sort of error.

scripts/metricsdocgen/metricsdocgen.go
ZenGround0: Oh I see these were just not getting documented before?

harmony/harmonytask/task_type_handler.go
```diff
@@ -388,6 +388,13 @@ VALUES ($1, $2, $3, $4, $5, $6, $7, $8) RETURNING id`, tID, h.Name, postedTime.U
 		}
 		goto retryRecordCompletion
 	}
+
+	_ = stats.RecordWithTags(context.Background(), []tag.Mutator{
+		tag.Upsert(taskNameTag, h.Name),
+	},
+		TaskMeasures.TaskRuntime.M(workEnd.Sub(workStart).Seconds()),
```
ZenGround0: What's the difference between TaskDuration and TaskRuntime?  Do we need both?

tasks/proofshare/task_request.go
```diff
@@ -372,10 +322,10 @@ func (t *TaskRequestProofs) Do(taskID harmonytask.TaskID, stillOwned func() bool
 			}
 			asksCreated++
 			log.Infow("created new work ask", "askID", askID)
-			newlyAddedCounter.Inc()
+			recordProofshareNewlyAdded()
 		}
-		if asksCreated > 0 {
-			createAsksDuration.Observe(time.Since(startCreate).Seconds())
+		if neededAsks > 0 {
```
ZenGround0: Was the changing of the conditional here intentional?  I don't have quite enough context to immediately see that this is a bug fix so it looks suspicious.

ZenGround0 (approved):
Some thoughts, nothing blocking

tasks/proofshare/task_request.go
LexLuthr: This was unintentional. I will fix it. Used the wrong var.

deps/stats/wallet_exporter.go
LexLuthr: I think that might be AIs or full on build mode. We never had a standard for metrics. This is the reason for the PR. We need a standard lib.

scripts/metricsdocgen/metricsdocgen.go
LexLuthr: Yes. Just documenting now

documentation/en/configuration/metrics-reference.md
LexLuthr: Labels are logic groups within a metrics. Example: harmony_task failure count is 100 - we have task name as label. So, now we can create dashboards to tell us which kind of tasks are failing the most instead of 100 tasks failed.

harmony/harmonytask/task_type_handler.go
LexLuthr: Not really. One is label aware and other is not. I remember there was a concern about the compute cost of histogram but with float measure, I think I can update the old one and remove the new.

cmd/curio/run.go
LexLuthr: Logging only now

Issue #970, opened 2026-02-11
pdpv0: Review tasks for reorg handling


ZenGround0:
@rvagg's recently hypothesized the theory that the "piece not scheduled for removal" bug may trace back to a lack of correct reorg handling: https://github.com/filecoin-project/curio/pull/947/changes/56ce87819a36229ec2674ab83a4a53f64c5a9b1e#r2793179839

We should do a thorough review of curio tasks to see how badly reorgs impact us during operations.  For all curio db state changes that rely on messages being delivered we should 

1. Look into waiting for finalization before propagating state change.  This could be lookback based or use the new lotus finalization apis
2. Look into adding fallback cases to handle state changes that get reorged.

snadrus:
There are impossible cases here: how do we roll back 'deletePiece' when it's reorg'ed out?

rvagg:
Well, it's a good thing that deletions are "schedule deletions", so they're either in the schedule for the next proving period or not. It might get dicey as we get closer to the end of the proving period, but maybe we can use the finality probability calculator in Lotus to tell us how long to wait before somethign can be "final" and be made un-undoable.

snadrus:
Monitored and logged in now-released code:
https://github.com/filecoin-project/curio/pull/1213

PR #1213, opened 2026-05-08
Reorg-pdpv0-cleanup-task

snadrus:
This task ensures messages are on chain or otherwise alerts about the related rows for FUTURE clean-up processes to delete disk data. 
Proposed as situation monitoring for theoretical problem https://github.com/filecoin-project/curio/issues/970


harmony/harmonydb/downgrade/20260509-pdpv0-reorg-chk.sql
```diff
@@ -0,0 +1,2 @@
+DROP TABLE IF EXISTS pdpv0_reorg_orphans;
+DROP TABLE IF EXISTS pdpv0_reorg_events;
```
LexLuthr: I thought we never did this.

tasks/pdpv0/task_reorg_check.go
```diff
@@ -0,0 +1,442 @@
...
+	if cutoff.Before(time.Now().UTC().Add(-reorgCheckDefaultPast * 2)) {
+		cutoff = time.Now().UTC().Add(-reorgCheckDefaultPast * 2)
+	}
+	return cutoff, nil
+}
+
+func (t *ReorgCheckTask) txReorgedFromChain(ctx context.Context, txHash common.Hash) (reorged bool, err error) {
+	receipt, err := t.eth.TransactionReceipt(ctx, txHash)
+	if err != nil {
+		if errors.Is(err, ethereum.NotFound) {
```
LexLuthr: Are we sure this works with snapshot imports?

LexLuthr: I don't think this is how we should be checking it. We should confirm the tipset of that block and if that remains in changed then we are fine.

LexLuthr (changes requested):
We should rethink this approach. Whole Eth tx subsystem if forced to be 1 epoch only. it seems unfair that we only care about reverts for PDPv0. This task should be part of message tasks and should works on a much bigger scale. Reverts should also be checked by how forks are detected instead of missing message or receipts. Current method is not a reliable one.

tasks/pdpv0/task_reorg_check_test.go
LexLuthr, line 1: Test would also probably changed based on my review. Test should be causing a reorg on chain level. So, perhaps, we should have more subsystems involved in this and make it an itest. We must have DB, message watcher, sender, eth watcher, sender etc in the mix to have true understanding of what chaos a revert can cause and what inconsistencies it will create. This is test is too narrow.

harmony/harmonydb/downgrade/20260509-pdpv0-reorg-chk.sql
snadrus: they're historic tables but yes I shouldn't drop them.

tasks/pdpv0/task_reorg_check.go
snadrus: done

tasks/pdpv0/task_reorg_check_test.go
snadrus: I simulate a different block there on the new itest. let me know if that's sufficient.

LexLuthr:
Reverting after reorg:
1. We should not check for whole chainFinality. That is too long. 5 epoch confidence is enough same as PoRep.
2. What we are doing is just clearing tables in this. We need to think in terms of pipeline instead of just table state.

- **Create Message:** Only affects data set tables
- **Create and Add Message:** Affects data tables, piece tables, index, IPNI records. We also need to be careful with pdp_pieceref table here. Due design choices there is no guarantee that ref now does not have a correct piece attached to it. Ideally, we should use delete flow (modified) for this to be reliable.
- **Add Message:** Same suggestion as above
- **Remove Piece Message:** This can be a simple retry in pipeline. Ideally, already dropped refs, index and IPNI should not cause an issue.


After all this we still have issues from newer messages landing on chain correctly after a reorg and taking up dataSet IDs and pieceIDs. We also need to account for that.

We need to understand impact of and guard against user mistakes like importing wrong snapshot, long running forks, bad build, `lotus chain sethead` etc.


What we really need to question is: Is this worth doing with so many conditions instead of just removing the bad data?

tasks/pdpv0/reorg_rollback_integration_test.go
```diff
@@ -0,0 +1,87 @@
+//go:build harmony_itest
+
+package pdpv0
+
+import (
```
ZenGround0, line 5: Thx for tests

tasks/pdpv0/task_reorg_check.go
```diff
@@ -0,0 +1,437 @@
...
+	reasonPDPDeletePiece,
+	reasonPDPProvingInit,
+	reasonPDPProvingPeriod,
+	reasonPDPProve,
+	reasonPDPTerminateSvc,
+	reasonPDPTerminateDataSet,
+}
+
+// ReorgCheckFilAPI is the minimal Filecoin API needed for finality depth.
+type ReorgCheckFilAPI interface {
```
ZenGround0, line 67: I suggest reworking using the recent lotus finality api:

https://github.com/filecoin-project/lotus/pull/13299/changes#diff-358549c6bb769b03735c4b9eff579a6325ba8ba615b745b0bfb9b36a8b1af9e8R120

snadrus:
@LexLuthr Since we are going for clean-up rather than re-run, most just erase the ask from DB. Delete is the most sensitive because our deleter must be chain-aware.

tasks/pdpv0/task_reorg_check.go
snadrus: no thanks, I like gardening only 3x/day.

tasks/pdpv0/task_reorg_check.go
```diff
@@ -0,0 +1,437 @@
...
+// This only uses BlockByNumber, which works on snapshot-imported nodes
+// (block headers are always available, unlike TransactionReceipt which needs
+// the MsgIndex that isn't populated for pre-snapshot epochs).
+func (t *ReorgCheckTask) TxReorgedFromChain(ctx context.Context, confirmedHeight int64, storedBlockHash common.Hash) (reorged bool, err error) {
+	blk, err := t.eth.BlockByNumber(ctx, big.NewInt(confirmedHeight))
+	if err != nil {
+		return false, xerrors.Errorf("block by number %d: %w", confirmedHeight, err)
+	}
+	return blk.Hash() != storedBlockHash, nil
+}
```
LexLuthr: This assumes that when a reorg happens, our message never get's executed. That's not true. With small reorgs, there is a high chance that our message was executed and is included in heaviest tipset. Messages are not dependent on the parent and a forked node's messages are still seen and executed by correctly synced nodes. This will basically undo real work and delete correctly created dataSet/pieces etc.

LexLuthr: The only reliable way to check message inclusion is to check message inclusion. Thus, my older suggestion of checking only for 5 blocks and run this via chain watcher.

tasks/pdpv0/task_reorg_check.go
```diff
@@ -0,0 +1,437 @@
...
+		INNER JOIN message_sends_eth mse ON LOWER(TRIM(BOTH FROM mse.signed_hash)) = LOWER(TRIM(BOTH FROM mwe.signed_tx_hash))
+		WHERE mwe.tx_status = 'confirmed'
+		  AND mwe.tx_success = TRUE
+		  AND mwe.confirmed_block_number IS NOT NULL
+		  AND mse.send_success = TRUE
+		  AND mse.send_time IS NOT NULL
+		  AND mse.send_time >= $1
+		  AND mse.send_reason = ANY($2)
+		  AND ($3 - mwe.confirmed_block_number) >= $4
+	`, since, pdpv0SendReasons, headEpoch, int64(policy.ChainFinality))
```
LexLuthr: This scan window is based on `mse.send_time >= since`, but reorg eligibility depends on when the tx is confirmed/final enough, not when it was sent. A tx sent before the previous successful reorg-check run can confirm afterward and then be skipped forever because its `send_time` is older than `since`.

tasks/pdpv0/task_reorg_check.go
```diff
@@ -0,0 +1,437 @@
...
+		if rbErr != nil {
+			log.Errorw("reorg rollback failed", "tx", c.TxHash, "reason", c.SendReason, "err", rbErr)
+			continue
+		}
+
+		_, err = t.db.Exec(ctx, `
+			INSERT INTO pdpv0_reorg_events (tx_hash, send_reason, rollback_summary, data_set_id)
+			VALUES ($1, $2, $3, NULL)
+			ON CONFLICT (tx_hash) DO NOTHING
+		`, strings.ToLower(c.TxHash), c.SendReason, summary)
```
LexLuthr: We probably should handle one message changes in 1 DB tx to avoid fragmented DB state.

tasks/pdpv0/task_reorg_check.go
```diff
@@ -0,0 +1,437 @@
...
+	return true, nil
+}
+
+func (t *ReorgCheckTask) lastSuccessfulRunCutoff(ctx context.Context) (time.Time, error) {
+	var lastEnd time.Time
+	err := t.db.QueryRow(ctx, `
+		SELECT work_end FROM harmony_task_history
+		WHERE name = $1 AND result = TRUE
+		ORDER BY work_end DESC
+		LIMIT 1
```
LexLuthr: If checking or rolling back a candidate fails, the loop logs the error and continues. The task still returns success at the end, so `harmony_task_history` records this reorg-check run as successful. On the next run, `lastSuccessfulRunCutoff()` can use this successful `work_end` as the new cutoff, which may exclude the failed candidate from being retried.

tasks/pdpv0/task_reorg_check.go
LexLuthr: This query requires an INNER JOIN message_waits_eth which means any message we sent without inserting into watcher will not be checked i.e. pdp-prove and pdp-terminate-service.

tasks/pdpv0/task_reorg_check.go
```diff
@@ -0,0 +1,437 @@
...
+
+func (t *ReorgCheckTask) rollbackProvingPeriod(ctx context.Context, txHash string) (string, error) {
+	var summary string
+	_, err := t.db.BeginTransaction(ctx, func(tx *harmonydb.Tx) (bool, error) {
+		n, err := tx.Exec(`
+			UPDATE pdp_data_sets
+			SET challenge_request_msg_hash = NULL,
+			    prove_at_epoch = NULL,
+			    prev_challenge_request_epoch = NULL
+			WHERE LOWER(challenge_request_msg_hash) = $1`, txHash)
```
LexLuthr: This has a race bug.

1. init/nextProvingPeriod tx confirms.
2. `pdp_data_sets.challenge_request_msg_hash = <txHash>`.
3. Prove scheduler sees the confirmed tx and schedules a prove task.
4. Prove scheduler clears `challenge_request_msg_hash`.
5. Later, reorg checker detects that the init/nextProvingPeriod tx was reorged.
6. Rollback tries `WHERE challenge_request_msg_hash = <txHash>`, finds 0 rows, and treats that as success.


This becomes a much bigger problem when we do this after 8 hours. We are probably too late to the party there.

LexLuthr (changes requested):
We need to account for some missing chain reorg caveats here. I would also suggest checking what impact 8 hour time line has on proving subsystem. AFAIK, we would have faulted by then and this might not even be adding any value to it.

tasks/pdpv0/task_reorg_check.go
snadrus: ok for 2nd thing. For 1st thing: Reorg's boundary is 5ish blocks, but this is 8 hours. I'm not concerned there unless we lower the frequency to considerably less.

tasks/pdpv0/task_reorg_check.go
snadrus: fixed

snadrus:
@LexLuthr I see some conflicting things here. I thought this was a "clean-up" task, but If we reduce it to 5 blocks, we are making a guess about reorgs & not solving clean-up unless we "permit" failures for longer (more tables & states). If we run less frequently than proving period (30m?), they'll see a fault. A high volume of uploads cannot get chain walks one-by-one, so this must be a batched operation. 

I see these options: 
A:  "Always right, always late"
  This is just an 8h clean-up task so things get resolved eventually. 

B:  "Two bumps each" 
Say we only validate 30m to 1h, those will see a fault, but will likely be past reorg. It'll also be a reasonable batch size to amortize a short-ish chain walk. Timing vs proving is weird. 

C: "Fewest faults, gambling on definition of reorg"
We put this check at proof-time (30m?) and roll-back messages sent "5 blocks ago" that are missing. That minimizes faults. 

These each do something different. I could see a world with A & partial-C being useful. You seem to not want me to build "A". Where do I go from here?

ZenGround0:
To anchor the conversation a little just an FYI that lotus handles reorgs in a best effort way by re-inserting messages into the mpool whenever it notices a head change reverting its local chain:

https://github.com/filecoin-project/lotus/blob/master/chain/messagepool/messagepool.go#L1298 

While it's true that things can go wrong and messages can be missed, note that the original motivation for issue #970 is purely theoretical at this point.  We've never observed to my knowledge messages getting mysteriously missed after confirmation.

All of this to say -- its great to be paranoid and have something to handle long tail risk.  But let's not boil the ocean.

tasks/pdpv0/task_reorg_check.go
```diff
@@ -0,0 +1,585 @@
...
+	var lastEnd time.Time
+	err := t.db.QueryRow(ctx, `
+		SELECT work_end FROM harmony_task_history
+		WHERE name = $1 AND result = TRUE
+		ORDER BY work_end DESC
+		LIMIT 1
+	`, tasknames.PDPv0_ReorgChk).Scan(&lastEnd)
+	if err != nil {
+		if errors.Is(err, pgx.ErrNoRows) {
+			return time.Now().UTC().Add(-reorgCheckDefaultPast), nil
```
ZenGround0: So if I understand correctly at startup this will ignore messages sent 24 hours before the current time?  Just double checking that we don't go back to the beginning of time and DOS ourselves hard since I know mwe mse are notoriously unpruned.

tasks/pdpv0/task_reorg_check.go
```diff
@@ -0,0 +1,585 @@
...
+		}
+		return time.Time{}, xerrors.Errorf("last successful reorg check: %w", err)
+	}
+	if lastEnd.IsZero() {
+		return time.Now().UTC().Add(-reorgCheckDefaultPast), nil
+	}
+	cutoff := lastEnd.Add(-reorgCheckLookbackPad)
+	if cutoff.Before(time.Now().UTC().Add(-reorgCheckDefaultPast * 2)) {
+		cutoff = time.Now().UTC().Add(-reorgCheckDefaultPast * 2)
```
ZenGround0: ok nice and this handles the restart problem.  If you take your node offline for 6 weeks and turn it back on it will never check messages sent more than 2 days ago?

tasks/pdpv0/task_reorg_check.go
```diff
@@ -0,0 +1,585 @@
...
+		     AND mwe.tx_success = TRUE
+		     AND mwe.confirmed_block_number IS NOT NULL
+		     AND ($3 - mwe.confirmed_block_number) >= $4
+		     AND mse.send_time >= $1)
+		    OR (
+		      mwe.signed_tx_hash IS NULL
+		      AND mse.send_time <= NOW() - ($4::bigint * INTERVAL '30 seconds')
+		      AND NOT EXISTS (
+		        SELECT 1 FROM pdpv0_reorg_events re
+		        WHERE LOWER(re.tx_hash) = LOWER(TRIM(BOTH FROM mse.signed_hash))
```
ZenGround0, line 146: I think this is related to @LexLuthr 's comment about mse messages that are not inserted into mwe.  Perhaps useful to add a comment to clarify the purpose of this branch of the OR

ZenGround0: And just confirming my understanding:
In the first branch we only select mwe confirmed messages taht have been confirmed over a finality ago.  And in this branch we only select mse messages sent over a finality ago

tasks/pdpv0/task_reorg_check.go
ZenGround0: > For 1st thing: Reorg's boundary is 5ish blocks, but this is 8 hours. I'm not concerned there unless we lower the frequency to considerably less.

I think you are saying the gap between send time and confirmation time is negligible.  Given the vast overkill of hard finality this is is 💯 correct

tasks/pdpv0/task_reorg_check.go
```diff
@@ -0,0 +1,585 @@
...
+		`, strings.ToLower(c.TxHash), c.SendReason, summary)
+		if err != nil {
+			return false, xerrors.Errorf("insert reorg event %s: %w", c.TxHash, err)
+		}
+	}
+
+	return true, nil
+}
+
+type reorgCheckCandidate struct {
```
ZenGround0: nit: might be nice to move this to the top along with other types before it used in Do

tasks/pdpv0/task_reorg_check.go
```diff
@@ -0,0 +1,585 @@
...
+		return true, nil
+	}, harmonydb.OptionRetry())
+	if err == nil && lostRefs > 0 {
+		log.Errorw("reorg rolled back piece deletion but pieceref data already cleaned up — DATA LOSS",
+			"tx", txHash, "lost_piecerefs", lostRefs)
+	}
+	return summary, err
+}
+
+func (t *ReorgCheckTask) rollbackProvingPeriod(ctx context.Context, txHash string) (string, error) {
```
ZenGround0: My first reaction to ProvingPeriod and Prove is that these handlers should just be noops over PDP state.  Waiting for finality is never going to be acceptable for these scheduling jobs so I don't see the point handling something that indicates programmer error.  The proving and next_proving_period tasks should be robustly and much more frequently checking state and rescheduling jobs if messages somehow didn't land on chain.  If I'm wrong and reorgs could block scheduling logic indefinitely then those tasks should be fixed.

ZenGround0 (commented):
I'll finish later today.  So far I like what I see

snadrus:
I suspect this will happen one random day in a few months, and by the time we look at BetterStack traces, everything will have cleared up with the 7.5h Finality. 
If it becomes a daily matter, we will have more than enough chances to define it to the best possible efficiency.

tasks/pdpv0/task_reorg_check.go
ZenGround0: I don't think I see the problem.  If prove scheduler clears challenge_request_msg_hash and proving scheduling moves on past a missed message as it should then the reorg checker SHOULD see this as a case it doesn't need to handle.

ZenGround0: But fundamentally I think this should be a noop on PDP DB state like rollbackProve

tasks/pdpv0/task_reorg_check.go
```diff
@@ -0,0 +1,585 @@
...
+			return false, err
+		}
+		if err := t.markWaitReorged(ctx, tx, txHash); err != nil {
+			return false, err
+		}
+		summary = fmt.Sprintf("deletePiece rollback unmarked_rows=%d piecerefs_already_missing=%d", n, lostRefs)
+		return true, nil
+	}, harmonydb.OptionRetry())
+	if err == nil && lostRefs > 0 {
+		log.Errorw("reorg rolled back piece deletion but pieceref data already cleaned up — DATA LOSS",
```
ZenGround0: Logging is nice but I wonder if it would also be nice to record this in DB state somewhere either by extending orphan table or creating another one?  Not sure just wanted to throw it out there.

tasks/pdpv0/watch_piece_delete.go
```diff
@@ -81,6 +87,26 @@ func _processPendingCleanup(ctx context.Context, db *harmonydb.DB, ethClient eth
 		}
 
 		if !live {
+			// Double-check that the rm message actually landed on-chain before
+			// dropping the DB row. A reorg or otherwise dropped tx could leave
+			// us with removed=TRUE locally but no corresponding on-chain effect;
+			// in that case we want to skip and let the reorg-rollback path
+			// reset removed=FALSE rather than losing the row.
+			receipt, err := ethClient.TransactionReceipt(ctx, common.HexToHash(piece.RmTxHash))
```
ZenGround0: I think this logic ends up being a duplication of `live, err := verifier.PieceLive(contract.EthCallOpts(ctx), big.NewInt(piece.DataSetID), big.NewInt(piece.PieceID))` 
That line checks on chain state, this checks that the message doing the delete made it on.  I think this code can be removed

ZenGround0 (approved):
At this point I don't see a reason to block.

ZenGround0 (commented):
Oh I had missed this 

> Create and Add Message: Affects data tables, piece tables, index, IPNI records. We also need to be careful with pdp_pieceref table here. Due design choices there is no guarantee that ref now does not have a correct piece attached to it. Ideally, we should use delete flow (modified) for this to be reliable.
Add Message: Same suggestion as above

I agree this is a problem.  My suggestion would be to only handle reorgs of creates and deletes (deletes were kind of the original scope of the issue).  If you want to go with a full solution here then I think you do need to think about index, ipni and using a delete flow for DB state is probably the way to go.

tasks/pdpv0/task_reorg_check.go
snadrus: changes made. we won't go back more than 2000 epochs

tasks/pdpv0/task_reorg_check.go
snadrus: both branches are work reduction. watched messages have better bounds. for the worse-bound things (finality-bound), we just check if we've logged on it already.

tasks/pdpv0/task_reorg_check.go
snadrus: agreed. I added transactions now so that the logged reorg is shared with the other db changes

tasks/pdpv0/task_reorg_check.go
snadrus: now I boxed it into 2000 epochs as that's the snapshot record length, so restoring from snapshot will work.

tasks/pdpv0/task_reorg_check.go
snadrus: we now walk all, with an explicit 2000 epoch cap to be snapshot-compatible.

tasks/pdpv0/task_reorg_check.go
snadrus: we write to pdpv0_reorg_events already. Is that enough?

tasks/pdpv0/watch_piece_delete.go
snadrus: I'm pretty sure one asks if the piece is live while the other verifies that we should delete it.

tasks/pdpv0/task_reorg_check.go
snadrus: I agree. I will leave the non-PdpState update of message_waits_eth to indicate that it was reorged. Note it's now on the half-hour, so there could be scheduling impact, but this is not the right place to fiddle with that.

tasks/pdpv0/task_reorg_check.go
snadrus: Updated. Now it does no PDP state change. It just records the reorg on message_sends_eth and the pdp reorgs audit table.

tasks/pdpv0/task_reorg_check.go
snadrus: Now at the half-hour we process the 20min to last_run+20min entities. This is still way beyond the reasonable reorg window, so this is not a problem.

tasks/pdpv0/task_reorg_check.go
snadrus: fixed

snadrus:
> Oh I had missed this
> 
> > Create and Add Message: Affects data tables, piece tables, index, IPNI records. We also need to be careful with pdp_pieceref table here. Due design choices there is no guarantee that ref now does not have a correct piece attached to it. Ideally, we should use delete flow (modified) for this to be reliable.
> > Add Message: Same suggestion as above
> 
> I agree this is a problem. My suggestion would be to only handle reorgs of creates and deletes (deletes were kind of the original scope of the issue). If you want to go with a full solution here then I think you do need to think about index, ipni and using a delete flow for DB state is probably the way to go.

This creates a whole new set of scope. 
- For delete_reorg, those things should not be removed until the piece delete, which I hope we can have after the 1 hour when this runs, so that we don't need to re-add indexes and IPNI records.
- For create_reorg, we need to use a modified delete flow. 

Again, until we have a working delete flow, I cannot modify it for either of these needs.

rjan90:
Currently blocked by https://github.com/filecoin-project/curio/pull/1239 landing

tasks/pdpv0/watch_piece_delete.go
ZenGround0: No this receipt check is the same as live check.  They are both checking the same state. If you want to include the warning log in the else branch of if !live that would be ok

LexLuthr (approved):

ZenGround0:
> Currently blocked by https://github.com/filecoin-project/curio/pull/1239 landing

We're going to move forward landing this ~ as is and plug in full deletion pipeline integration in a follow on issue.  The main reasons being that 
1. As it stands this is a strict improvement over existing noop reorg handling
2. Reorgs clobbering state are currently theoretical.  The motivating issue was hypothesizing this as a reason behind an existing error but no observations support it actually happening

tasks/pdpv0/watch_piece_delete.go
ZenGround0: lmk if you want to talk more about it but basically the liveness check is going directly to PDPVerifier state to confirm that the deletion was processed on chain.  And I think that is exactly the purpose of your RmTx check.  The approaches are slightly different in that a) on if somehow a different message processed the removal then the PDPVerifier state would catch the deletion whereas the receipt would not b) if a successful deletion message was executed that somehow missed doing this deletion then PDPVerifier state would catch this as a problem but RmTx check would not.  So the PDPVerifier state check is strictly better.

ZenGround0: And AFAICT the "verify that we should delete it" check comes from the DB removed=true condition

snadrus: confirmed

snadrus:
> > Currently blocked by #1239 landing
> 
> We're going to move forward landing this ~ as is and plug in full deletion pipeline integration in a follow on issue. The main reasons being that
> 
> 1. As it stands this is a strict improvement over existing noop reorg handling
> 2. Reorgs clobbering state are currently theoretical.  The motivating issue was hypothesizing this as a reason behind an existing error but no observations support it actually happening

I've switched to log-only under a dedicated logger for easy locating.

beck-8:
from claude review
  
  ### 1. Main candidate SELECT blows up `temp_file_limit` (the actual production failure)
  `task_reorg_check.go:123-150`

  ```sql
  LEFT JOIN message_waits_eth mwe
    ON LOWER(TRIM(BOTH FROM mse.signed_hash)) = LOWER(TRIM(BOTH FROM mwe.signed_tx_hash))
  ...
  NOT EXISTS (SELECT 1 FROM pdpv0_reorg_events re
              WHERE LOWER(re.tx_hash) = LOWER(TRIM(BOTH FROM mse.signed_hash)))
  ```

  - `LOWER(TRIM(...))` defeats the PK on `message_waits_eth.signed_tx_hash` → forced Hash Join across two large tables.
  - WHERE filters on `message_sends_eth` (`send_success`, `send_time`, `send_reason`) have no usable index → seq scan on the whole table.
  - On a real node this fails with `ERROR: temporary file size exceeds temp_file_limit (1048576kB) (SQLSTATE 53400)`. Task retries forever and never makes progress.

  ### 2. Same anti-pattern in the "skipped" counter
  `task_reorg_check.go:279-291` — same `LOWER(TRIM())` NOT EXISTS, same un-indexed WHERE. Only fires when `lastEnd < epochLookback`, which is exactly what happens once (1) keeps failing.

  ### 3. Redundant `LOWER()` on rollback lookups
  `task_reorg_check.go:462, 466, 470, 486, 490, 508, 526, 560, 575, 596` — wrap PK/UNIQUE columns in `LOWER()`, forcing seq scans. Hashes are written by `tx.Hash().Hex()` (already lowercase, no whitespace), so
  `LOWER()`/`TRIM()` is unnecessary in every call site. Low-frequency code path, but should be cleaned up.

  ### 4. `pdpv0_reorg_events` index unusable as written
  `harmony/harmonydb/sql/20260509-pdpv0-reorg-chk.sql:10` — `UNIQUE (tx_hash)` is on the raw column, but every query compares `LOWER(tx_hash)`. Index will never be used once the table grows.

  Repro: node with hundreds of thousands of rows in `message_sends_eth`; query times out at 9–38 s with SQLSTATE 53400.

PR #1214, opened 2026-05-11
disable batching for rail settlement

### LINKED ISSUE #1092: pdpv0: settleRail batching uses Multicall3 aggregate which reverts entire batch on single call failure
## Problem

`BatchCallGenerate` in `lib/multicall/utils.go` packs settle calls using the `aggregate` method, which has `require(success, "Multicall3: call failed")` — if any single `settleRail` call in a batch fails, the entire batch reverts and no rails get settled.

We're seeing this on mainnet: 22 settlement failures in a single settle cycle, all with `revert reason=[Error(Multicall3: call failed)]`.

## References

- `lib/multicall/utils.go:30` — `mabi.Pack("aggregate", calls)`
- `lib/filecoinpayment/utils.go:111` — `settleRail` batched in groups of 10
- Multicall3 source: https://github.com/mds1/multicall/blob/main/src/Multicall3.sol

LexLuthr:
Fixes https://github.com/filecoin-project/curio/issues/1092
After discussion with @ZenGround0, I am disabling batching to FileocoinPay settle. It can re-enabled later if required after resolving the remaining TODO in the code.


tasks/pay/watcher.go
```diff
@@ -138,6 +153,38 @@ func verifySettle(ctx context.Context, db *harmonydb.DB, ethClient ethchain.EthC
...
+			// TODO: Should we continue here?
+			return xerrors.Errorf("failed to check rail settlement event: %w", err)
+		}
+
+		if !check.SettledEvent && !check.FinalizedEvent {
+			log.Debugw("rail settlement event not observed", "txHash", settle.Hash, "railId", railId)
+			//TODO: What should we do here?
+			/*
+				Possible reasons:
+				1. inner call failed/reverted/out-of-gas
```
ZenGround0: This will fail on chain so I don't think we can get within verifySettle

LexLuthr: We are allowing multicall to tolerate failure so actual batch message will still be successful but one of these call can fail.

lib/filecoinpayment/utils.go
```diff
@@ -24,6 +24,8 @@ import (
 
 var log = logging.Logger("filecoin-pay")
 
+const UseMulticall = false
```
ZenGround0: It would be useful to include a comment here on our current choice and what we could observe to start reconsidering

tasks/pay/watcher.go
```diff
@@ -138,6 +166,40 @@ func verifySettle(ctx context.Context, db *harmonydb.DB, ethClient ethchain.EthC
...
+		if eventErr != nil {
+			return xerrors.Errorf("failed to check rail settlement event: %w", eventErr)
+		}
+
+		if check.SettledEvent {
+			log.Debugw("rail settlement event observed", "txHash", settle.Hash, "railId", railId, "settledUpTo", check.SettledUpTo)
+			if settle.SettleUpTo.Valid {
+				if check.SettledUpTo.Int64() < settle.SettleUpTo.Int64 {
+					// Rail was settled but not up to the expected Epoch. Assume a client ran out of funds.
+					railDefaulting = true
```
ZenGround0: This is overly punitive.  The point of the grace period is to make termination safe even if client's wallet drains for a few days without noticing.  We don't want to consider a rail in default until we are at risk of 30 days of unpaid fees.

ZenGround0: In case this is unknown to you the way the terminated rail's grace period works is that the grace period covers X epochs since the last *settle* not X epochs since termination.  So the provider gains nothing from terminating aggressively like this -- the client can always pay their dues through lockup until one grace period past settlement

tasks/pay/watcher.go
```diff
@@ -201,12 +263,23 @@ func verifySettle(ctx context.Context, db *harmonydb.DB, ethClient ethchain.EthC
...
+			} else {
+				// For live rails, check if we are fully unsettled 1 day before the lockup period ends.
+				// If so assume payer is in default and schedule deletion
+				threshold := big.NewInt(0).Add(view.SettledUpTo, view.LockupPeriod)
+				thresholdWithGrace := big.NewInt(0).Sub(threshold, big.NewInt(builtin.EpochsInDay))
+				railDefaulting = thresholdWithGrace.Uint64() < current
+			}
+		}
 
-		if thresholdWithGrace.Uint64() < current {
```
ZenGround0, line 209: I don't think I see the value in checking when we could do this instead.  We are replacing a simple approach that handles the thing we care about with something much more complex that handles something we care about less.

I might be missing the value added from inspecting receipt / events etc but I suggest backing out of the diff in this file and the schema change

ZenGround0 (changes requested):
The filecoinpayment package change LGTM.  But I suggest backing out of the watcher diff entirely.

tasks/pay/watcher.go
LexLuthr: Commented it out. I would prefer not to remove the code as it also serves as a reminder of problems we face when watching for partially successful batch

lib/filecoinpayment/utils.go
LexLuthr: Added

tasks/pay/watcher.go
LexLuthr: Back to original code here.

LexLuthr:
~@ZenGround0 I have also addressed https://github.com/filecoin-project/curio/issues/1218 in here.~

ZenGround0:
> @ZenGround0 I have also addressed https://github.com/filecoin-project/curio/issues/1218 in here.

I'll figure out what's going on this time but please in the future address one issue per one PR.  I know it is a little more work to sequence things into digestible chunks when you're already deep in a particular piece of code but it makes staying organized quite difficult for everyone else.

tasks/pay/watcher.go
```diff
@@ -138,6 +166,42 @@ func verifySettle(ctx context.Context, db *harmonydb.DB, ethClient ethchain.EthC
...
+		//check, eventErr := checkRailSettlementEvent(settle.Receipt, payment, paymentContractAddr, railId)
+		//if eventErr != nil {
+		//	return xerrors.Errorf("failed to check rail settlement event: %w", eventErr)
+		//}
+		//
+		//if check.SettledEvent {
+		//	log.Debugw("rail settlement event observed", "txHash", settle.Hash, "railId", railId, "settledUpTo", check.SettledUpTo)
+		//	if settle.SettleUpTo.Valid {
+		//		if check.SettledUpTo.Int64() < settle.SettleUpTo.Int64 {
+		//			// Rail was settled but not up to the expected Epoch. Assume a client ran out of funds.
```
ZenGround0: I don't want this code to stay -- even commented out -- because `check.SettledUpTo.Int64() < settle.SettleUpTo.Int64 ` is NOT a good condition for considering a rail as defaulted.  This code as written will be introducing subtle problems for us when we transition over to multicall.

ZenGround0: My general advice is to remove commented out code for paths we are deferring.  In my experience it has rarely provided utility.  Occasionally I will keep a local branch or a gist for personal use around.  But rarely do I come back to them.

lib/filecoinpayment/utils.go
```diff
@@ -24,6 +24,11 @@ import (
 
 var log = logging.Logger("filecoin-pay")
 
+// UseMulticall determines whether to batch multiple contract calls into a single transaction to improve efficiency.
+// We are currently not using batching. Maintiners should recosider enabling batching when we start to see the below issues:
+// 1. Gas cost is too high, and the cost of sending these messages is cutting into SP profoits too deeply
```
ZenGround0: nit: profits

lib/filecoinpayment/utils.go
```diff
@@ -24,6 +24,13 @@ import (
...
+// 2. The high message count itself is impacting Curio's function. Ideally, we should improve the message sender first and then only consider batching
+// Please review the tasks/pay/watcher.go before enabling this option and ensure it is ready to handle batching
+const UseMulticall = false
```
ZenGround0: Thinking about this more -- the code in the watcher is not today ready for `UseMultiCall = true` and I'd rather we not spend our time solving the problems we'd need to to make that code work.  So if we leave this boolean guard variable we'll have a whole code path that we don't expect to work that will be exercised by changing this one constant.  For this reason I'd prefer if we just removed the multicall handling code entirely from this file.

ZenGround0: It comes down to an assessment of the benefits of leaving pieces of the solution for the road not taken vs removing those pieces.

The benefit: we have a partial solution so we'll need to write less code if/ when we pick this up again.
The cost: the code carries more complexity for an equal amount of functionality. 

The benefit is especially small for the multicall handling code in this particular instances given that we can search back in git history for our negative diffs and get back all of this code with a few minutes of searching.

ZenGround0: I'm making a similar suggestion in the watcher code for a similar reason.  In that case it is a little more beneficial to leave commented code since nothing is actually checked into git history.  But I'd prefer preserving that elsewhere, maybe the diff of the relevant commit pasted into an issue for example.  And again the motivation is reducing complexity

harmony/harmonydb/sql/20260512-filecoin-pay-settle-epoch.sql
```diff
@@ -0,0 +1,11 @@
+DO $$
+BEGIN
+    IF NOT EXISTS (
+        SELECT 1 FROM information_schema.columns
+        WHERE table_name = 'filecoin_payment_transactions'
+        AND table_schema = current_schema()
+        AND column_name = 'settle_upto_epoch'
```
ZenGround0: My point stands that we would not want to use this for implementing multicall.  So I don't think we should keep this.

LexLuthr: Removed

tasks/pay/watcher.go
LexLuthr: Removed

lib/filecoinpayment/utils.go
LexLuthr: Removed

ZenGround0:
A good exhibit of one of the edge cases that makes separating different issues into different PRs: https://github.com/filecoin-project/curio/issues/1218#issuecomment-4447712313

It looks like maybe that issue was not based on an accurate premise and so we could probably skip or defer the PR addressing it to unblock you rather than untangling it from this PR which is otherwise good to go.

ZenGround0 (approved):
Thanks!

PR #1216, opened 2026-05-12
detailed ipni piece status

### LINKED ISSUE #1156: pdpv0: IPNI latency is our fault?
From discussion with @aschmaman 

It appears that IPNI processes the ad chain quite quickly and spends most of its time waiting on getting the ad chain from curio's http handler.  Something on the order of ~1s for filecoin pin.contact ingestion and 5-10s for waiting for curio to respond to request.

This is a mystery.  Curio http handlers should not be blocked by task scheduling artifacts.  The ad chain should be stored in DB before talking to IPNI with the first push so curio code shouldn't obviously be taking so long.

LexLuthr:
Fixes https://github.com/filecoin-project/curio/issues/1156
This is the final PR for the issue. This couples with metrices should give us enough data to understand IPNI latencies.

### DIFF
--- a/pdp/handlers.go
+++ b/pdp/handlers.go
@@ -358,16 +358,16 @@ func (p *PDPService) handleGetPieceStatus(w http.ResponseWriter, r *http.Request
 		if publishedAt != nil && publishedAt.After(result.AdvertisementCreatedAt.Time) {
 			response.Advertised = true
 			response.AdvertisedAt = new(*publishedAt)
-			if publishedAt.After(time.Now().Add(time.Second * 5)) {
-				response.AdvertisedAt = new(result.AdvertisementCreatedAt.Time.Add(time.Second * 5))
+			if publishedAt.After(time.Now().Add(ipni_provider.PublishInterval)) {
+				response.AdvertisedAt = new(result.AdvertisementCreatedAt.Time.Add(ipni_provider.PublishInterval))
 			}
 		}
 	} else {
 		if result.AdvertisementCreated && !result.AdvertisementRetrievedAt.Valid {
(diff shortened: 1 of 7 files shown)


pdp/handlers.go
```diff
@@ -302,19 +316,63 @@ func (p *PDPService) handleGetPieceStatus(w http.ResponseWriter, r *http.Request
...
+		response.RetrievedAt = &result.AdvertisementRetrievedAt.Time
+	}
+
+	// If we have IPNI provider, check if advertisement was published directly from IPNI provider
+	if p.ipp != nil && result.Provider.Valid {
+		publishedAt := p.ipp.LastPublishTime(result.Provider.String)
+		if publishedAt != nil && publishedAt.After(result.AdvertisementCreatedAt.Time) {
+			response.Advertised = true
+			response.AdvertisedAt = new(*publishedAt)
+			if publishedAt.After(time.Now().Add(time.Second * 5)) {
```
BigLep: `5` is hardcoded here and below. Can that be tied to a constant that is used for scheduling the advertisement announcement task to run every 5 seconds to drive home that these are linked and to make sure they don't diverge? (Basically I'm suggesting to get rid of the "5" magic number.)

BigLep (commented):
New response structure looks good - thanks guys!

pdp/handlers.go
ZenGround0: We could reuse `publishInterval` after exporting it from `ipni_provider` package

ZenGround0 (approved):
Looks great

pdp/handlers.go
LexLuthr: My bad. I missed changing that.

Reiers (approved):

Issue #1191, opened 2026-05-01
PDP: Sweep stuck deletion-pipeline rows for finalised rails not seen by SettleWatcher


rvagg:
The `GetRail`-revert fix in https://github.com/filecoin-project/curio/pull/1190 only catches finalisation at the moment a settle tx submitted by _this SP_ confirms. There's at least two cases it misses and we can't handle in the existing watcher:

1. **Already-finalised backlog**: rails that finalised before the fix landed. `pdp_delete_data_set` rows sit at `deletion_allowed = FALSE` forever.
2. **External finalisation**: client calls `settleRail` or `settleTerminatedRailWithoutValidation` directly, or another operator settles. Curio never submits a settle tx for it, so `SettleWatcher` never fires. The settle task can't rediscover the rail because `GetRailsForPayeeAndToken` returns nothing for a zeroed rail (see #1190 for explanation of the zeroed rail thing).

This all results in stuck `pdp_delete_data_set` rows, data set stays live in PDPVerifier indefinitely.

## Proposed fix

A periodic sweep (`IAmBored` or polling task) over `pdp_delete_data_set` rows where `after_terminate_service = TRUE AND deletion_allowed = FALSE AND service_termination_epoch IS NOT NULL AND service_termination_epoch <= current`. For each:

1. Look up `pdpRailId` via `fwssv.GetDataSet(dataSetId)`.
2. Call `payment.GetRail(pdpRailId)`. If it reverts, we're finalised and can `ensureDataSetDeletion`. If it's readable and `EndEpoch == SettledUpTo` then we're also finalised.

This is in addition to the `GetRail`-revert detection in `SettleWatcher` (#1190); that path stays as the fast path, the sweep is the backstop for the non-standard paths that are still possible.

LexLuthr:
Agreed! But, I would suggest a broader scope. This task runs once or twice a day. We can put all missing sync things in this task instead of just deletion. So basically, this task ensure that our local DB is in sync with chain status. If something is amiss, we fix it. 

Method 1 -> Check FWSS termination, fix if required
Method 2 -> Check PDP termination, fix is required
Method 3 -> Remove pieces locally if onChain no longer exists
......

D0() {
Method 1
Method 2
Method 3
}



This saves us grief over one task per kind of sync we need.

rvagg:
[...]

Issue #889, opened 2026-01-15
PDP Doctor CMD DB state with On-Chain State


rjan90:
Rough, advisory sketch of how to take a stab at the DB vs On-Chain state inconsistency problem, written by @rvagg:

## Curio PDP Doctor Command Design

A diagnostic tool to check consistency between local DB state and on-chain state across PDPVerifier, FWSS, and FilecoinPay contracts.

## Project Structure

**Initial approach**: Start as a standalone external project/tool. This allows rapid iteration without curio release cycles and keeps the diagnostic tooling decoupled from production code.

**Future integration**: May eventually be integrated into the `curio` CLI as a subcommand (e.g., `curio pdp doctor`) depending on quality and maintainer preference. Design with this in mind - use curio's existing contract bindings and database utilities where possible.

**Dependencies** (for standalone version):
- `github.com/filecoin-project/curio/pdp/contract` - Contract bindings (PDPVerifier, FWSS, FilecoinPay)
- `github.com/filecoin-project/curio/harmony/harmonydb` - Database connection utilities
- `github.com/ethereum/go-ethereum` - Ethereum client
- PostgreSQL driver for Yugabyte/Postgres connection

## Context: Where This Fits

**Repository**: Initially standalone; eventual integration target is `github.com/filecoin-project/curio` (branch: `pdpv0`)

**Related Files**:
- `tasks/pdp/` - All PDP task implementations
- `tasks/pay/` - Settlement task (`settle_task.go`), watcher
- `pdp/` - HTTP API handlers
- `pdp/contract/` - Generated contract bindings
- `pdp/contract/addresses.go` - Contract addresses per network
- `lib/chainsched/` - Chain event system
- `lib/filecoinpayment/` - Payment utilities, Multicall3 batching
- `harmony/harmonydb/` - Database utilities

**Contract Addresses** (from `pdp/contract/addresses.go`):

| Contract | Calibnet | Mainnet |
|----------|----------|---------|
| PDPVerifier | `0x85e366Cf9DD2c0aE37E963d9556F5f4718d6417C` | `0xBADd0B92C1c71d02E7d520f64c0876538fa2557F` |
| FWSS | `0x02925630df557F957f70E112bA06e50965417CA0` | `0x8408502033C418E1bbC97cE9ac48E5528F371A9f` |
[...]

PR #1220, opened 2026-05-13
pdpv0: chain state sync task

### LINKED ISSUE #473: PDP unrecoverable termination should update FWSS dataset state
`FilecoinWarmStorageService.piecesAdded` gates writes on FWSS payment state via `requirePaymentNotTerminated`, which checks `pdpEndEpoch == 0`. It does not check PDP liveness.

Today `pdpEndEpoch` flips through FWSS/payment termination paths:

- `terminateService(dataSetId)`, authorized for the payer or service provider
- `railTerminated`, the FilecoinPay callback

PDP can mark a dataset terminated, for example after unrecoverable proving failure, without FWSS state changing. The dataset then still looks usable through FWSS views while writes fail at the PDP layer. `dataSetDeleted` is downstream cleanup, not a sync point, because it already requires `pdpEndEpoch != 0`.

### Proposal

Add a PDP → FWSS synchronization path for unrecoverable PDP dataset termination. FWSS should set `pdpEndEpoch` or an equivalent terminal-state flag when PDP terminates a dataset, so FWSS views and downstream consumers observe coherent dataset state without each caller needing to reactively call `terminateService`.

### Context

Encountered in FilOzone/dealbot#379. Dealbot is tracking a reactive workaround there, but the underlying PDP/FWSS state divergence can affect other consumers too.

### Related prior work/context:

- #136 covered exposing whether datasets are usable through FWSS views. This issue is a narrower remaining gap: PDP can mark a dataset non-live while FWSS still reports `pdpEndEpoch == 0`.
- #380 is an adjacent incident where terminated/unprovable datasets caused downstream operational issues in Curio. It did not address PDP → FWSS state synchronization.
- #169 is broader lifecycle/status context if this fix ends up changing FWSS dataset status semantics.

LexLuthr:
This task is a single place to handle issues stemming from chain state and local DB not being in sync.

Currently, it handles the following things:

1. Clears stale terminate/delete task IDs after Harmony has exhausted and removed the old task.
2. Resets PDP proving backoff state when chain confirms the dataset was actually proven.
3. Enables dataset deletion once the post-termination payment rail is finalized.

Fixes https://github.com/FilOzone/filecoin-services/issues/473 and https://github.com/filecoin-project/curio/issues/1191


ZenGround0:
Nice!  FYI there is this related discussion #889

cmd/curio/tasks/tasks.go
```diff
@@ -359,9 +359,10 @@ func StartTasks(ctx context.Context, dependencies *deps.Deps, shutdownChan chan
 
 			pdpTerminate := pdpv0.NewTerminateServiceTask(db, must.One(dependencies.EthClient.Val()), senderEth)
 			pdpDelete := pdpv0.NewDeleteDataSetTask(db, must.One(dependencies.EthClient.Val()), senderEth)
+			pdpChainSync := pdpv0.NewTaskChainSync(db, must.One(dependencies.EthClient.Val()), senderEth)
```
ZenGround0: nit: since ChainSync has a very particular meaning in filecoin / blockchain context I'd suggest renaming to ChainDBSync or something like that

tasks/pdpv0/failure_handling.go
```diff
@@ -126,38 +126,42 @@ func ResetProvingFailures(ctx context.Context, db *harmonydb.DB, dataSetId int64
 //   - Tier 2: Other contract reverts, apply backoff, may terminate after repeated failures
 //   - Tier 3: Transient errors, return error for harmony retry
 //
-// Returns (done, err) where done=true means the task should complete (not retry),
+// Returns (err) where err==nil means the task should complete (not retry),
 // and err!=nil means harmony should retry the task.
-func HandleProvingSendError(ctx context.Context, db *harmonydb.DB, dataSetId int64, currentHeight int64, sendErr error) (done bool, err error) {
+func HandleProvingSendError(tx *harmonydb.Tx, dataSetId int64, currentHeight int64, sendErr error) error {
```
ZenGround0, line 131: Thanks for moving to tx.  I think I am finally over the hill of understanding with postgress transactions.

tasks/pdpv0/task_chain_sync.go
```diff
@@ -0,0 +1,385 @@
...
+		return xerrors.Errorf("failed to commit stale task id cleanup")
+	}
+
+	return nil
+}
+
+// syncProvenDataSetFailureState reconciles local proving backoff after a proof has been accepted on-chain.
+// The reset is intentionally conservative: the data set must still be live, PDPVerifier must report a useful last proven epoch,
+// and FWSS must mark the corresponding proving period as proven.
+func (t *TaskChainSync) syncProvenDataSetFailureState(ctx context.Context) error {
```
ZenGround0, line 139: Is this necessary?  Aren't we doing this be default when handling proving tasks?

tasks/pdpv0/task_chain_sync.go
```diff
@@ -0,0 +1,385 @@
...
+	period := new(big.Int).Sub(epoch, activationEpoch)
+	period.Sub(period, big.NewInt(1))
+	period.Div(period, new(big.Int).SetUint64(maxProvingPeriod))
+	return period, true
+}
+
+// syncFinalizedDataSetDeletionRails moves terminated PDP data sets to the local deletion-allowed state once the payment rail is final.
+// A rail lookup revert or EndEpoch == SettledUpTo means settlement is complete; this pass only toggles
+// deletion_allowed and leaves the actual delete to the existing delete task.
+func (t *TaskChainSync) syncFinalizedDataSetDeletionRails(ctx context.Context) error {
```
ZenGround0, line 219: Again don't we already do this in settlement?  I am a little worried about the scope of this task going from handling things that aren't handled anywhere else to trying to act as a backup to existing mechanisms that we expect to work.

ZenGround0 (commented):
Overall I like it.  I want to talk more about the additional two state sync jobs beyond `syncStaleDeletionTaskIDs`.  Where is the motivation for these coming from?  Are we not handling these state transitions sufficiently in pdp proving and rail settlement?

tasks/pdpv0/task_chain_sync.go
ZenGround0: Is this to handle the case where the client actually initializes the settlement?

ZenGround0: ok my bad this is what is described in #1191 makes sense

ZenGround0 (approved):
I'm going to leave this approved with the caveat that I think `syncProvenDataSetFailureState` may be unnecessary and if so should be removed.  We should be doing retries during proving scheduling and we remove the failure condition if that retried message is successfully sent.

tasks/pdpv0/task_chain_sync.go
```diff
@@ -0,0 +1,385 @@
...
+
+	fwssv, err := FWSS.NewFilecoinWarmStorageServiceStateView(viewAddr, t.ethClient)
+	if err != nil {
+		return xerrors.Errorf("failed to instantiate FWSS service state view: %w", err)
+	}
+
+	maxProvingPeriod, err := fwssv.GetMaxProvingPeriod(contract.EthCallOpts(ctx))
+	if err != nil {
+		return xerrors.Errorf("failed to get FWSS max proving period: %w", err)
+	}
```
rvagg: You shouldn't need to involve FWSService directly here, grep around the codebase for `GetPDPConfig`, what you want is to find the listener of the data set, then call it as a `IPDPProvingSchedule` to `GetPDPConfig` which has `maxProvingPeriod` as the first field in the return value. Since the listener is likely to be the same for all of the data sets you check you could keep a cached value associated with the listener address so you don't need to call it every time.
Try and avoid doing FWSS-specific stuff where possible.

tasks/pdpv0/task_chain_sync.go
```diff
@@ -0,0 +1,385 @@
...
+			continue
+		}
+
+		proven, err := fwssv.ProvenPeriods(contract.EthCallOpts(ctx), dataSetID, provingPeriod)
+		if err != nil {
+			return xerrors.Errorf("failed to check proven period %s for data set %d: %w", provingPeriod, dataSet.ID, err)
+		}
+		if !proven {
+			continue
+		}
```
rvagg: these FWSS-specific jobs are probably unnecessary, they're just sanity checking, while `GetDataSetLastProvenEpoch` is the one you care about; the outcome of these calls is mostly just logging and y our `if !proven continue` should be covered by the `if lastProvenEpoch == nil continue` above this block.

tasks/pdpv0/task_chain_sync.go
```diff
@@ -0,0 +1,385 @@
...
+
+		lastProvenEpoch, err := pdpVerifier.GetDataSetLastProvenEpoch(contract.EthCallOpts(ctx), dataSetID)
+		if err != nil {
+			return xerrors.Errorf("failed to get last proven epoch for data set %d: %w", dataSet.ID, err)
+		}
+		if lastProvenEpoch == nil || lastProvenEpoch.Sign() <= 0 {
+			continue
+		}
```
rvagg, line 181: `dataSetProofCoversLocalFailure` has this same check, maybe just do it once?

tasks/pdpv0/task_chain_sync.go
rvagg: Good question, I'd say it can't hurt to make sure we're in sync and we haven't had different task workers get us into a desync state (maybe we proved but didn't get to update) but it'd be worth checking whether the proving task is resillient to this so maybe it auto-corrects itself later.

tasks/pdpv0/task_chain_sync.go
```diff
@@ -0,0 +1,385 @@
...
+		return false
+	}
+
+	lastFailureEpoch := nextProveAttemptEpoch.Int64 - int64(CalculateBackoffBlocks(consecutiveFailures))
+	return lastProvenEpoch.Cmp(big.NewInt(lastFailureEpoch)) > 0
+}
+
+// provingPeriodForEpoch mirrors FWSS _provingPeriodForEpoch. The state-view binding exposes provenPeriods but not this helper,
+// so chain sync computes the period locally before checking the durable proven-period bitmap.
+func provingPeriodForEpoch(activationEpoch *big.Int, epoch *big.Int, maxProvingPeriod uint64) (*big.Int, bool) {
```
rvagg: as per my comment above, anything FWSS-specific should be avoided

but if it's interesting and informational, just make it optional, if listener is not FWSS then skip, don't let it be a cause to stop doing the important stuff

tasks/pdpv0/task_chain_sync.go
```diff
@@ -0,0 +1,385 @@
...
+	for _, dataSet := range pending {
+
+		ds, err := fwssv.GetDataSet(contract.EthCallOpts(ctx), big.NewInt(dataSet.ID))
+		if err != nil {
+			return xerrors.Errorf("failed to get data set %d from FWSS view: %w", dataSet.ID, err)
+		}
+
+		rail, err := payments.GetRail(contract.EthCallOpts(ctx), ds.PdpRailId)
+		if err != nil {
+			if IsContractRevert(err) {
```
rvagg: `IsContractRevert` is mostly fine but covers a bunch of things so it's a wider net than we need. We could add a function that does a specific check for `RailInactiveOrSettled` like we do with the `IsUnrecoverableError` in error_detection.go to be specific.

pdp/contract/FWSS/utils.go
```diff
@@ -19,3 +19,14 @@ func EnsureServiceTermination(ctx context.Context, db *harmonydb.DB, dataSetID i
 	}
 	return nil
 }
+
+func EnsureServiceTerminationTx(tx *harmonydb.Tx, dataSetID int64) error {
```
rvagg: can we delete `EnsureServiceTermination` now that we have this? same sql

rvagg (approved):
Pre-emptive approval from me given the feedback is considered; mostly I'm concerned about baking in FWSS-specific stuff where it's not need, better to keep the PDPVerifier focus wherever possible and only branch outward where necessary (like needing to care about payment rails).
Pretty nice.

pdp/contract/FWSS/utils.go
LexLuthr: Yes, It would require few more changes. Especially, to the settlement watcher but it should be easy.

tasks/pdpv0/task_chain_sync.go
LexLuthr: Removed

tasks/pdpv0/task_chain_sync.go
LexLuthr: Removed

tasks/pdpv0/task_chain_sync.go
LexLuthr: Removed

tasks/pdpv0/task_chain_sync.go
LexLuthr: Removed

tasks/pdpv0/task_chain_sync.go
LexLuthr: Removed maxProvingPeriod from FWSS but we still need FSWW for rail related calls.

tasks/pdpv0/task_chain_sync.go
LexLuthr: Yes. We don't fail the task when we successfully prove a DS but fail to reset the same in DB. I have added a comment in proving task to explain the same.

tasks/pdpv0/task_chain_sync.go
LexLuthr: This is meant to handle https://github.com/filecoin-project/curio/issues/1191.

Settle watcher retry only helps while the settle tx row still exists in filecoin_payment_transactions. With the latest watcher changes, DB failures from EnsureServiceTermination / ensureDataSetDeletion now return an error, so the watcher does not delete the settle row and can retry. That fixes the “DB update failed but settle row got consumed” problem for future cases.
But it does not cover all stuck cases:

1. Historical stuck rows likely already lost their filecoin_payment_transactions row under old logic, so settle watcher has nothing to retry.

2. The watcher still has intentional skip paths that do not return an error, then it deletes the settle tx row: RailToDataSet failure, non-RailInactiveOrSettled GetRail failure, and RailInactiveOrSettled before local pipeline state has caught up.

3. The sync task starts from the durable stuck state itself: pdp_delete_data_set.after_terminate_service = TRUE, deletion_allowed = FALSE, service_termination_epoch <= current. That is exactly the state we care about, independent of whether the original settle event is still retryable.

Settle retry fixes some future watcher failure modes, but it does not replace syncFinalizedDataSetDeletionRails. The sync func remains the reconciliation backstop for already-consumed settle rows, local/chain timing races, and any watcher skip path that left the local deletion row behind.

cmd/curio/tasks/tasks.go
LexLuthr: Yes, it does. But, Curio has it's own terminology and I don't think a variable name somewhere deep in task init system would somehow mean Curio is now doing chain sync. I would still rename it as per your suggestion.

tasks/pdpv0/task_prove.go
```diff
@@ -458,7 +458,8 @@ func (p *ProveTask) Do(taskID harmonytask.TaskID, stillOwned func() bool) (done
 		return true, nil
 	}
 
-	// Success, reset any accumulated failure count
+	// Success, reset any accumulated failure count. We cannot fail the task here to avoid sending proof for same dataSet multiplt times
+	// In case, this reset fails, it should be handled in task_chain_sync
```
ZenGround0, line 462: 🙏

tasks/pdpv0/task_prove.go
```diff
@@ -458,7 +458,8 @@ func (p *ProveTask) Do(taskID harmonytask.TaskID, stillOwned func() bool) (done
 		return true, nil
 	}
 
-	// Success, reset any accumulated failure count
+	// Success, reset any accumulated failure count. We cannot fail the task here to avoid sending proof for same dataSet multiplt times
```
ZenGround0: nit: multiple

tasks/pdpv0/task_prove.go
ZenGround0: One thing to note is that this function can only error if db.exec errors.  It is nice that we have a way to recover from such errors but I have been mentally modeling this as something that should basically never happen.  Am I being too optimistic about db errors being rare?  Should I instead be thinking about defensive programming against a fairly frequence db error rate?

ZenGround0: I guess the most obvious thing to defend against are shut down edge cases since ctx.Cancelled will error out db.Exec.  Im not sure how often post gres connection errors should be expected.

tasks/pdpv0/task_chain_sync.go
```diff
@@ -162,25 +161,6 @@ func (t *TaskChainSync) syncProvenDataSetFailureState(ctx context.Context) error
 		return xerrors.Errorf("failed to instantiate PDPVerifier contract: %w", err)
 	}
 
-	sAddr := contract.ContractAddresses().AllowedPublicRecordKeepers.FWSService
```
ZenGround0: My bad I missed this before.  It is really good to be cutting out this FWSS stuff here.

tasks/pdpv0/task_chain_sync.go
```diff
@@ -235,24 +197,20 @@ func (t *TaskChainSync) syncProvenDataSetFailureState(ctx context.Context) error
...
 	return nil
 }
 
-// dataSetProofCoversLocalFailure checks whether the on-chain proof happened late enough to explain the local failure
-// state we are about to clear. Prefer prove_at_epoch when it is still present; fall back to the reconstructed
+// dataSetProofCoversLocalFailure checks whether the PDPVerifier state advanced
+// late enough to explain the local failure state we are about to clear. Prefer
+// prove_at_epoch when it is still present; fall back to the reconstructed
 // failure epoch when the normal proof state has already moved on.
 func dataSetProofCoversLocalFailure(proveAtEpoch sql.NullInt64, consecutiveFailures int, nextProveAttemptEpoch sql.NullInt64, lastProvenEpoch *big.Int) bool {
```
ZenGround0: nit: `dataSetProofAfterLocalFailure` OR `dataProofFollowsLocalFailure` or `dataSetProofMoreRecentThanLocalFailure` are slightly clearer to me

ZenGround0: But actually given the comparitive size of this and `syncProvenDataSetFailureState` and the fact that this function is very conceptually coupled to that one I think this function would make more sense to keep inline with a block comment on top of it describing what is going on.

ZenGround0 (approved):

tasks/pdpv0/task_prove.go
LexLuthr: DB errors can happen for multiple reasons. We learned the hard way that you always guard against all possible error and don't trust any subsystem blindly. Also, be extra careful when dealing with DB as it is very easy to have inconsistent states. 
You are only thinking of DB in process where only CTX matters. This DB is a whole system of its own and any transient error on DB server side can result in Exec errors.

PR #1222, opened 2026-05-14
fix: only update provider details for new provider in upsert

LexLuthr:
Follow up on #1221 to fix the intended behaviour and clarify the same with a comment. This was more incorrect than I initially understood.
List of things fixed in this PR:

1. retrievedAt now uses the oldest recorded ipni_ad_fetches.fetched_at.
2. Retrieved == true now forces Advertised == true.
3. AdvertisedAt for retrieved ads is capped at RetrievedAt using min(adCreatedAt + PublishInterval, retrievedAt).
4. Provider LastPublishTime is only used when there is no retrieval record.
5. Added a comment explaining the advertised timestamp derivation.

This should eliminate cases of when provider starts and LastPublishTime is nil. It should also give a consistent retrievedAt time.

### DIFF
--- a/market/ipni/ipni-provider/ipni-provider.go
+++ b/market/ipni/ipni-provider/ipni-provider.go
@@ -6,9 +6,11 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"maps"
 	"net/http"
 	"net/url"
 	"path"
+	"slices"
 	"strings"
 	"sync"
 	"time"
@@ -136,10 +138,10 @@ func NewProvider(d *deps.Deps) (*Provider, error) {
 	return p, nil
 }
 
-// upsertProvider inserts new providers and updates the providerInfos map
+// insertProvider inserts new providers and updates the providerInfos map
 // It does not change the details of the existing provider as all detailed are tied to a private key.
 // To remove an existing provider, a restart of all market nodes is required.
-func (p *Provider) upsertProvider(priv []byte, peerID string, sp int64) error {
+func (p *Provider) insertProvider(priv []byte, peerID string, sp int64) error {
 	pkey, err := crypto.UnmarshalPrivateKey(priv)
 	if err != nil {
 		return xerrors.Errorf("unmarshaling private key: %w", err)
@@ -187,6 +189,7 @@ func (p *Provider) upsertProvider(priv []byte, peerID string, sp int64) error {
 		info.providerType = PoRepProviderType
 	}
 
+	// Let's double-check before we update
 	p.mu.Lock()
 	_, existed := p.providerInfos[peerID]
 	if !existed {
@@ -205,6 +208,10 @@ func (p *Provider) refreshProviders(ctx context.Context) error {
 	}
 	defer rows.Close()
 
+	p.mu.RLock()
+	peers := slices.Sorted(maps.Keys(p.providerInfos))
+	p.mu.RUnlock()
+
 	for rows.Next() && rows.Err() == nil {
 		var priv []byte
 		var peerID string
@@ -213,7 +220,10 @@ func (p *Provider) refreshProviders(ctx context.Context) error {
 			return xerrors.Errorf("failed to scan refreshed ipni peer row: %w", err)
 		}
 
-		if err := p.upsertProvider(priv, peerID, sp); err != nil {
+		if lo.Contains(peers, strings.TrimSpace(peerID)) {
+			continue
+		}
+		if err := p.insertProvider(priv, peerID, sp); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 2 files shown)


magik6k (approved):

beck-8:
Looks fine.
But I just happened to find a situation here that I wanted to deal with here.

After reboot, before new transactions come. `advertised` status is lost. Or let's check if adCreatedAt has a value, or check the `retrieved` status?
```
{
  "pieceCid": "bafkzcibccybqydervgx5wwjigwvagiz2u732og7kfbjdw7oxbixg22hd7wc7qhi",
  "status": "retrieved",
  "indexed": true,
  "indexedAt": "2026-05-14T20:50:42.195625+08:00",
  "adCreated": true,
  "adCreatedAt": "2026-05-14T20:51:00.258678+08:00",
  "advertised": false,
  "retrieved": true,
  "retrievedAt": "2026-05-14T21:00:09.572079+08:00"
}
```

LexLuthr:
> Looks fine. But I just happened to find a situation here that I wanted to deal with here.
> 
> After reboot, before new transactions come. `advertised` status is lost. Or let's check if adCreatedAt has a value, or check the `retrieved` status?
> 
> ```
> {
>   "pieceCid": "bafkzcibccybqydervgx5wwjigwvagiz2u732og7kfbjdw7oxbixg22hd7wc7qhi",
>   "status": "retrieved",
>   "indexed": true,
>   "indexedAt": "2026-05-14T20:50:42.195625+08:00",
>   "adCreated": true,
>   "adCreatedAt": "2026-05-14T20:51:00.258678+08:00",
>   "advertised": false,
>   "retrieved": true,
>   "retrievedAt": "2026-05-14T21:00:09.572079+08:00"
> }
> ```

We had a incorrect conditional. Latest commit should fix that.

beck-8:
I thought this place was intentional.

But this does not solve the problem of state loss after restarting.

pdp/handlers.go
```diff
@@ -352,18 +360,58 @@ func (p *PDPService) handleGetPieceStatus(w http.ResponseWriter, r *http.Request
...
+	//    This keeps AdvertisedAt from appearing after RetrievedAt.
+	// 2. The in-process IPNI provider exposes LastPublishTime per provider, not
+	//    per ad. It is useful only when there is no fetch record and we know
+	//    when this ad was created. A provider publish after this ad was created
+	//    means the ad should have been included in the announced head; an older
+	//    provider publish means it was not, and we do not fall through to the
+	//    timing heuristic.
+	// 3. Without either signal, fall back to the old timing heuristic: after
+	//    PublishInterval has elapsed from ad creation, assume the ad was
+	//    announced.
```
ZenGround0, line 378: Thanks for the comment it is very clarifying and helpful

market/ipni/ipni-provider/ipni-provider.go
```diff
@@ -136,6 +136,9 @@ func NewProvider(d *deps.Deps) (*Provider, error) {
 	return p, nil
 }
 
+// upsertProvider inserts new providers and updates the providerInfos map
+// It does not change the details of the existing provider as all detailed are tied to a private key.
+// To remove an existing provider, a restart of all market nodes is required.
```
ZenGround0, line 143: nit: now the name "upsert" is a bit of a misnomer as we never do any "up"ing, only inserting if no provider.  So maybe favor `insertProvider` or `insertProviderIfNone`.  Not a strong preference though.

ZenGround0 (commented):
LGTM

pdp/handlers.go
```diff
@@ -352,18 +360,58 @@ func (p *PDPService) handleGetPieceStatus(w http.ResponseWriter, r *http.Request
...
-			response.Advertised = true
-			response.AdvertisedAt = new(*publishedAt)
-			if publishedAt.After(time.Now().Add(ipni_provider.PublishInterval)) {
-				response.AdvertisedAt = new(result.AdvertisementCreatedAt.Time.Add(ipni_provider.PublishInterval))
+		if publishedAt != nil {
+			advertisedFromProvider = true
+			if publishedAt.After(result.AdvertisementCreatedAt.Time) {
+				response.Advertised = true
+				response.AdvertisedAt = new(*publishedAt)
+				if publishedAt.After(time.Now().Add(ipni_provider.PublishInterval)) {
```
beck-8: Is it never possible to execute here? @LexLuthr 
I'm not so sure

beck-8:
Except for the review behavior that needs to be confirmed, other aspects performed well in my environment and were in line with expectations.

market/ipni/ipni-provider/ipni-provider.go
LexLuthr: renamed to insertProvider

PR #1224, opened 2026-05-14
fix: alert bugs

LexLuthr:
1. Fix incorrect windowPost message missing alert
2. Fix alert mute
3. Allow deleting mute rules
4. Fix message pending execution alert


magik6k (approved):
Seems quite a bit more correct

PR #1232, opened 2026-05-18
Move C1 check to FFISelect

LexLuthr:
This PR moves Curio’s C1 validation checks behind the existing short-lived curio ffi process boundary.
TreeRC already runs PC2 through ffiselect, but the follow-up C1 “checking PreCommit” loop still called ffi.SealCommitPhase1 directly from the long-running parent process. SyntheticProofs had the same direct C1 validation path. 

Those calls can leave large freed native/Rust allocations retained in the parent process by glibc arenas, causing idle RSS to stay high after sealing work completes.

The change adds CheckSealCommitPhase1 to ffiselect and routes both TreeRC and SyntheticProofs C1 validation loops through it. The C1 proof bytes are discarded in the child, preserving validation behavior while keeping native allocator retention out of the parent.

### DIFF
--- a/lib/ffiselect/ffidirect/ffi-direct.go
+++ b/lib/ffiselect/ffidirect/ffi-direct.go
@@ -53,6 +53,22 @@ func (FFI) SealPreCommitPhase2(
 	}, nil
 }
 
+func (FFI) CheckSealCommitPhase1(
+	proofType abi.RegisteredSealProof,
+	sealedCID cid.Cid,
+	unsealedCID cid.Cid,
+	cacheDirPath string,
+	sealedSectorPath string,
+	sectorNum abi.SectorNumber,
+	minerID abi.ActorID,
+	ticket abi.SealRandomness,
+	seed abi.InteractiveSealRandomness,
+	pieces []abi.PieceInfo,
+) error {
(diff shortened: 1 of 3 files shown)


lib/ffiselect/ffidirect/ffi-direct.go
```diff
@@ -53,6 +53,22 @@ func (FFI) SealPreCommitPhase2(
 	}, nil
 }
 
+func (FFI) CheckSealCommitPhase1(
+	proofType abi.RegisteredSealProof,
```
ZenGround0, line 57: How is this building without a ctx?  Is building this file not checked by the compiler?

ZenGround0 (approved):
Im confused about the lack of ctx and how things are building.  Other than that this lgtm so I'll leave approved assuming you either fix that or confirm it is not a problem.

snadrus (approved):

snadrus:
> Im confused about the lack of ctx and how things are building.  Other than that this lgtm so I'll leave approved assuming you either fix that or confirm it is not a problem.

I think ctx is only for the process killer. We don't have it in the subprocess because we don't ferry anything through that way nor back

Issue #1236, opened 2026-05-20
PDP: Update pdpv0 termination cleanup state machine for PDPVerifier cleanup mode


rvagg:
Ref: https://www.notion.so/filecoindev/Data-Set-Terminations-Clean-up-360dc41950c1801ebf0aff017a322e7f

The pdpv0 termination pipeline must be updated for the PDPVerifier cleanup-deposit changes in https://github.com/FilOzone/pdp/pull/270.

Current behaviour assumes that after FWSS termination, Curio can call PDPVerifier `deleteDataSet`, observe the data set is not live, then delete local data set rows and enqueue local piece cleanup. With the new contract behaviour, `deleteDataSet` may only enter PDP cleanup mode. Curio must then call `cleanupPieces(setId, maxPieces)` until PDP cleanup is complete and the 0.1 FIL cleanup deposit is paid out.

Scope:

- Update only pdpv0 flow (`tasks/pdpv0`, `pdp_delete_data_set`, related watchers).
- Preserve the order:
  1. FWSS `terminateService`
  2. Confirm service termination epoch
  3. Confirm PDP payment rail finality / inactive-or-settled
  4. PDPVerifier `deleteDataSet`
  5. PDPVerifier `cleanupPieces` until complete, if pieces remain
  6. Local data set deletion / piece cleanup / IPNI/index cleanup
- Do not delete local data set rows immediately after `deleteDataSet` if the PDPVerifier data set is in cleanup mode.
- Handle zero-piece data sets: `deleteDataSet` should finalise cleanup immediately; do not schedule `cleanupPieces`.
- Cleanup can be async, but [client-authorised termination](https://github.com/filecoin-project/curio/issues/1028) can zero the lockup period and finalise quickly. Curio should not leave this path waiting on the normal coarse settlement cadence before it even attempts to progress deletion.
- Keep rail finality authoritative before PDP deletion. Client termination may be fast, but Curio still must not skip FWSS/FilecoinPay finalisation checks.
- Tighten the termination-to-deletion latency path: once FWSS termination is confirmed, Curio should promptly attempt/check PDP rail finalisation and move to `deletion_allowed` as soon as FilecoinPay state permits. Existing periodic settlement/payment watcher/chain-sync paths should remain backstops, not the expected fast path.
- Retain detection/recovery for externally initiated termination/deletion, and extend it through the new PDP cleanup mode. This must not depend only on termination initiated through Curio’s HTTP endpoint.
[...]

Issue #1235, opened 2026-05-20
PDP: Implement pdpv0 client termination endpoint


rvagg:
_Replaces https://github.com/filecoin-project/curio/issues/1028_

Ref: https://www.notion.so/filecoindev/Data-Set-Terminations-Clean-up-360dc41950c1801ebf0aff017a322e7f

Curio currently has a data set delete route stub in `pdp/handlers.go`, but this should become client-requested termination, not direct deletion. Deletion is an internal Curio/PDP cleanup operation after service termination and payment finalisation.

Scope:

- Add a client-facing terminate endpoint for pdpv0 data sets, `POST /pdp/data-sets/{id}/terminate`.
- Implement/replace the existing handler stub path in `pdp/handlers.go` (currently `handleDeleteDataSet`).
- Verify the data set exists and belongs to the requesting service.
- **Require non-empty** `extraData` for client-requested termination.
- No needt to validate EIP-712 in Curio. Pass `extraData` through unchanged; FWSS/PDPVerifier remains authoritative for verifying the client signature.
- Empty `extraData` should remain an internal/provider termination path only, not accepted as a successful client termination request.
- Submit the FWSS termination transaction via the existing pdpv0 termination path and return the transaction hash, matching other PDP endpoints that submit transactions.
- Return `201 Created`/similar with a `Location` header carrying the tx hash, matching create/addPieces style.
- Add a termination status endpoint, e.g. `GET /pdp/data-sets/{id}/terminated/{txHash}`.
- The status endpoint should report tx status/success, whether FWSS termination has been confirmed, and the service termination epoch when known.
- Make repeated calls idempotent while termination is already pending or complete.

LexLuthr:
@rvagg I need a clarification - do we respond to `GET /pdp/data-sets/{id}/terminated/{txHash}` in perpetuity or just while we are handling the termination and deletion. Perpetuity means, one more table keeping track of this since we clean up deleted dataSet from their table.

rvagg:
[...]

Issue #940, opened 2026-01-30
pdpv0: update message_waits_eth to handle stuck messages


ZenGround0:
See discussion here: https://github.com/filecoin-project/curio/pull/937#issuecomment-3825478640.

BigLep:
2026-04-09 conversation: 
- we have refocused this to be about pruning old messages
- We have put it post-GA unless it comes up again before then.
- We discussed about doing this in a migration, but agreed this should be "another cleanup task" (look for other db cleanups) that we can model from.

ZenGround0:
@snadrus I know you mentioned that there were broader efforts to prune tables in curio more systematically.  If we have any issues for that we should link them here.

snadrus:
@ZenGround0 we need a ticket for unbounded db table growth in pdpV0, then the same for PoRep. However people have been running PoRep for years without issue (yet), so these tickets are low priority

PR #1239, opened 2026-05-20
fix: pdpv0 deletion pipeline

LexLuthr:
In prepreration for implementing #1236  and #1235 , I found a few problems in current termination and deletion pipeline.

## Summary
• Fixes TerminateFWSSTask so it records an already-terminated FWSS dataset locally and exits instead of submitting another terminate transaction.
• Records FWSS terminate txs and their message_waits_eth rows atomically.
• Updates terminate and delete watchers to avoid joining against message_waits_eth; they now load pipeline rows first, fetch wait rows by tx hash, then group success/failure states in memory.
• Resets failed terminate/delete tx pipeline rows so the existing schedulers can retry them.
• Adds TaskChainSync reconciliation for rows with tx hashes but missing message_waits_eth rows, using current contract state instead of tx lookup.
• Keeps task-name constants consistent by moving Settle into tasknames.
• Adds TODOs for the upcoming post-delete cleanup stage and future FWSS ABI update.

### DIFF
--- a/tasks/pdpv0/task_chain_sync.go
+++ b/tasks/pdpv0/task_chain_sync.go
@@ -44,14 +44,14 @@ func (t *TaskChainSync) Do(taskID harmonytask.TaskID, stillOwned func() bool) (d
 		return false, xerrors.Errorf("syncing stale PDP deletion task ids: %w", err)
 	}
 
-	if !stillOwned(){
+	if !stillOwned() {
 		return false, nil
 	}
 	if err := t.syncMissingDataSetTerminationMessageWaits(ctx); err != nil {
 		return false, xerrors.Errorf("syncing missing PDP termination message waits: %w", err)
 	}
 
-	if !stillOwned(){
+	if !stillOwned() {
 		return false, nil
 	}
 	if err := t.syncMissingDataSetDeleteMessageWaits(ctx); err != nil {
--- a/tasks/pdpv0/task_chain_sync.go
+++ b/tasks/pdpv0/task_chain_sync.go
@@ -44,11 +44,18 @@ func (t *TaskChainSync) Do(taskID harmonytask.TaskID, stillOwned func() bool) (d
 		return false, xerrors.Errorf("syncing stale PDP deletion task ids: %w", err)
 	}
 
-	if !stillOwned() {
+	if !stillOwned(){
 		return false, nil
 	}
-	if err := t.syncMissingDeletionMessageWaits(ctx); err != nil {
-		return false, xerrors.Errorf("syncing missing PDP deletion message waits: %w", err)
+	if err := t.syncMissingDataSetTerminationMessageWaits(ctx); err != nil {
+		return false, xerrors.Errorf("syncing missing PDP termination message waits: %w", err)
+	}
+
+	if !stillOwned(){
+		return false, nil
+	}
+	if err := t.syncMissingDataSetDeleteMessageWaits(ctx); err != nil {
+		return false, xerrors.Errorf("syncing missing PDP delete message waits: %w", err)
 	}
 
 	if !stillOwned() {
@@ -151,17 +158,7 @@ type missingDeleteMessageWait struct {
 	TxHash string `db:"delete_tx_hash"`
 }
 
-// syncMissingDeletionMessageWaits repairs deletion pipeline rows whose tx hash was recorded
-// without a matching message_waits_eth row. The tx may no longer be available from the
-// connected node, so this reconciles from contract state instead of tx history.
-func (t *TaskChainSync) syncMissingDeletionMessageWaits(ctx context.Context) error {
-	if err := t.syncMissingTerminationMessageWaits(ctx); err != nil {
-		return err
-	}
-	return t.syncMissingDeleteMessageWaits(ctx)
-}
-
-func (t *TaskChainSync) syncMissingTerminationMessageWaits(ctx context.Context) error {
+func (t *TaskChainSync) syncMissingDataSetTerminationMessageWaits(ctx context.Context) error {
 	var missing []missingTerminationMessageWait
 	if err := t.db.Select(ctx, &missing, `
 		SELECT id, terminate_tx_hash
@@ -239,13 +236,15 @@ func (t *TaskChainSync) syncMissingTerminationMessageWaits(ctx context.Context) 
 		}
 		if n == 1 {
 			log.Warnw("reset service termination missing message wait for retry", "dataSetId", detail.ID, "txHash", detail.TxHash)
+		} else {
+			continue
 		}
 	}
 
 	return nil
 }
 
-func (t *TaskChainSync) syncMissingDeleteMessageWaits(ctx context.Context) error {
+func (t *TaskChainSync) syncMissingDataSetDeleteMessageWaits(ctx context.Context) error {
 	var missing []missingDeleteMessageWait
 	if err := t.db.Select(ctx, &missing, `
 		SELECT id, delete_tx_hash
@@ -306,6 +305,8 @@ func (t *TaskChainSync) syncMissingDeleteMessageWaits(ctx context.Context) error
 		}
 		if n == 1 {
 			log.Warnw("reset data set delete missing message wait for retry", "dataSetId", detail.ID, "txHash", detail.TxHash)
+		} else {
+			continue
 		}
 	}
 
--- a/tasks/pdpv0/task_chain_sync.go
+++ b/tasks/pdpv0/task_chain_sync.go
@@ -47,6 +47,13 @@ func (t *TaskChainSync) Do(taskID harmonytask.TaskID, stillOwned func() bool) (d
 	if !stillOwned() {
 		return false, nil
 	}
+	if err := t.syncMissingDeletionMessageWaits(ctx); err != nil {
+		return false, xerrors.Errorf("syncing missing PDP deletion message waits: %w", err)
+	}
+
+	if !stillOwned() {
+		return false, nil
+	}
 	if err := t.syncProvenDataSetFailureState(ctx); err != nil {
 		return false, xerrors.Errorf("syncing proven PDP data set failure state: %w", err)
 	}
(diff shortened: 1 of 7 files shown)


tasks/pdpv0/watch_fwss_terminate.go
```diff
@@ -95,3 +147,25 @@ func processPendingTerminations(ctx context.Context, db *harmonydb.DB, ethClient
...
+			  AND terminate_tx_hash = $2
+			  AND after_terminate_service = TRUE
+			  AND service_termination_epoch IS NULL
+		`, detail.DataSetId, detail.TxHash)
+		if err != nil {
+			return xerrors.Errorf("failed to reset failed service termination for data set %d: %w", detail.DataSetId, err)
+		}
+
+		log.Warnw("reset failed provider service termination for retry", "dataSetId", detail.DataSetId, "txHash", detail.TxHash)
+	}
```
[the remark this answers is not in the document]
LexLuthr: As usual, incorrect review without context.

tasks/pdpv0/watch_data_set_delete.go
```diff
@@ -79,60 +131,93 @@ func processPendingDeletes(ctx context.Context, db *harmonydb.DB, ethClient ethc
...
+		`, detail.ID, detail.TxHash)
 		if err != nil {
-			return xerrors.Errorf("failed to commit transaction: %w", err)
+			return xerrors.Errorf("failed to reset failed data set delete for data set %d: %w", detail.ID, err)
 		}
-		if !comm {
-			return xerrors.Errorf("failed to commit transaction")
+
+		log.Warnw("reset failed data set delete for retry", "dataSetId", detail.ID, "txHash", detail.TxHash)
+	}
```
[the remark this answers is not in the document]
LexLuthr: As usual, incorrect review without context.

tasks/pdpv0/watch_data_set_delete.go
```diff
@@ -79,60 +131,93 @@ func processPendingDeletes(ctx context.Context, db *harmonydb.DB, ethClient ethc
...
+
+		return true, nil
+	}, harmonydb.OptionRetry())
+
+	if err != nil {
+		return xerrors.Errorf("failed to commit transaction: %w", err)
+	}
+	if !comm {
+		return xerrors.Errorf("failed to commit transaction")
 	}
```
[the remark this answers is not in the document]
LexLuthr: Copilots opinion not a review.

tasks/pdpv0/task_delete_data_set.go
```diff
@@ -64,6 +64,9 @@ func (t *DeleteDataSetTask) Do(taskID harmonytask.TaskID, stillOwned func() bool
 	}
 
 	if !live {
+		// TODO: When the post-delete cleanup stage lands, route already-deleted. Currently, this missed the DB cleanup
+		// TODO: In the upcoming PR to handle https://github.com/filecoin-project/curio/issues/1236 this will be fixed
+		// data sets into that stage instead of marking this pipeline terminal here.
```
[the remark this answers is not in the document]
LexLuthr: Again. Not really a review.

tasks/pdpv0/watch_data_set_delete.go
```diff
@@ -79,60 +131,93 @@ func processPendingDeletes(ctx context.Context, db *harmonydb.DB, ethClient ethc
...
+		_, err = tx.Exec(`DELETE FROM pdp_data_sets WHERE id = $1`, dataSetID)
+		if err != nil {
+			return false, xerrors.Errorf("failed to delete data set %d: %w", dataSetID, err)
+		}
+
+		_, err = tx.Exec(`DELETE FROM  pdp_delete_data_set WHERE id = $1 AND delete_tx_hash = $2`, dataSetID, deleteTxHash)
+		if err != nil {
+			return false, xerrors.Errorf("failed to delete row from pdp_delete_data_set: %w", err)
 		}
+
```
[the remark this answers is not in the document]
LexLuthr: DELETE should not be guarded by number of rows unless you can guarantee atomicity across the cluster. This is on best effort basis.

tasks/pdpv0/task_chain_sync.go
```diff
@@ -134,6 +141,177 @@ func (t *TaskChainSync) syncStaleDeletionTaskIDs(ctx context.Context) error {
 	return nil
 }
 
+type missingTerminationMessageWait struct {
+	ID     int64  `db:"id"`
+	TxHash string `db:"terminate_tx_hash"`
+}
+
+type missingDeleteMessageWait struct {
```
ZenGround0, line 156: nit: Probably useful to specify dataset deletion to disambiguate from piece deletion

tasks/pdpv0/task_chain_sync.go
```diff
@@ -134,6 +141,177 @@ func (t *TaskChainSync) syncStaleDeletionTaskIDs(ctx context.Context) error {
...
+	ID     int64  `db:"id"`
+	TxHash string `db:"delete_tx_hash"`
+}
+
+// syncMissingDeletionMessageWaits repairs deletion pipeline rows whose tx hash was recorded
+// without a matching message_waits_eth row. The tx may no longer be available from the
+// connected node, so this reconciles from contract state instead of tx history.
```
ZenGround0: 👍

tasks/pdpv0/task_chain_sync.go
```diff
@@ -134,6 +141,177 @@ func (t *TaskChainSync) syncStaleDeletionTaskIDs(ctx context.Context) error {
...
+			if err != nil {
+				return xerrors.Errorf("failed to reconcile terminated data set %d: %w", detail.ID, err)
+			}
+			if n > 1 {
+				return xerrors.Errorf("expected to update 0 or 1 rows for data set %d, updated %d", detail.ID, n)
+			}
+			if n == 1 {
+				log.Infow("reconciled missing service termination message wait from chain state", "dataSetId", detail.ID, "txHash", detail.TxHash, "epoch", ds.PdpEndEpoch.Int64())
+			}
+			continue
```
ZenGround0, line 218: nit: I'd prefer an if else to if {continue}

tasks/pdpv0/task_chain_sync.go
```diff
@@ -134,6 +141,177 @@ func (t *TaskChainSync) syncStaleDeletionTaskIDs(ctx context.Context) error {
...
+
+type missingDeleteMessageWait struct {
+	ID     int64  `db:"id"`
+	TxHash string `db:"delete_tx_hash"`
+}
+
+// syncMissingDeletionMessageWaits repairs deletion pipeline rows whose tx hash was recorded
+// without a matching message_waits_eth row. The tx may no longer be available from the
+// connected node, so this reconciles from contract state instead of tx history.
+func (t *TaskChainSync) syncMissingDeletionMessageWaits(ctx context.Context) error {
```
ZenGround0: nit: I'd probably prefer just calling these two internal methods inline in the Do loop.  The added function wrapping them doesn't add any value and the naming is awkward.  

Random design heuristic: sometimes awkward naming is nice because it is a clue that an abstraction is unneeded.

tasks/pdpv0/task_delete_data_set.go
ZenGround0, line 69: 👍

tasks/pdpv0/task_chain_sync.go
```diff
@@ -134,6 +141,177 @@ func (t *TaskChainSync) syncStaleDeletionTaskIDs(ctx context.Context) error {
...
+		if err != nil {
+			return xerrors.Errorf("failed to check if data set %d is live: %w", detail.ID, err)
+		}
+
+		if !live {
+			if err := cleanupDeletedDataSet(ctx, t.db, detail.ID, detail.TxHash); err != nil {
+				return xerrors.Errorf("failed to reconcile deleted data set %d: %w", detail.ID, err)
+			}
+			log.Infow("reconciled missing data set delete message wait from chain state", "dataSetId", detail.ID, "txHash", detail.TxHash)
+			continue
```
ZenGround0, line 286: nit: same if else preference

tasks/pdpv0/task_terminate_fwss.go
```diff
@@ -90,6 +92,8 @@ func (t *TerminateFWSSTask) Do(taskID harmonytask.TaskID, stillOwned func() bool
 		return false, xerrors.Errorf("failed to get FWSS ABI: %w", err)
 	}
 
+	// TODO: Regenerate the FWSS ABI for terminateService(uint256,bytes)
+	// and pass empty extraData here for provider-initiated termination.
```
ZenGround0, line 96: 👍

tasks/pdpv0/watch_data_set_delete.go
```diff
@@ -28,48 +28,100 @@ func NewDataSetDeleteWatcher(db *harmonydb.DB, ethClient ethchain.EthClient, pcs
...
+		log.Warnw("data set delete tx missing message_waits_eth row", "txHash", detail.TxHash, "dataSetId", detail.ID)
+	}
+
+	successErr := processSuccessfulDeletes(ctx, db, ethClient, successes)
+	failureErr := processFailedDeletes(ctx, db, failures)
+	if successErr != nil {
+		return successErr
+	}
+	return failureErr
+}
```
ZenGround0: If either error is non nil we should return a concatenation of the two.  We don't want to swallow failureErr

tasks/pdpv0/watch_fwss_terminate.go
```diff
@@ -28,27 +39,77 @@ func NewTerminateServiceWatcher(db *harmonydb.DB, ethClient ethchain.EthClient,
...
+
+	if len(pending) == 0 {
+		return nil
+	}
 
-	var details []struct {
-		DataSetId int64        `db:"id"`
-		TxHash    string       `db:"terminate_tx_hash"`
-		Success   sql.NullBool `db:"tx_success"`
+	byHash := make(map[string]pendingServiceTermination, len(pending))
```
ZenGround0, line 59: I guess this and the thing above are related to:

> Updates terminate and delete watchers to avoid joining against message_waits_eth; they now load pipeline rows first, fetch wait rows by tx hash, then group success/failure states in memory.

Is the motivation just for efficiency because we know that table is big?   Or am I mistaken and it is actually bad practice to join tables in sql?

ZenGround0: I am ok with this as a stop gap for now if you have a strong opinion that left joining is untenably slow but we really should go about fixing the underlying cause ( #940 ).

tasks/pdpv0/watch_fwss_terminate.go
```diff
@@ -28,27 +39,77 @@ func NewTerminateServiceWatcher(db *harmonydb.DB, ethClient ethchain.EthClient,
...
+	hashes := make([]string, 0, len(pending))
+	for _, detail := range pending {
+		hashes = append(hashes, detail.TxHash)
+		byHash[detail.TxHash] = detail
```
ZenGround0, line 63: As long as we are doubling down on doing this grouping outside of sql queries we should probably have a little library utility that does this group by hash thing sorting into successes and failure in a generic way.

ZenGround0 (approved):
Thanks!  The TODOs are ok to leave as they are.

tasks/pdpv0/watch_fwss_terminate.go
LexLuthr: I don't believe it is slow. We know big clusters who are running similar JOINs all over. This was meant for efficiency and being able to manipulate some things on Go side. Since the number of watcher is going, I am trying to be as efficient as possible. My fears might be completely unjustified gives the size of some clusters we are aware of.

tasks/pdpv0/watch_fwss_terminate.go
LexLuthr: Yes, I am planning to create a little helper lib with a bunch of pdpv0 funcs spread all over in one place. That would be the finalization PR after all the functional changes.

Issue #1241, opened 2026-05-20
pdpv0: PullPiece backlog grows unboundedly under sustained pull load; pieces lost silently and on restart — node paralysed


TippyFlitsUK:
**Environment**
Mainnet PDP node, curio `pdpv0` branch (v1.28.0-rc1). One curio process, layers `gui,pdp`. `/filecoin-sealing` on local NVMe, `/filecoin-storage` mounted via NFS.

**What I'm seeing**

Under sustained `POST /pdp/piece/pull` traffic from a Storacha migration (~640 items every 10 minutes, steady for hours), the node has gone into a bad state:

1. **PullPiece queue is growing for hours and won't drain.** Currently 5,500+ pull items pending. Arrival rate is faster than what the node can process, and the gap doesn't close.

2. **Silent server-side data loss.** `harmony_task_history` shows zero `PDPv0_PullPiece` failures, but `pdp_piece_pull_items.failed = true` with `fail_reason = 'pull budget exceeded'` has accumulated 1,373 items — 1,300 in a single hour. These never become harmony tasks, so there's no operator-visible signal that they failed. The synapse-sdk client just sees pieces that never appear.

3. **Restart appears to lose in-flight pulls.** After curio restarts, I saw 78 `StorePiece` tasks stuck at retries 8–9 of 10, all failing with `"no suitable data URL found for piece_id <N>"`. The `parked_piece_refs` rows pointed to stash UUIDs that no longer exist in `/filecoin-sealing/.pdp/stash/`. Net effect: pulled pieces were downloaded, written to stash, and then thrown away on restart — so the client re-POSTs them, and the queue grows again.

4. **Retry storms freeze the scheduler.** While those 78 doomed `StorePiece` tasks cycled through their retries (~10 minutes total), `PullPiece` throughput dropped to near zero in monitoring samples even though the pull queue was 5,000+ deep and the box was at ~30% CPU.

5. **External impact.** nfts2me (the migration source) reports timeouts on `https://main.ezpdpz.net`.

**Impact on the node**

The pull backlog has **paralysed the entire task queue**, not just pull traffic:

- All harmony task types compete for the same scheduler attention. When PullPiece and StorePiece are both backlogged (which they are continuously under this load), other PDP tasks — `PDPv0_Notify`, `PDPv0_SaveCache`, `PDPv0_Indexing`, `PDPv0_Prove`, `SendTransaction` — also stall or run at a fraction of normal throughput.
[...]

PR #1242, opened 2026-05-21
feat: bound PullPiece/StorePiece concurrency, drop CPU reservation

rvagg:
PullPiece was hardcoded to 4; StorePiece had no cap (MoveStorageMaxTasks default 0). Add PDPPullPieceMaxTasks (default 8) and StorePieceMaxTasks (default 12) so both are configurable independently.

Both tasks are IO-bound; the 1-CPU reservation overstated real load and let StorePiece starve Prove and other critical tasks under sustained pull load. Set Cost.Cpu=0 so Max is the honest concurrency dial. ParkPiece (sealing pipeline) keeps its 1-CPU reservation since it shares slots with sealing tasks that do meaningful CPU work.

Ref: filecoin-project/curio#1241

### DIFF
--- a/api/proxy_gen.go
+++ b/api/proxy_gen.go
@@ -5,8 +5,16 @@ package api
 import (
 	"context"
 	"fmt"
-	ltypes "github.com/filecoin-project/curio/api/types"
-	"github.com/filecoin-project/curio/lib/storiface"
+	"net/http"
+	"net/url"
+	"reflect"
+
+	"github.com/google/uuid"
+	blocks "github.com/ipfs/go-block-format"
+	"github.com/ipfs/go-cid"
+	"github.com/multiformats/go-multihash"
+	"golang.org/x/xerrors"
+
 	"github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-bitfield"
 	"github.com/filecoin-project/go-jsonrpc/auth"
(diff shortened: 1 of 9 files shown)


TippyFlitsUK:
Deployed `6cd536e4` on the mainnet PDP node (.80) at 01:40 UTC.

Build: `curio version 1.28.1+mainnet+git_6cd536e4_2026-05-21T11:25:30+10:00`
Branch: `rvagg/store-pull-task-limits`
Defaults left as shipped (`PDPPullPieceMaxTasks=8`, `StorePieceMaxTasks=12`)

### What's working

- Caps are binding as designed: PullPiece inflight saturates at 8, StorePiece at 12 (verified continuously across the last 15 min of `harmony_task` snapshots)
- Cross-task starvation is gone. Last 5 min showed 639 task completions across 9 task types:

| Task | Done (5min) |
|---|---|
| StorePiece | 296 |
| PDPv0_PullPiece | 136 |
| PDPv0_Notify | 110 |
| PDPv0_SaveCache | 60 |
| PDPNotify | 24 |
| SendTransaction | 9 |
| PDPv0_Indexing | 2 |
| PDPv0_IPNI | 1 |

  Previously under the same workload, Notify/SaveCache/Indexing/SendTransaction were near-zero while Pull/Store monopolised everything. Dropping `Cost.Cpu` to 0 is what unblocked them — those tasks were being denied scheduler slots by the CPU reservation accounting, not by actual CPU contention.

- Stash-orphan retry cluster from earlier restarts cleared naturally (25 retries → 0 over ~10 min as `MaxFailures: 10` was reached and the doomed `StorePiece` tasks dropped off the queue).

### Downstream benefit

The host's Lotus daemon was previously eating ~784% CPU and continuously logging `"reorg channel is heavily backlogged"`. With Pull/Store now capped at 8/12, the burst disk IO they generate is much lower at any one moment, and Lotus has dropped to ~256% CPU with no reorg spam. The chain-side health restoration was a side effect we weren't specifically looking for, but it's significant for the overall node.

### Caveat

The Storacha migration source has been paused since ~01:00 UTC, so we haven't yet observed the patched node under live arrival pressure (the typical 640 items per 10-min batch). Will report back once nfts2me resumes — that's the real test of whether the 8/12 defaults are tight enough or whether they need raising for normal operation.

### Numbers in case useful

- Pull queue at install: ~3,750
- Pull queue 15 min later: ~3,370
- Net drain rate without arrivals: ~24/min
- Combined throughput across all task types: ~128/min
- Curio CPU: ~86% (~55 cores of 64); Lotus CPU: ~256% (~2.5 cores)
- Host: 64 cores / 503 GiB RAM, `/filecoin-sealing` on local NVMe, `/filecoin-storage` NFS-mounted to .50

ZenGround0 (approved):
LGTM

LexLuthr:
Can we hold off on merging this? Affected SPs can run this branch. I will be done with refactor tomorrow. I would like to avoid removing all this code and config again.

LexLuthr:
#1245 supersedes this

snadrus (approved):

rjan90:
Closing, since this is being superseded by: https://github.com/filecoin-project/curio/pull/1245

PR #1244, opened 2026-05-22
Integ/task

### LINKED ISSUE #1282: Idle DB load from web task polling and PDPv0 save-cache scheduler scans
## Summary

The database still shows intermittent 100-150% CPU usage even when there appears to be little active work.

I investigated active SQL while the Curio web UI was open and also after closing the web UI. The remaining load appears to come mainly from two sources:

1. Duplicate high-frequency `ClusterTaskSummary` polling on the web homepage.
2. `PDPv0_SaveCache` repeatedly scanning `pdp_piecerefs` even when no save-cache work exists.

## 1. Web homepage runs duplicate ClusterTaskSummary polling

The homepage renders two `<cluster-tasks>` components:

```html
<cluster-tasks></cluster-tasks>

<ui-drawer class="drawer" label="Cluster Tasks">
  <cluster-tasks slot="content"></cluster-tasks>
</ui-drawer>
```

Each `cluster-tasks` instance starts its own polling loop in the constructor:

```js
pollRPC(async () => {
  this.data = (await RPCCall('ClusterTaskSummary')) || [];
}, 1000);
```

So opening the homepage creates at least two `ClusterTaskSummary` RPC calls per second. Multiple browser tabs multiply this further.

The backend query is expensive on YugabyteDB:

```sql
SELECT
  t.id,
  t.name,
  t.update_time,
  t.owner_id,
  hm.host_and_port
FROM harmony_task t
LEFT JOIN harmony_machines hm ON hm.id = t.owner_id
ORDER BY CASE WHEN t.owner_id IS NULL THEN 1 ELSE 0 END, t.update_time ASC;
```

Observed execution plan:

```text
Seq Scan on harmony_task
rows=1
Execution Time: ~480 ms
```

Current cumulative stats:

```text
calls    = 333580
mean_ms  = 584.7
max_ms   = 54327.3
```

This means simply leaving the homepage open can generate sustained DB load.

## 2. PDPv0_SaveCache scheduler scans pdp_piecerefs every 3 seconds

`TaskPDPSaveCache` schedules every 3 seconds and runs:

```sql
SELECT id
FROM pdp_piecerefs
WHERE save_cache_task_id IS NULL
  AND needs_save_cache = TRUE
ORDER BY created_at ASC
LIMIT 1;
```

Current DB state:

```text
pdp_piecerefs total rows      = 87312
pending save-cache rows       = 0
```

But the query still performs a sequential scan:

```text
Seq Scan on pdp_piecerefs
Execution Time: ~226 ms
rows=0
```

Cumulative stats:

```text
calls    = 241732
mean_ms  = 271.7
max_ms   = 56253.8
total_ms = 65687837.7
```

There are pending indexes for related workflows:

```text
idx_pdp_piecerefs_indexing_pending
idx_pdp_piecerefs_ipni_pending
```

But there does not appear to be an equivalent pending index for save-cache scheduling.

## Expected behavior

- The web homepage should not create duplicate high-frequency polling loops for the same task summary.
- Hidden drawer content should not poll continuously if it is not visible, or the data should be shared between the visible table and drawer.
- `ClusterTaskSummary` polling should probably be slower, cached, or otherwise made cheaper.
- `PDPv0_SaveCache` should avoid full-table scans when there is no work, either by using an appropriate index or by backing off when the queue is empty.

snadrus:
Renovate the task system to start work faster, allow progress on multiple busy pipelines, and better select tasks. Also, complete winningPoST more reliably.

This is a set of LGTMed updates to the task system to:
- add a signaling mechanism (using http to Curio nodes) to express interesting events (new task, task success)
- Anti-Starvation to run the oldest available task's downstream oldest tasks. #1137
- a reduction in harmony_task polling frequency as signaling is sufficient
- a notification system for the other work-pollers that start tasks such that they reduce their polling frequency
- Moving DB operations out of the "harmony_task main thread" so the race-free resource allocation to tasks still happens but much faster. CanAccept inputs & outputs get cached
- Multiple tasks of a type can be started in one pass, reducing processing for backlogs. 
- Task cancellation so that "time-sensitive" tasks can be started immediately on busy hardware. 

New Concepts:
MayFollow: The ordering that indicates pipelines for purpose of starting tasks at the end of a pipeline first. 
TimeSensitive: A battering-ram that claims any resources it needs. 

Out-of-scope: 
- Racing 2+ Winning-PoST runners
- Dealing with starvation around big tasks never getting to run because new, small tasks overwhelm the system.


tasks/pdpv0/task_delete_data_set.go
```diff
@@ -143,8 +141,9 @@ func (t *DeleteDataSetTask) CanAccept(ids []harmonytask.TaskID, engine *harmonyt
 
 func (t *DeleteDataSetTask) TypeDetails() harmonytask.TaskTypeDetails {
 	return harmonytask.TaskTypeDetails{
-		Max:  taskhelp.Max(10),
-		Name: tasknames.PDPv0_DelDataSet,
+		Name:      tasknames.PDPv0_DelDataSet,
+		MayFollow: []string{tasknames.PDPv0_Prove},
```
ZenGround0: Also it maybe follows Prov_Period ?

tasks/pdpv0/task_prove.go
```diff
@@ -966,6 +964,11 @@ func (p *ProveTask) TypeDetails() harmonytask.TaskTypeDetails {
 	return harmonytask.TaskTypeDetails{
 		Name:          tasknames.PDPv0_Prove,
 		TimeSensitive: true,
+		// MayFollow must stay acyclic for harmonytask scheduling (see treehelper).
+		// PDPv0_ProvPeriod already follows PDPv0_Prove; do not list ProvPeriod here.
+		// Proving pipeline only: onboarding ends before PDPv0_InitPP (see task_init_pp);
+		// chain is PDPv0_InitPP → PDPv0_Prove → PDPv0_ProvPeriod.
```
ZenGround0, line 917: So does this mean we don't get symmetrically fast/cheap scheduling benefits for PDPv0_Prove ?

ZenGround0: I would guess it doesn't matter since challenge window duration >>> polling interval

ZenGround0 (commented):
I think we're ready to merge this and give it a try.  The only questions remaining are our testing approach and our sequencing wrt outstanding GA readiness bugfixes

snadrus:
@TippyFlitsUK This will need a few iterations of testing. The first one is likely to break rather quickly.

TippyFlitsUK:
I'm always up for breaking things @snadrus! 😉 

Standing by....

harmony/harmonytask/scheduler.go
```diff
@@ -0,0 +1,521 @@
...
+// The local implementation reads from the in-memory map; a DB-backed
+// implementation could be used for fallback scenarios.
+type taskSource interface {
+	GetTasks(taskName string) []task
+}
+
+func (e *TaskEngine) tryStartTask(taskName string, taskSource taskSource, eventEmitter eventEmitter) error {
+	h := e.taskMap[taskName]
+	if h != nil && h.TimeSensitive {
+		if _, capErr := h.AssertMachineHasCapacity(); capErr != nil {
```
LexLuthr: What if we have capacity over here? There is no else to proceed to scheduling.

harmony/harmonytask/task_type_handler.go
```diff
@@ -255,55 +278,78 @@ func (h *taskTypeHandler) considerWork(from string, ids []TaskID) (workAccepted
...
+				h.recordCompletion(tID, sectorID, workStart, done, doErr, preempted)
+				success := done && doErr == nil
+				if success {
+					h.TaskEngine.invokeTaskCompleteCallbacks(h.Name, meta, tID)
+				}
+				eventEmitter.EmitTaskCompleted(h.Name, success)
+				if !done {
+					for _, t := range tasks {
+						if t.ID == tID {
+							eventEmitter.EmitTaskNew(h.Name, t)
```
LexLuthr: I think we discussed in one of the other PRs as well. IIRC, you said it is okay to treat this as a fresh task. Now that I look at this again, that might be a problem. We bypass RetryWait completely. Effectively recreating the problem it was solving by introducing the delay to allow certain things to change before we try again. This will also likely mess with metrics.

cmd/curio/tasks/tasks.go
```diff
@@ -397,10 +415,139 @@ func StartTasks(ctx context.Context, dependencies *deps.Deps, shutdownChan chan
...
+			}
+			if parkPiecePoll != nil {
+				parkPiecePoll.WakePoll()
+			}
+			if storePiecePoll != nil {
+				storePiecePoll.WakePoll()
+			}
+		})
+	}
 	go machineDetails(dependencies, activeTasks, ht.ResourcesAvailable().MachineID, dependencies.Name)
```
LexLuthr: This should become synchronous now and should be either called from inside the harmonytask.New() or some similar approach. We start peering in NEW() immediately, for the new nodes, race condition can lead to us calling on a peer without row existing in DB. This will cause problem with stale peer. Basically, when this node send first message we handle it in `func (p *PeerHTTP) ServeHTTP(w http.ResponseWriter, r *http.Request) {}` and we store the connection. Then when we call `go p.onConnect(peerAddr, conn)` which will fail because of the missing row, we simply defer the connection close but we never actually delete the connection from our map. That means next time if details change, we still will accept tasks on this stale connection.

LexLuthr: Basically, we should have a way to remove peer from onConnect

```
conn, exists := p.connections[peerAddr]
	if !exists {
		conn = &peerHTTPConnection{
			parent:   p,
			peerAddr: peerAddr,
			incoming: make(chan []byte, 100),
		}
		p.connections[peerAddr] = conn
		p.connMu.Unlock()

		log.Infow("new peer connection via HTTP", "peer", peerAddr)

		if p.onConnect != nil {
			go p.onConnect(peerAddr, conn)
		}
	} else {
		p.co
```

harmony/harmonytask/harmonytask.go
```diff
@@ -265,47 +395,32 @@ func (e *TaskEngine) GracefullyTerminate() {
...
+// The function iterates handlers in registration order, filtering tasks by
+// retry-wait eligibility, then delegates to considerWork for CanAccept +
+// SQL claim + goroutine launch. Resource metrics are updated on exit.
+//
+// When capacity remains after processing all known work, IAmBored callbacks
+// are invoked in separate goroutines to avoid blocking the scheduler with
+// DB writes from speculative work creation.
+func (e *TaskEngine) pollerTryAllWork(taskSource taskSource, eventEmitter eventEmitter) error {
+	schedulable := !e.atomics.yieldBackground.Load()
+	if e.atomics.yieldBackground.Load() {
```
LexLuthr: But then we cannot schedule the overrides here. We are creating the var for it and then existing immediately without using the var to distinguish what should be and should not be scheduled. This would mainly hit finalize for SupraSeal. We will end up completing the task and not finalizing and requiring to start again.

```
// Allow finalize to be scheduled when batch tasks are still running even when the node is not schedulable.
// This allows finalize to unblock move-storage and PoRep for multiple hours while the node is technically not schedulable,
// but is still finishing another batch. In most cases this behavior enables nearly zero-waste restarts of supraseal nodes.
SchedulingOverrides: batchTaskNameGrid(),
```

harmony/harmonytask/scheduler.go
```diff
@@ -0,0 +1,521 @@
...
+						continue
+					}
+					plan := e.computePreemptionPlan(h.Cost)
+					if plan == nil {
+						log.Debugw("preemption plan no longer viable", "task", event.TaskType, "taskID", event.TaskID)
+						continue
+					}
+					e.executePreemption(plan)
+					tasks := taskSourceLocal{availableTasks}.GetTasks(event.TaskType)
+					if len(tasks) > 0 {
```
LexLuthr, line 305: Would it be better to check before we nuke the current work. That avoid us emptying out the queue when maybe someone else also did that and got the task.

LexLuthr (commented):
Noticed a few minor issues. Otherwise, looks solid and much easier to understand as a whole.

harmony/harmonytask/scheduler_test.go
LexLuthr, line 1: Delete this file?

LexLuthr (commented):
It seems like new itests are not wired to CI. They also using old harmonyDB setup methods.

tasks/pdpv0/task_prove.go
snadrus: on discussion, this doesn't establish any kind of strict ordering nor dependency, just how to handle overload, which shouldn't (hopefully) happen here.

tasks/pdpv0/task_delete_data_set.go
snadrus: done

harmony/harmonytask/scheduler.go
snadrus: ??? availableTasks is live-updated. Anyone else online has been passing notes to this node that updates availableTasks

harmony/harmonytask/scheduler.go
snadrus: oops, fixed

harmony/harmonytask/harmonytask.go
snadrus: That original code still exists below. This was an oops that shouldn't have been added

tasks/pdpv0/task_delete_data_set.go
snadrus: done

cmd/curio/tasks/tasks.go
snadrus: DONE: 
1. made this sync
2. always reconnect (no exists check)
3. Verified that dropPeer seems nicely wired for other peer-die cases.

harmony/harmonytask/task_type_handler.go
snadrus: Added a retrywait-respecting system.
This still goes through the pipeline: 
   EmitTaskNew() --> schedulerChannel<-scheddulerEvent{ Source: Added }
So I'm not sure which metrics are getting messed-with

LexLuthr: harmonytask_task_scheduled_wait_seconds, harmonytask_tasks_started indirectly and maybe 1 more. I don't remeber which one now. But your fix should fix all of it.

harmony/harmonytask/scheduler.go
LexLuthr: The issue is:
1. A time-sensitive task is seen as available, so this node starts preemption negotiation from tryStartTask at scheduler.go:263 / scheduler.go:333.
2. preemptForTimeSensitive later queues schedulerSourceStartTimeSensitive at preempt.go:181.
3. In between, the task can stop being locally available. That can happen from a peer started note deleting it at scheduler.go:278, or from a DB poll replacing the available snapshot at scheduler.go:193.
4. When schedulerSourceStartTimeSensitive is processed, it preempts first at scheduler.go:297 and scheduler.go:302.
5. It checks availableTasks only after preemption at scheduler.go:303.

The availableTasks is live-updated, but the code does not consult the live-updated map until after killing local work.

LexLuthr (commented):
https://github.com/filecoin-project/curio/pull/1244#pullrequestreview-4390263252 This is still not fixed

harmony/harmonytask/scheduler.go
LexLuthr: Still not fixed. Condition still returns early if we have the capacity.

LexLuthr: Also, can we skip canAccept cache for time sensitive task? In case cache has old ones we don't try new ones till cache is empty. Or am I reading it wrong?

harmony/harmonytask/harmonytask.go
LexLuthr: Now, we have another problem. We will reach below but then it won't work.

```go
for _, v := range oldestFirstSeq(e.taskMap, taskSource, e.cfg.preferredTaskRunOrder) {
		if !schedulable {
			for relatedTaskName := range v.SchedulingOverrides {
				if len(taskSource.GetTasks(relatedTaskName)) > 0 {
					goto doScheduling
				}
			}
			continue
		}
```

taskSource.GetTasks(relatedTaskName) -> We are already running the task, would it still exist in GetTasks? If not then this won't work.

snadrus: itest added that can ONLY succeed via preemption: TestTimeSensitivePathWhenResourcesOccupied

harmony/harmonytask/scheduler.go
snadrus: Thinking about the reverse: 
- CanAccept (cache) should inform the preemption system so we don't try running on storage machines that happen to have the task enabled.

And then, it's a cache. It's barely a detail compared to the SQL to claim the task.

LexLuthr (approved):

BigLep:
2026-06-17 conversation:

Next steps:
1. Need to test in pdp flows (will ask Tippy)
2. Need to test porep (don't have anyone on contract for this)

BigLep:
Slack thread about seeing if @TippyFlitsUK or @beck-8 can help with Curio PoRep testing in general: https://filecoinproject.slack.com/archives/C08TVNKJV7C/p1781821751061449

TippyFlitsUK:
Tested at `5ef34bb5` on two calibnet PDP nodes (YugabyteDB), following a ~24h clean baseline on v1.28.2-rc2 (zero task failures, 100% dealbot deals on both).

**Piece-to-IPNI-announce latency increased 10-40x, causing dealbot deal failures.** Measured stored-to-announce per piece (dealbot events x curio logs, same node, same day):
- Stable, 9 samples: 11-40s (median ~24s). 9/9 deals passed.
- This branch, 5 samples: 107s, 231s, 274s, 754s, 928s. The 107s deal passed; the four exceeding dealbot's IPNI verification window all failed.

**Localised to scheduling-pass cadence.** The task chain behind the announce (Notify -> Indexing -> IPNI) is created via IAmBored callbacks, which only run during scheduling passes. Identical task code in both builds; `pollRarely = 30s` vs the previous 3s poll is the changed variable. As a test we rebuilt with `pollRarely` set back to 3s on both nodes: announce latency returned to ~15s on the first measured pair and all 8 deals since have passed at stable-grade end-to-end times (~60-70s upload to completion).

One related question: does the `pollFrequently` fallback (peer-connection failure path) ever engage on a single-node deployment? It did not during our unpatched run.

**`Work not accepted ... already Taken` warnings, 500+/hour** on a single-node setup (zero on stable). The task runs via the winning claim path; the losing path logs the warning.

**`CanAccept() refused` for SendTransaction is new on this branch**: zero occurrences in weeks of stable logs, 31 across both nodes in the first hours here. Coincides with commit-window deal failures on one node; not causally isolated.

No panics on either node. Happy to share raw timestamps or run specific tests.

snadrus:
Fixes added: 
- iAmBored is important: run it even when in signalling mode
- avoid the "two flows to pollerTryAllWork" condition that results in task-list rug-pulls
- when we take a task, tell the signal mechanism that it's taken (not sure how I missed this)

TippyFlitsUK:
**Test report: integ/task @ `f7ebf14f` on calibnet**

Retested on calib2.ezpdpz.net with no local patch, so `pollRarely` is at the branch default of 30s. Our 23 July report was run on `5ef34bb5` with a local `pollRarely=3s` patch; this run tests the branch as-is, including `0d52ee7e`.

**Setup.** Node restarted 2026-07-28 13:16Z onto `f7ebf14f`. Previous build was `5ef34bb5` plus the local 3s patch. Same host, same Lotus (`1.36.0-rc1+calibnet+git.f0f80cadf`), same config.

**Failure rate, matched 20.5 hour windows:**

| branch | tasks | failed | rate |
|---|---|---|---|
| `5ef34bb5` + local pollRarely=3s (27-28 Jul) | 40,743 | 52 | 0.13% |
| `f7ebf14f` unpatched (28-29 Jul) | 47,429 | 139 | 0.29% |

**Error composition on `f7ebf14f`:** 133 `chain: i/o timeout`, 5 proof-submission failures, 1 pre-existing YugabyteDB read-restart. No panics, no SIGSEGV.

**IPNI.** 131 of 132 `PDPv0_IPNI` tasks succeeded. Dealbot IPNI verification checks against this node passed consistently, typically 8 to 12 per hour with zero timeouts. On the unpatched 23 July build those checks were timing out and failing deals when announce latency exceeded the verification window.

Happy to run additional queries or provide raw counts.

PR #1245, opened 2026-05-22
Refactor PDPv0 PullPiece scheduling, parking, and backpressure

### LINKED ISSUE #1241: pdpv0: PullPiece backlog grows unboundedly under sustained pull load; pieces lost silently and on restart — node paralysed

LexLuthr:
## Summary
This refactors PDPv0 pull handling so /pdp/piece/pull only records accepted pull work, while PDPPullPieceTask owns completion, failure, ref attachment, and download execution.

### Changes:
• Adds explicit pull item state with complete, created_at, parked_piece_ref, and pull_parked_piece_id.
• Stores client_address from FWSS payer extraction for per-client pull backpressure.
• Changes pull item identity to (fetch_id, piece_cid, source_url) so one pull can supply multiple URLs for the same piece.
• Groups PullPiece work by (piece_cid, piece_raw_size) and tries all source URLs for that piece.
• Parks pulled data directly into long-term piece storage when PullPiece creates the parked row.
• Reuses existing long-term parked rows when present instead of downloading again.
• Creates one parked_piece_ref / pdp_piecerefs path per pull item, even when the data came from another URL.
• Adds stale pull expiry using a 30-minute item budget and a 10-minute per-download attempt timeout.
• Cleans only PullPiece-owned refs, and flips skip=false when non-pull refs remain on an incomplete parked row.
• Adds handler backpressure with global/per-client pending-piece limits and 429 Retry-After.
• Fixes concurrent idempotent pull creation with ON CONFLICT DO NOTHING.

### Tests:
• Handler backpressure stress test with concurrent clients and random completion.
• Pull validation for same piece with multiple source URLs.
• PullPiece task tests for already-parked completion, pull-owned cleanup, and stale-expiry handoff to normal park processing.

### DIFF
--- a/harmony/harmonydb/sql/20260521-pdp-v0-pull-refactor.sql
+++ b/harmony/harmonydb/sql/20260521-pdp-v0-pull-refactor.sql
@@ -8,10 +8,22 @@ ALTER TABLE pdp_piece_pull_items
     ADD COLUMN IF NOT EXISTS created_at TIMESTAMPTZ NOT NULL DEFAULT NOW();
 
 ALTER TABLE pdp_piece_pull_items
+    ADD COLUMN IF NOT EXISTS attempt_count INTEGER NOT NULL DEFAULT 0;
+
+ALTER TABLE pdp_piece_pull_items
+    ADD COLUMN IF NOT EXISTS next_attempt_at TIMESTAMPTZ NOT NULL DEFAULT NOW();
+
+ALTER TABLE pdp_piece_pull_items
     ADD COLUMN IF NOT EXISTS parked_piece_ref BIGINT REFERENCES parked_piece_refs(ref_id) ON DELETE SET NULL;
 
 ALTER TABLE pdp_piece_pull_items
     ADD COLUMN IF NOT EXISTS pull_parked_piece_id BIGINT REFERENCES parked_pieces(id) ON DELETE SET NULL;
+
+COMMENT ON COLUMN pdp_piece_pull_items.parked_piece_ref IS
+    'Per-pull-item parked_piece_refs row; preserves the item source URL and is promoted into pdp_piecerefs on success.';
+
+COMMENT ON COLUMN pdp_piece_pull_items.pull_parked_piece_id IS
+    'Set only when PullPiece created the parked_pieces row; ownership marker for retry/expiry cleanup, distinct from parked_piece_ref.piece_id for refs attached to rows owned by other flows.';
 
 ALTER TABLE pdp_piece_pull_items
     DROP CONSTRAINT IF EXISTS pdp_piece_pull_items_pkey;
--- a/harmony/harmonydb/sql/20260521-pdp-v0-pull-refactor.sql
+++ b/harmony/harmonydb/sql/20260521-pdp-v0-pull-refactor.sql
@@ -1,9 +1,3 @@
--- PDPv0 pull pipeline refactor.
---
--- Pull items are now scheduled by unique piece key and written directly to
--- long-term piece storage. Pull item terminal state is explicit so status no
--- longer has to be inferred from StorePiece/parked_pieces task state.
-
 ALTER TABLE pdp_piece_pulls
     ADD COLUMN IF NOT EXISTS client_address TEXT NOT NULL DEFAULT '';
 
@@ -31,6 +25,14 @@ ALTER TABLE pdp_piece_pull_items
 ALTER TABLE pdp_piece_pull_items
     ADD CONSTRAINT pdp_piece_pull_items_terminal_state_check
     CHECK (NOT (complete AND failed));
+
+-- Support ON DELETE SET NULL from parked_piece_refs and parked_pieces without
+-- scanning all pull items for each deleted parent row.
+CREATE INDEX IF NOT EXISTS idx_pdp_piece_pull_items_parked_piece_ref
+    ON pdp_piece_pull_items (parked_piece_ref);
+
+CREATE INDEX IF NOT EXISTS idx_pdp_piece_pull_items_pull_parked_piece_id
+    ON pdp_piece_pull_items (pull_parked_piece_id);
 
 UPDATE pdp_piece_pull_items fi
 SET created_at = COALESCE(pp.created_at, fi.created_at)
--- a/harmony/harmonydb/sql/20260521-pdp-v0-pull-refactor.sql
+++ b/harmony/harmonydb/sql/20260521-pdp-v0-pull-refactor.sql
@@ -1,0 +1,54 @@
+-- PDPv0 pull pipeline refactor.
+--
+-- Pull items are now scheduled by unique piece key and written directly to
+-- long-term piece storage. Pull item terminal state is explicit so status no
+-- longer has to be inferred from StorePiece/parked_pieces task state.
+
+ALTER TABLE pdp_piece_pulls
+    ADD COLUMN IF NOT EXISTS client_address TEXT NOT NULL DEFAULT '';
+
+ALTER TABLE pdp_piece_pull_items
+    ADD COLUMN IF NOT EXISTS complete BOOLEAN NOT NULL DEFAULT FALSE;
+
+ALTER TABLE pdp_piece_pull_items
+    ADD COLUMN IF NOT EXISTS created_at TIMESTAMPTZ;
+
+UPDATE pdp_piece_pull_items fi
+SET created_at = COALESCE(pp.created_at, NOW())
+FROM pdp_piece_pulls pp
+WHERE pp.id = fi.fetch_id
+    AND fi.created_at IS NULL;
+
+ALTER TABLE pdp_piece_pull_items
+    ALTER COLUMN created_at SET DEFAULT NOW();
+
+ALTER TABLE pdp_piece_pull_items
+    ALTER COLUMN created_at SET NOT NULL;
+
+ALTER TABLE pdp_piece_pull_items
+    ADD COLUMN IF NOT EXISTS parked_piece_ref BIGINT REFERENCES parked_piece_refs(ref_id) ON DELETE SET NULL;
+
+ALTER TABLE pdp_piece_pull_items
+    ADD COLUMN IF NOT EXISTS pull_parked_piece_id BIGINT REFERENCES parked_pieces(id) ON DELETE SET NULL;
+
+ALTER TABLE pdp_piece_pull_items
+    DROP CONSTRAINT IF EXISTS pdp_piece_pull_items_pkey;
+
+ALTER TABLE pdp_piece_pull_items
+    ADD CONSTRAINT pdp_piece_pull_items_pkey PRIMARY KEY (fetch_id, piece_cid, source_url);
+
+UPDATE pdp_piece_pull_items fi
+SET task_id = NULL
+WHERE fi.task_id IS NOT NULL
+    AND NOT EXISTS (
+        SELECT 1
+        FROM harmony_task ht
+        WHERE ht.id = fi.task_id
+    );
+
+ALTER TABLE pdp_piece_pull_items
+    DROP CONSTRAINT IF EXISTS pdp_piece_pull_items_terminal_state_check;
+
+ALTER TABLE pdp_piece_pull_items
+    ADD CONSTRAINT pdp_piece_pull_items_terminal_state_check
+    CHECK (NOT (complete AND failed));
--- a/harmony/harmonydb/sql/20260521-pdp-v0-pull-refactor.sql
+++ b/harmony/harmonydb/sql/20260521-pdp-v0-pull-refactor.sql
@@ -13,11 +13,6 @@ ALTER TABLE pdp_piece_pull_items
 ALTER TABLE pdp_piece_pull_items
     ADD COLUMN IF NOT EXISTS created_at TIMESTAMPTZ NOT NULL DEFAULT NOW();
 
-UPDATE pdp_piece_pull_items fi
-SET created_at = COALESCE(pp.created_at, fi.created_at)
-FROM pdp_piece_pulls pp
-WHERE pp.id = fi.fetch_id;
-
 ALTER TABLE pdp_piece_pull_items
     ADD COLUMN IF NOT EXISTS parked_piece_ref BIGINT REFERENCES parked_piece_refs(ref_id) ON DELETE SET NULL;
 
@@ -30,18 +25,23 @@ ALTER TABLE pdp_piece_pull_items
 ALTER TABLE pdp_piece_pull_items
     ADD CONSTRAINT pdp_piece_pull_items_pkey PRIMARY KEY (fetch_id, piece_cid, source_url);
 
-UPDATE pdp_piece_pull_items fi
-SET task_id = NULL
-WHERE fi.task_id IS NOT NULL
-    AND NOT EXISTS (
-        SELECT 1
-        FROM harmony_task ht
-        WHERE ht.id = fi.task_id
-    );
-
 ALTER TABLE pdp_piece_pull_items
     DROP CONSTRAINT IF EXISTS pdp_piece_pull_items_terminal_state_check;
 
 ALTER TABLE pdp_piece_pull_items
     ADD CONSTRAINT pdp_piece_pull_items_terminal_state_check
     CHECK (NOT (complete AND failed));
+
+UPDATE pdp_piece_pull_items fi
+SET created_at = COALESCE(pp.created_at, fi.created_at)
+    FROM pdp_piece_pulls pp
+WHERE pp.id = fi.fetch_id;
+
+UPDATE pdp_piece_pull_items fi
+SET task_id = NULL
+WHERE fi.task_id IS NOT NULL
+  AND NOT EXISTS (
+    SELECT 1
+    FROM harmony_task ht
+    WHERE ht.id = fi.task_id
+);
--- a/harmony/harmonydb/sql/20260521-pdp-v0-pull-refactor.sql
+++ b/harmony/harmonydb/sql/20260521-pdp-v0-pull-refactor.sql
@@ -11,23 +11,12 @@ ALTER TABLE pdp_piece_pull_items
     ADD COLUMN IF NOT EXISTS complete BOOLEAN NOT NULL DEFAULT FALSE;
 
 ALTER TABLE pdp_piece_pull_items
-    ADD COLUMN IF NOT EXISTS created_at TIMESTAMPTZ;
+    ADD COLUMN IF NOT EXISTS created_at TIMESTAMPTZ NOT NULL DEFAULT NOW();
 
 UPDATE pdp_piece_pull_items fi
-SET created_at = COALESCE(pp.created_at, NOW())
+SET created_at = COALESCE(pp.created_at, fi.created_at)
 FROM pdp_piece_pulls pp
-WHERE pp.id = fi.fetch_id
-    AND fi.created_at IS NULL;
-
-UPDATE pdp_piece_pull_items
-SET created_at = NOW()
-WHERE created_at IS NULL;
-
-ALTER TABLE pdp_piece_pull_items
-    ALTER COLUMN created_at SET DEFAULT NOW();
-
-ALTER TABLE pdp_piece_pull_items
-    ALTER COLUMN created_at SET NOT NULL;
+WHERE pp.id = fi.fetch_id;
 
 ALTER TABLE pdp_piece_pull_items
     ADD COLUMN IF NOT EXISTS parked_piece_ref BIGINT REFERENCES parked_piece_refs(ref_id) ON DELETE SET NULL;
--- a/harmony/harmonydb/sql/20260521-pdp-v0-pull-refactor.sql
+++ b/harmony/harmonydb/sql/20260521-pdp-v0-pull-refactor.sql
@@ -19,6 +19,10 @@ FROM pdp_piece_pulls pp
 WHERE pp.id = fi.fetch_id
     AND fi.created_at IS NULL;
 
+UPDATE pdp_piece_pull_items
+SET created_at = NOW()
+WHERE created_at IS NULL;
+
 ALTER TABLE pdp_piece_pull_items
     ALTER COLUMN created_at SET DEFAULT NOW();
 
--- a/pdp/handlers.go
+++ b/pdp/handlers.go
@@ -1256,9 +1256,44 @@ func (p *PDPService) cleanup(ctx context.Context) {
 			}
 		}
 
-		// Clean up old piece pull records (older than 5 days)
-		// CASCADE deletes pdp_piece_pull_items automatically
-		_, err = db.Exec(ctx, `DELETE FROM pdp_piece_pulls WHERE created_at < NOW() - INTERVAL '5 days'`)
+		// Clean up old piece pull records (older than 5 days). Pull items only
+		// store refs created by PullPiece, so delete unused refs first;
+		// otherwise the CASCADE from pdp_piece_pulls removes the pointer to
+		// those refs.
+		_, err = db.BeginTransaction(ctx, func(tx *harmonydb.Tx) (bool, error) {
+			_, err = tx.Exec(`
+				WITH old_pull_refs AS (
+					SELECT DISTINCT fi.parked_piece_ref AS ref_id
+					FROM pdp_piece_pull_items fi
+					JOIN pdp_piece_pulls pp ON pp.id = fi.fetch_id
+					WHERE pp.created_at < NOW() - INTERVAL '5 days'
+						AND fi.parked_piece_ref IS NOT NULL
+				)
+				DELETE FROM parked_piece_refs pr
+				USING old_pull_refs old
+				WHERE pr.ref_id = old.ref_id
+					AND NOT EXISTS (
+						SELECT 1 FROM pdp_piecerefs ppr WHERE ppr.piece_ref = pr.ref_id
+					)
+					AND NOT EXISTS (
+						SELECT 1
+						FROM pdp_piece_pull_items live_fi
+						JOIN pdp_piece_pulls live_pp ON live_pp.id = live_fi.fetch_id
+						WHERE live_fi.parked_piece_ref = pr.ref_id
+							AND live_pp.created_at >= NOW() - INTERVAL '5 days'
+					)
+				`)
+			if err != nil {
+				return false, fmt.Errorf("delete old pull refs: %w", err)
+			}
+
+			_, err = tx.Exec(`DELETE FROM pdp_piece_pulls WHERE created_at < NOW() - INTERVAL '5 days'`)
+			if err != nil {
+				return false, fmt.Errorf("delete old pull records: %w", err)
+			}
+
+			return true, nil
+		}, harmonydb.OptionRetry())
 		if err != nil {
 			log.Errorw("failed to delete old piece pull records", "error", err)
 		}
--- a/pdp/handlers_pull.go
+++ b/pdp/handlers_pull.go
@@ -177,22 +177,19 @@ func NewPullHandler(auth Auth, store PullStore, validator AddPiecesValidator) *P
 //  1. Client calls POST /pdp/piece/pull with piece CIDs and source URLs
 //  2. Server validates extraData via eth_call (ensures authorization)
 //  3. Server creates pull tracking records; duplicate pieces with different
-//     source URLs are kept so PullPiece can try each supplied source.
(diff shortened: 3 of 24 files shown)


harmony/harmonydb/sql/20260521-pdp-v0-pull-refactor.sql
```diff
@@ -19,6 +19,10 @@ FROM pdp_piece_pulls pp
 WHERE pp.id = fi.fetch_id
     AND fi.created_at IS NULL;
 
+UPDATE pdp_piece_pull_items
```
beck-8: The problem is still there
because ALTER COLUMN SET DEFAULT NOW() (a DDL statement) triggers an implicit transaction restart, rolling back the preceding UPDATE
  statements that fill in the NULL values. When SET NOT NULL runs afterward, the NULLs are back and the constraint check fails.

      24  WHERE created_at IS NULL;
      25
      26  ALTER TABLE pdp_piece_pull_items
      27 -    ALTER COLUMN created_at SET DEFAULT NOW();
      27 +    ALTER COLUMN created_at SET NOT NULL;
      28
      29  ALTER TABLE pdp_piece_pull_items
      30 -    ALTER COLUMN created_at SET NOT NULL;
      30 +    ALTER COLUMN created_at SET DEFAULT NOW();
      31
      32  ALTER TABLE pdp_piece_pull_items
      33      ADD COLUMN IF NOT EXISTS parked_piece_ref BIGINT REFERENCES parked_piece_refs(ref_id) ON DELETE SET NULL;

LexLuthr: Reversed the order of things. It should work now.

ZenGround0:
Some very high level feedback is that this would have been a good place to create a sequence of PRs based off of Rod's breakdown of the issue into separate jobs.  Avoiding > 1000 line PRs when possible really helps keep things manageable for review and partial progress.

A good compromise is to at least keep individual commits focused on conceptually separate pieces.  Each 1-2 of your bullet points above could have been a separate commit for example.

ZenGround0:
Since this is superceding #1242 which was tested to resolve some aspects of the problem I'd expect it to include all the remediations there.  I understand that you are using the backpressure constants instead of configuring `PDPPullPieceMaxTasks` which is fine. But why not include @rvagg's `StorePieceMaxTasks` limit? Also if I'm reading you changed PullPiece `Cost.Cpu = 0` as in #1242 but not StorePiece `Cost.Cpu = 0`.  Was this intentional?

pdp/handlers.go
ZenGround0, line 1240: This is not your problem at all or relevant to the scope of this bug but this thing should probably be a Harmony task triggered from an infrequent IAMBored trigger right?

LexLuthr:
> Since this is superceding #1242 which was tested to resolve some aspects of the problem I'd expect it to include all the remediations there. I understand that you are using the backpressure constants instead of configuring `PDPPullPieceMaxTasks` which is fine. But why not include @rvagg's `StorePieceMaxTasks` limit? Also if I'm reading you changed PullPiece `Cost.Cpu = 0` as in #1242 but not StorePiece `Cost.Cpu = 0`. Was this intentional?

That task requires CPU. We are forgetting the CPU cost of doing commP there. Curio's principle has been good default instead and least config knobs. I don't think we have exhausted that yet. If there is no other recourse, a config item would make sense.

LexLuthr:
> Some very high level feedback is that this would have been a good place to create a sequence of PRs based off of Rod's breakdown of the issue into separate jobs. Avoiding > 1000 line PRs when possible really helps keep things manageable for review and partial progress.
> 
> A good compromise is to at least keep individual commits focused on conceptually separate pieces. Each 1-2 of your bullet points above could have been a separate commit for example.

There are no separate pieces in this PR. It basically rewrites the whole pull_piece subsystem. It stabilized the [problem](https://github.com/filecoin-project/curio/issues/1248#issuecomment-4526303137). If I change ingress even tiny bit, I have to stabilize the task for it. This in turn requires me to check if it break anything in the pipelines which it did.
My suggestion would be not treat it as a improvement but a completely rewrite.

After stabilization, I added more commits to eliminate the issues due to missing DB indexes after some query plan testing.

Some indexing commits are in deletion PR series as they were more relevant there. But, they also have an impact on this overall.

pdp/piece_cid.go
```diff
@@ -126,12 +126,5 @@ func PadPieceSize(rawSize int64) int64 {
...
 		return 0
 	}
-	// Apply FR32 ratio: ceil(rawSize * 128 / 127)
-	fr32Padded := (uint64(rawSize)*128 + 126) / 127
-	// Round up to next power of 2
-	if fr32Padded == 0 {
-		return 1
-	}
-	fr32Padded--
-	return int64(1 << bits.Len64(fr32Padded))
```
ZenGround0, line 136: Nice!  This is exactly the sort of cleanup that would benefit from being applied in a later follow on PR.  It is a valuable contribution but takes time to review and would be better to sequence after a more critical fix.

pdp/handlers.go
LexLuthr: Yes. A follow up PR. Along with handler changes. I have opened a few issues today to document all these things that we move to a different trigger/architecture.

harmony/harmonydb/sql/20260521-pdp-v0-pull-refactor.sql
```diff
@@ -0,0 +1,49 @@
...
+
+ALTER TABLE pdp_piece_pull_items
+    ADD COLUMN IF NOT EXISTS created_at TIMESTAMPTZ NOT NULL DEFAULT NOW();
+
+ALTER TABLE pdp_piece_pull_items
+    ADD COLUMN IF NOT EXISTS parked_piece_ref BIGINT REFERENCES parked_piece_refs(ref_id) ON DELETE SET NULL;
+
+ALTER TABLE pdp_piece_pull_items
+    ADD COLUMN IF NOT EXISTS pull_parked_piece_id BIGINT REFERENCES parked_pieces(id) ON DELETE SET NULL;
+
```
ZenGround0, line 21: It seems a little off that we reference both a parked_piece_ref and a parked_pieces id.  Aren't parked piece refs thin wrappers over parked_pieces ids? parked_piece_ref.piece_id should be the same thing right?

pdp/handlers_pull.go
```diff
@@ -250,20 +176,23 @@ func NewPullHandler(auth Auth, store PullStore, validator AddPiecesValidator) *P
...
-//   - pending: Piece is queued but download hasn't started
-//   - inProgress: Download task is actively running (first attempt)
-//   - retrying: Download task is running after one or more failures
+//   - pending: Piece is queued or waiting on an existing parked_pieces row
+//   - inProgress: PullPiece task is actively running
+//   - retrying: The current task row has a retry count greater than zero
 //   - complete: Piece successfully downloaded and verified
-//   - failed: Piece permanently failed after exhausting retries (currently 5 attempts)
+//   - failed: Piece permanently failed; the task enforces this through explicit
+//     item state rather than inferring it from parked_pieces.
```
rvagg: I'm not sure this comment is all that helpful here (the failed one in particular) - think of someone implementing a client for this, how are they to understand what the states mean? these comments are now rewritten to talk about internal details a bit too much IMO

pdp/indexing.go
```diff
@@ -58,66 +60,114 @@ func CheckIfIndexingNeededFromExtraData(extraData []byte) (bool, error) {
...
+
+	return payload.Payer, nil
+}
+
+type fwssCreatePayload struct {
+	Payer        common.Address
+	MetadataKeys []string
+}
+
+func decodeFWSSCreatePayload(extraData []byte) (*fwssCreatePayload, error) {
```
rvagg: we should pull this out into a utils; not critical here but now we have multiple uses for it we may as well put it somewhere separate to "indexing"

pdp/pull_types.go
```diff
@@ -228,3 +299,295 @@ func (r *PullResponse) ComputeOverallStatus() {
...
+	db *harmonydb.DB
+	mu sync.Mutex
+}
+
+// NewDBPullStore creates a PullStore backed by harmonydb
+func NewDBPullStore(db *harmonydb.DB) PullStore {
+	return &dbPullStore{db: db}
+}
+
+func (s *dbPullStore) GetPullByKey(ctx context.Context, service string, hash []byte, dataSetId uint64, recordKeeper string) (*PullRecord, error) {
```
rvagg, line 325: these are more than "types" now, not critical but IMO this file and its matching test could be renamed, or put  strict types in one file and then these functions that are dependent on the outside world in another

pdp/pull_types.go
```diff
@@ -228,3 +299,295 @@ func (r *PullResponse) ComputeOverallStatus() {
...
+	}
+
+	return nil, nil
+}
+
+func pullClientBorrowLimit() int {
+	return pullGlobalPendingLimit * pullSoloClientPendingPercentage / 100
+}
+
+func pullEffectiveClientPendingLimit(otherPending int) int {
```
rvagg, line 596: Oh, this is clever .. so you get to borrow up to 90% of available total capacity if nobody else is doing anything.
This does mean we need to care about performance of ~108 parallell pulls if a client can spawn that many, if an SP doesn't do 108 in a ~reasonable amount of time then it's going to be more painful on the receiving end than getting a 429. I think my numbers using Tippy's SP suggested this might push into risky territory with medium->large pieces.
@hugomrdias' feedback on whether they're getting timeouts vs 429's would be valuable calibaration on this. My hunch is that 90% may be too much, or alternatively the total pending is too much (if the server's really busy with lots of clients then it's the ~same effect).

pdp/pull_types.go
```diff
@@ -228,3 +299,295 @@ func (r *PullResponse) ComputeOverallStatus() {
...
+	limit := pullClientBorrowLimit() - otherPending
+	if limit < pullPerClientPendingLimit {
+		return pullPerClientPendingLimit
+	}
+	return limit
+}
+
+func pullRetryAfter(pending, limit int) time.Duration {
```
rvagg, line 606: docs for these functions with convoluted logic would be nice, e.g.:


```suggestion
// pullRetryAfter suggests how long a rejected pull should wait: 1 minute per 10
// pieces over the limit, clamped to [1, 5] minutes.
func pullRetryAfter(pending, limit int) time.Duration {
```

pdp/pull_types.go
```diff
@@ -228,3 +299,295 @@ func (r *PullResponse) ComputeOverallStatus() {
...
+	if taskExists {
+		if retries > 0 {
+			return PullStatusRetrying
+		}
+		return PullStatusInProgress
+	}
+	return PullStatusPending
+}
+
+func (s *dbPullStore) enforceBackpressure(tx *harmonydb.Tx, pull *PullRecord, pieces []PullPiece) (*PullBackpressure, error) {
```
rvagg, line 498: a terse high-level doc for this would be nice too, how does it do this? when someone's diving in here what are they looking for to make sense of?

tasks/pdpv0/task_pull_piece.go
```diff
@@ -2,417 +2,1031 @@ package pdpv0
...
+// Do first loads every active pull item for the assigned piece group. That includes
+// late arrivals for the same piece, not just rows that originally held this
+// task_id.
+//
+// It then claims the active long-term parked_pieces row for that piece key:
+//
+//   - If an ordinary row already exists, PullPiece attaches one parked_piece_ref
+//     per pull item and exits. ParkPiece/StorePiece continues normal processing,
+//     or a later scheduler pass marks the pull items complete if the row is
+//     already complete.
```
rvagg:
```suggestion
//   - If an ordinary row from another flow (push/upload, market) already exists,
//     PullPiece attaches one parked_piece_ref per pull item and exits.
//     ParkPiece/StorePiece continues normal processing, or a later scheduler
//     pass marks the pull items complete if the row is already complete.
```

nitpick but I read this and was confused why StorePiece was still implicated here, I get it now but a hint to other readers would be nice

tasks/pdpv0/task_pull_piece.go
```diff
@@ -2,417 +2,1031 @@ package pdpv0
...
+			WHERE complete = FALSE
+				AND failed = FALSE
+				AND task_id IS NOT NULL
+		) active
+	`).Scan(&activeGroups)
 	if err != nil {
-		return false, xerrors.Errorf("fetching task data: %w", err)
+		return xerrors.Errorf("count active pull item groups: %w", err)
+	}
+	if activeGroups >= t.max {
```
rvagg, line 521: comments for blocks like this would be good, something like "Check whether more than max are currently assigned" .. it took me a bit to figure find the `AND task_id IS NOT NULL` to grok this block

tasks/pdpv0/task_pull_piece.go
```diff
@@ -424,55 +1038,33 @@ func (t *PDPPullPieceTask) TypeDetails() harmonytask.TaskTypeDetails {
...
-		MaxFailures: 5,
-		RetryWait: func(retries int) time.Duration {
-			const baseWait, maxWait, factor = 10 * time.Second, 5 * time.Minute, 2.0
-			return min(time.Duration(float64(baseWait)*math.Pow(factor, float64(retries))), maxWait)
+			Cpu:     0,
+			Gpu:     0,
+			Ram:     128 << 20,
+			Storage: nil,
 		},
+		MaxFailures: 1,
```
rvagg, line 1356: I don't think this is going to have the effect you want. A failure will make Harmony `DELETE FROM harmony_task WHERE id=$1`, but in harmony/harmonydb/sql/20260109-pdp-v0-pull.sql we have a FK making us clear `pdp_piece_pull_items.task_id` when `harmony_task` is removed. This then would put it back into the pool of eligible pulls to be picked up again, so 5s later we're on again, and we only give up on it after 30 minutes. So if our URL doesn't work, a 404 perhaps, then we keep on hitting it every 5 seconds anyway.

Was the intent to fail fast on permanently-bad pieces? If so we'd need a permanent-vs-transient classification (404/410 fail the item now) or an attempt counter, since right now nothing bounds it below the budget. I'd prefer this, ideally with a retry counter, but I think in Slack you were suggesting any kind of failure should quickly knock it out.

I think what you probably want is another `failed=TRUE`, the only one you have now is the `fail_reason = 'pull budget exceeded'` one.

rvagg:
> That task requires CPU. We are forgetting the CPU cost of doing commP there

it's true, but commp CPU is very simple _and_ it's bounded by the speed of the download since it's done in a multi reader, so it's a trickle of CPU usage

rvagg:
This lgtm other than the concern I raised inline about the retries, I think we're only bounded on failure by the total 30 minute timeout, bad download URLs get retried every 5 seconds from what I can tell. Also I think more inline comments would be good, just terse notes where complex stuff happens or heavy functions have multiple blocks of separate work, just to help people trying to understand the code (it does help AI too).

pdp/indexing.go
LexLuthr: Done

pdp/pull_types.go
LexLuthr: @hugomrdias is testing this against deployed SP. He will let me know if he wants to tweak anything. We already discussed this design and he is good with it.

LexLuthr: Taking 108 does not mean I will be doing 108 pieces (that is governed by task not handler). Most of them will expire before we retry if they are unique 108 pieces. Unless these are tiny 108 piece. This design allows us to saturate system when we have small pieces and also be a little more opinionated when pieces are large.

tasks/pdpv0/task_pull_piece.go
LexLuthr: Hmm! I have been thinking about this with bundle of URLs and not one. Then abstracting away one as a special case. Your concern is valid here, current design would make bad URL enjoy more retries can good ones.

I think we can try to combine both:

1. We will keep PullPiece’s current breadth-first task model, but add pull-item retry state so bad sources do not churn every poll cycle.
2. Each pull item gets attempt_count and next_attempt_at. Transient failures increment attempts and retry with backoff, up to 3 attempts or the existing 30-minute budget. Permanent failures fail the whole piece group immediately. If any URL succeeds, the piece group completes as today.
3. Permanent failures start with clear bad-source cases: 404, 410, invalid URL, CommP mismatch, size mismatch, and oversized response. Network errors, timeouts, 429, 5xx, and short reads remain transient.
4. We will not reset attempts based on queue emptiness. Queue pressure can affect retry delay later if needed, but not the attempt counter.

pdp/pull_types.go
LexLuthr: Added comments

harmony/harmonydb/sql/20260521-pdp-v0-pull-refactor.sql
LexLuthr: • parked_piece_ref: the per-pull-item ref we create for that request/source URL. This is what can later become the pdp_piecerefs entry.
• pull_parked_piece_id: ownership marker. It is set only when PullPiece created the parked_pieces row. It tells cleanup/retry logic, “this parked row is ours to clean up or hand back.” 

Without pull_parked_piece_id, cleanup would have to infer ownership through parked_piece_ref -> parked_piece_refs.piece_id, which is both noisier and semantically wrong for existing parked rows: PullPiece may attach refs to a parked row owned by another flow, but it must not treat that row as pull-owned.
Also, if parked_piece_ref is cleared by ON DELETE SET NULL, we would lose the direct pointer needed to clean up a pull-created parked row.

I have added comments to those in newer commit.

tasks/pdpv0/task_pull_piece.go
```diff
@@ -722,76 +861,11 @@ func (t *PDPPullPieceTask) Do(taskID harmonytask.TaskID, stillOwned func() bool)
...
-			if parsedURL.Scheme != "https" && (!pdp.PullAllowInsecure() || parsedURL.Scheme != "http") {
-				sourceErr = xerrors.Errorf("source URL must use HTTPS scheme, got: %s", parsedURL.Scheme)
-			}
-		}
-		downloadPolicy := ssrfPolicy
-		downloadPolicy.ResponseHeaderTimeout = 2 * time.Minute
-		if sourceErr == nil {
-			_, sourceErr = robusthttp.ValidateClientFetchURL(source.SourceURL, nil, &downloadPolicy)
-		}
+		sourceErr := t.tryPullSource(ctx, source.SourceURL, group, expectedCid, parkedPieceID, ssrfPolicy)
```
rvagg, line 864: 👌 nice to see this pulled out

tasks/pdpv0/task_pull_piece_test.go
```diff
@@ -243,6 +251,292 @@ func TestExpireStalePullItemsEnablesParkTaskWhenOtherRefsRemain(t *testing.T) {
...
+}
+
+func TestPullSourceErrorClassification(t *testing.T) {
+	tests := []struct {
+		name      string
+		failure   pullSourceError
+		transient bool
+	}{
+		{
+			name:      "404 is transient",
```
rvagg: why is 404 transient? we could permafail that one couldn't we?

rvagg (approved):
lgtm sans the question about 404 - if it's not there then it's probably not going to be there, although there is another note "Curio-side 404's", I'm not sure what that is?

tasks/pdpv0/task_pull_piece_test.go
LexLuthr: You are right. I was thinking about an edge case in PDPv1 but v0 does not have it. Once piece is available we allow it. I will remove 404 as well.

beck-8:
```
{"level":"info","ts":"2026-05-26T21:18:31.774+0800","logger":"metadata","caller":"metadata/task_sector_expirations.go:513","msg":"starting verifreg claim crawl"}
{"level":"info","ts":"2026-05-26T21:18:31.782+0800","logger":"metadata","caller":"metadata/task_sector_expirations.go:558","msg":"crawling verifreg claims","sp_count":0}
{"level":"info","ts":"2026-05-26T21:18:31.782+0800","logger":"metadata","caller":"metadata/task_sector_expirations.go:754","msg":"completed verifreg claim crawl","total_updates":0}
{"level":"info","ts":"2026-05-26T21:18:31.786+0800","logger":"metadata","caller":"metadata/task_sector_expirations.go:533","msg":"completed sector metadata refresh","epoch":"3749134","total_updates":0}
{"level":"info","ts":"2026-05-26T21:18:34.763+0800","logger":"harmonytask","caller":"harmonytask/task_type_handler.go:248","msg":"Beginning work on Task","id":30714759,"from":"bored","name":"AlertManager","sector":null}
{"level":"error","ts":"2026-05-26T21:18:34.883+0800","logger":"pdpv0","caller":"pdpv0/task_pull_piece.go:172","msg":"failed to schedule pull items: query pending pull item groups: ERROR: column fi.attempt_count does not exist (SQLSTATE 42703)"}
{"level":"warn","ts":"2026-05-26T21:18:34.894+0800","logger":"curio/alertmanager","caller":"alertmanager/task_alert.go:318","msg":"No alert plugins enabled, alerts recorded to DB but not sent externally"}
{"level":"error","ts":"2026-05-26T21:18:39.902+0800","logger":"pdpv0","caller":"pdpv0/task_pull_piece.go:172","msg":"failed to schedule pull items: query pending pull item groups: ERROR: column fi.attempt_count does not exist (SQLSTATE 42703)"}
{"level":"error","ts":"2026-05-26T21:18:44.903+0800","logger":"pdpv0","caller":"pdpv0/task_pull_piece.go:172","msg":"failed to schedule pull items: query pending pull item groups: ERROR: column fi.attempt_count does not exist (SQLSTATE 42703)"}
{"level":"error","ts":"2026-05-26T21:18:49.883+0800","logger":"pdpv0","caller":"pdpv0/task_pull_piece.go:172","msg":"failed to schedule pull items: query pending pull item groups: ERROR: column fi.attempt_count does not exist (SQLSTATE 42703)"}
{"level":"error","ts":"2026-05-26T21:18:54.905+0800","logger":"pdpv0","caller":"pdpv0/task_pull_piece.go:172","msg":"failed to schedule pull items: query pending pull item groups: ERROR: column fi.attempt_count does not exist (SQLSTATE 42703)"}
```

LexLuthr:
> ```
> {"level":"info","ts":"2026-05-26T21:18:31.774+0800","logger":"metadata","caller":"metadata/task_sector_expirations.go:513","msg":"starting verifreg claim crawl"}
> {"level":"info","ts":"2026-05-26T21:18:31.782+0800","logger":"metadata","caller":"metadata/task_sector_expirations.go:558","msg":"crawling verifreg claims","sp_count":0}
> {"level":"info","ts":"2026-05-26T21:18:31.782+0800","logger":"metadata","caller":"metadata/task_sector_expirations.go:754","msg":"completed verifreg claim crawl","total_updates":0}
> {"level":"info","ts":"2026-05-26T21:18:31.786+0800","logger":"metadata","caller":"metadata/task_sector_expirations.go:533","msg":"completed sector metadata refresh","epoch":"3749134","total_updates":0}
> {"level":"info","ts":"2026-05-26T21:18:34.763+0800","logger":"harmonytask","caller":"harmonytask/task_type_handler.go:248","msg":"Beginning work on Task","id":30714759,"from":"bored","name":"AlertManager","sector":null}
> {"level":"error","ts":"2026-05-26T21:18:34.883+0800","logger":"pdpv0","caller":"pdpv0/task_pull_piece.go:172","msg":"failed to schedule pull items: query pending pull item groups: ERROR: column fi.attempt_count does not exist (SQLSTATE 42703)"}
> {"level":"warn","ts":"2026-05-26T21:18:34.894+0800","logger":"curio/alertmanager","caller":"alertmanager/task_alert.go:318","msg":"No alert plugins enabled, alerts recorded to DB but not sent externally"}
> {"level":"error","ts":"2026-05-26T21:18:39.902+0800","logger":"pdpv0","caller":"pdpv0/task_pull_piece.go:172","msg":"failed to schedule pull items: query pending pull item groups: ERROR: column fi.attempt_count does not exist (SQLSTATE 42703)"}
> {"level":"error","ts":"2026-05-26T21:18:44.903+0800","logger":"pdpv0","caller":"pdpv0/task_pull_piece.go:172","msg":"failed to schedule pull items: query pending pull item groups: ERROR: column fi.attempt_count does not exist (SQLSTATE 42703)"}
> {"level":"error","ts":"2026-05-26T21:18:49.883+0800","logger":"pdpv0","caller":"pdpv0/task_pull_piece.go:172","msg":"failed to schedule pull items: query pending pull item groups: ERROR: column fi.attempt_count does not exist (SQLSTATE 42703)"}
> {"level":"error","ts":"2026-05-26T21:18:54.905+0800","logger":"pdpv0","caller":"pdpv0/task_pull_piece.go:172","msg":"failed to schedule pull items: query pending pull item groups: ERROR: column fi.attempt_count does not exist (SQLSTATE 42703)"}
> ```

You need to run the extra SQL manually. Downside of running a WIP PR.

ZenGround0 (approved):
>     --- FAIL: TestPullSourceErrorClassification/5xx_is_transient (0.00s)

Failing test on Pull needs some investigation before merge.  Other than that I am happy moving forward with @rvagg 's approval here

snadrus (approved):

LexLuthr:
> > ```
> > --- FAIL: TestPullSourceErrorClassification/5xx_is_transient (0.00s)
> > ```
> 
> Failing test on Pull needs some investigation before merge. Other than that I am happy moving forward with @rvagg 's approval here

That was http code 5xx test. I flipped the code but forgot to flip the test

PR #1246, opened 2026-05-22
pdpv0: user initiated termination

### LINKED ISSUE #1235: PDP: Implement pdpv0 client termination endpoint

LexLuthr:
fixes https://github.com/filecoin-project/curio/issues/1235

## Summary
Adds pdpv0 client-requested data set termination flow.
• Adds POST /pdp/data-sets/{id}/terminate to queue client termination with required non-empty extraData.
• Adds GET /pdp/data-sets/{id}/terminate for client termination status, including tx hash/status/success, FWSS termination confirmation, and service termination epoch.
• Tracks client-requested termination state in pdp_delete_data_set.
• Makes terminate retry/failure handling origin-aware:
◦ client failed txs are deleted so clients can retry
◦ provider/internal failed txs are reset for internal retry
◦ failed txs that already terminated FWSS still advance the pipeline
• Adds chain-sync reconciliation for missing termination message-wait rows using FWSS state.
• Adds immediate settlement task for client-requested terminations and marks settlement transactions with a reason so pay watcher can distinguish them.
• Adds race documentation for client termination/immediate settlement interactions.

### Notes
~FWSS ABI changes are still pending. Current ABI TODOs remain in TerminateFWSSTask and should be resolved when terminateService(uint256,bytes) is available.~
ABI has been regenerated for FWSS based on https://github.com/FilOzone/filecoin-services/pull/485


tasks/pay/watcher.go
```diff
@@ -88,27 +89,34 @@ func processPendingTransactions(ctx context.Context, db *harmonydb.DB, ethClient
 	}
 
 	for _, mwe := range mwes {
-		// wait for message to land
-		if mwe.Status != "confirmed" {
-			delete(goodSettles, mwe.Hash)
+		settle, ok := goodSettles[mwe.Hash]
```
ZenGround0: Why are you adding this? how could mwe ever not be in goodSettles given how all goodSettles keys are in hashes and mwes is selected via signed_tx_hash = ANY(hashes)?

tasks/pay/watcher.go
```diff
@@ -88,27 +89,34 @@ func processPendingTransactions(ctx context.Context, db *harmonydb.DB, ethClient
...
 			delete(goodSettles, mwe.Hash)
 			log.Errorw("settlement transaction failed on chain", "txHash", mwe.Hash)
-			al.Raise(at, map[string]interface{}{
-				"txHash": mwe.Hash,
-				"error":  "settlement transaction failed on chain",
-			})
-			_, err = db.Exec(ctx, `DELETE FROM filecoin_payment_transactions WHERE tx_hash = $1`, mwe.Hash)
-			if err != nil {
-				return xerrors.Errorf("failed to delete failed settlement from DB: %w", err)
+			if settle.Reason != SettleReasonPDPv0ClientTermination {
```
ZenGround0: why only alert on periodic settlement?  I'd expect us to alert on client termination too

tasks/pdpv0/watch_fwss_terminate.go
```diff
@@ -129,43 +128,168 @@ func processSuccessfulTerminations(ctx context.Context, db *harmonydb.DB, ethCli
...
+func fwssServiceTerminationEpoch(ds FWSS.FilecoinWarmStorageServiceDataSetInfoView) *int64 {
+	if ds.PdpEndEpoch == nil || ds.PdpEndEpoch.Sign() == 0 {
+		return nil
+	}
+	epoch := ds.PdpEndEpoch.Int64()
+	return &epoch
+}
+
+func markServiceTerminationConfirmed(ctx context.Context, db *harmonydb.DB, detail pendingServiceTermination, epoch int64) error {
+	committed, err := db.BeginTransaction(ctx, func(tx *harmonydb.Tx) (bool, error) {
```
ZenGround0: For my understanding what's the benefit of wrapping the exec plus error handling and updated row counting into a transaction as opposed to just running a db.Exec and then doing error handling outside of it? I was assuming we'd want to limit db transaction function as much as possible and try to put error handling outside of it.  I guess the idea is that if we are messing up so badly that multiple rows are updated then we don't want to commit the change to the db?

tasks/pdpv0/task_chain_sync.go
```diff
@@ -218,13 +222,35 @@ func (t *TaskChainSync) syncMissingDataSetTerminationMessageWaits(ctx context.Co
 			continue
 		}
 
+		if detail.ClientRequested {
+			n, err := t.db.Exec(ctx, `
+				DELETE FROM pdp_delete_data_set
```
ZenGround0, line 227: I don't quite get why we delete the row here but retry in the other case.  If we've logged the client request in the DB then why not try again?

ZenGround0: Is this part of the broader policy you have in mind for task_chain_sync maybe?  That information flows only from chain to db not from db to chain?

ZenGround0: If so I can see the value in that.  But I'm guessing there are more cases where we want something symmetric that goes the other way.

pdp/handlers_terminate.go
```diff
@@ -0,0 +1,225 @@
...
+	if dataSetIDStr == "" {
+		http.Error(w, "Missing data set ID in URL", http.StatusBadRequest)
+		return
+	}
+
+	dataSetID, err := strconv.ParseUint(dataSetIDStr, 10, 64)
+	if err != nil {
+		http.Error(w, "Invalid data set ID format", http.StatusBadRequest)
+		return
+	}
```
ZenGround0, line 52: nit: not your bug but just a thought that this and a few other parsing gadgets show up everywhere and would be good candidates for library functions

ZenGround0: (Similar story for extra data parsing below)

tasks/pay/immediate_settle_task.go
```diff
@@ -0,0 +1,437 @@
...
+	"github.com/filecoin-project/curio/lib/ethchain"
+	"github.com/filecoin-project/curio/lib/filecoinpayment"
+	"github.com/filecoin-project/curio/lib/passcall"
+	"github.com/filecoin-project/curio/pdp/contract"
+	"github.com/filecoin-project/curio/pdp/contract/FWSS"
+	"github.com/filecoin-project/curio/tasks/message"
+	"github.com/filecoin-project/curio/tasks/tasknames"
+)
+
+/*
```
ZenGround0: What is the strongest case you can make for immediate settlement on client termination ?  It is all well designed but I can't justify including it when the other alternatives are so good.  

"Periodic pay settlement remains the correctness path." tells me all I need to know.  This is just a speedup and I don't think the speedup from ~daily settlement frequency is at all worth the complexity of a new task and a new field in the `filecoin_payment_transactions` table

ZenGround0 (approved):
Thanks this is a really nice feature cleaning up as you go along.  The only thing I block on is including the immediate settle task which I view as well crafted but wholly unnecessary.  If you want to argue about keeping that please do but hold off on merge.  Everything is uncontroversial or not important enough to block

tasks/pdpv0/watch_fwss_terminate.go
LexLuthr: Atomic DB update. Using Tx for everything is bad but non atomic state changes are worse. So, whenever we are updating the pipelines, we should be atomic with a Tx.

tasks/pay/watcher.go
LexLuthr: Alerts are for things which require user attention. If immediate settlement fails, we let normal periodic take over it. If it fails there as well, we alert the user that they might be loosing money.

tasks/pay/immediate_settle_task.go
LexLuthr: This comes from the request https://github.com/filecoin-project/curio/issues/1236. It just made more logical sense to include this part in termination changes. Keep the work area separate for a cleaner PR. Otherwise, https://github.com/filecoin-project/curio/pull/1253 would end up going over all this code again

pdp/handlers_terminate.go
LexLuthr: Oh! It is already moved into common utils in https://github.com/filecoin-project/curio/pull/1245. I can do better clean up once all the current open PRs are merged and we have a holistic view.

tasks/pay/watcher.go
LexLuthr: If we add it to filecoin_payment_transactions but not message watcher then we get this state. Some of the inserts were not DB atomic before I started on this branch. So, I am ensuring we are always guarding against those. We are doing ANY() so it will give us all matches but not exactly all matches.

tasks/pay/immediate_settle_task.go
ZenGround0: I see. When making decisions to merge requirements from different issues you need to at least flag this in the PR description.  Otherwise reviewers are going to be lost.

ZenGround0: I guess a way to frame it that has clicked for me in the past is that PRs are just as much about communicating a change as adding code to the system

LexLuthr: Yes. My bad. I should have called out that I am bringing in some things from other issue. I will improve on this.

tasks/pdpv0/task_chain_sync.go
LexLuthr: Failed immediate-settle transactions must not reset immediate-settle fields for retry. Otherwise the immediate path can requeue forever. This way if user send us back request or some other systemic issue then we try and exist. user can request termination again later with correct params.

tasks/pdpv0/watch_fwss_terminate.go
ZenGround0: But that's my question if there is only a single sql query running as in here then db.Exec is also an atomic state change.  What do we gain by using Tx to wrap a single exec?

tasks/pay/watcher.go
ZenGround0: thanks makes sense

tasks/pay/watcher.go
ZenGround0: > We are doing ANY() so it will give us all matches but not exactly all matches.

Sorry I'm not getting it.  Forgetting about the rest of the system entirely I don't understand how we 
1. populate goodSettles with a key set exactly equal to the values of hashes slice
2. select mwes with from the DB such that we are only matching against rows of mwe with signed_tx_hash equal to some value of hashes slice
3. Populate `mwe.Hash` with value of `signed_tx_hash` 
4. expect to ever find a selected mwe where Hash is not found in goodSettles.  

The hashes of mwe's returned from an ANY call should be a strict subset of keys of goodSettles.

ZenGround0: I see we are deleting from goodSettles but signed_tx_hash is primary key on mwe so I don't think we can have duplicate mwe rows

tasks/pay/immediate_settle_task.go
ZenGround0: No worries at all just the normal process of adapting to working together more closely

tasks/pay/immediate_settle_task.go
```diff
@@ -0,0 +1,437 @@
...
+# PDPv0 Client Termination Settlement Races
```
ZenGround0: To expand on why I don't like this a bit more.  The main issue I see is adding a new code path that does something very similar to an existing code path.  When there are two ways to do something it increases the complexity of the system and makes it harder to think through what is happening.  The primary cost is that debugging becomes a little harder and slower for everyone.

My top preference would still be to push back against @rvagg's requirements in #1236. Since we could avoid all this code and instead route final settlment through our existing pipeline I am not seeing the value provided by this new task.  So @rvagg can explain it to us and we can then weigh the cost.  

If we still go down the route of implementing this task then my next preference would be unifying code between here and `SettleLockupPeriod` so that we can at least be sharing settlement triggering code even if we are doing the job in two tasks.  Conceptually this `Do` function and `SettleLockupPeriod` are doing something very similar handling the settlement of rails by retrieving and checking their state and then sending messages and updating the table tracking ongoing settlements to hand off to the watcher.

And finally if we still don't agree on unifying those two methods then I would at least ask that much of this logic be moved to filecoinpayment library.

ZenGround0 (approved):
I am happy enough with this that you should not consider me blocking.  I have some questions on small details that don't matter too much.  And a strong suggestions around avoiding immediate settle task right now at least until we get a strong reason for including this that you buy into from @rvagg.

ZenGround0:
Oh yeah also wrote down the fwss contract upgrade dependency in #1260. We'll want a little patch for keeping things working independent of upgrade / no upgrade before merge to main.

tasks/pay/immediate_settle_task.go
LexLuthr: My preference would be to not have this task and simple let pay handle it. We send the termination message with extra_data and that should stop the payments from client. Client is now free. SP should be allowed to clean up on their own pace. Even if we settle immediately, it doesn't really mean anything in terms of data deletion. Unless settling after 12 hours max has some serious impact, I would prefer to not try to settle immediately.

rvagg: Yeah sorry I'm not deep in this code but I think this comes from what I wrote:

> Tighten the termination-to-deletion latency path: once FWSS termination is confirmed, Curio should promptly attempt/check PDP rail finalisation and move to deletion_allowed as soon as FilecoinPay state permits. Existing periodic settlement/payment watcher/chain-sync paths should remain backstops, not the expected fast path.

What I meant was that we tighten the rules in the existing settle -> cleanup path which has a latency of [7 days in the normal case](https://github.com/filecoin-project/curio/blob/342cbfb03dabab6909e7ee2bd75720b8c7f0af4d/lib/filecoinpayment/utils.go#L87), so it'd just be a matter of just trying to get this data set into that pipeline on the next run so we see settle->fail=cleanup. Or whatever that path is.

rvagg: But honestly, it doesn't really matter all that much, this is SP business not protocol business, I just think that a client saying "please delete" might actually want it to no longer be available for retieval.

LexLuthr: Ahh! I see the confusion. For terminated rails we always try. So once termination message lands on chain termination epoch will be set for that rail. Next periodic settle will process it. So max delay is 12 hours. As for data unavailability that is not going to impacted by this. I have already moved GC to a task and it would be periodic. If we must stop serving - immediately then we should put that check in retrieval and not force Curio to process synchronously.

I am going to remove the individual settle task and it's impact from this PR

tasks/pdpv0/watch_fwss_terminate.go
LexLuthr: My bad! These had more than 1 statement in an earlier version. I should have converted them to tx once I removed those lines

LexLuthr: Fixed

tasks/pay/immediate_settle_task.go
LexLuthr: Removed the task

tasks/pay/watcher.go
LexLuthr: That was a mistake. Fixed. Thanks for catching this.

tasks/pay/immediate_settle_task.go
rvagg: > For terminated rails we always try

aye, I forgot that little helpful fact, so this is all free within the existing system 👌 ; my bad for complexifying the original request

LexLuthr:
@ZenGround0 I have addressed your review in last commit. I am merging this into the feature branch. Feel free to one more pass if you wish. I can address them in a separate follow up PR.

PR #1249, opened 2026-05-22
Shouldtest

### LINKED ISSUE #1247: Gate TestLibProof behind condition
TestLibProof does a realtime C2 without the GPU.
Only if certain things are changed. 

Ticket Importance: Housekeeping. Medium-high Dev impact.

snadrus:
Use a tool that lets us quickly define that some slow tests should first be examined to see if they are likely to have changed. 

Such a tool did not exist stand-alone, so I built one that's currently Go-specific at github.com/curiostorage/shouldtest
It's configured via files in the intended folder & in .github (for "global config").

### DIFF
--- a/.github/shouldtest.json
+++ b/.github/shouldtest.json
@@ -1,0 +1,7 @@
+{
+  "defaultBranch": "main",
+  "alwaysRunOnPushToDefaultBranch": true,
+  "workflows": [".github/workflows/ci.yml"],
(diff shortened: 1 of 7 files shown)


LexLuthr:
But the test ran. Is it supposed to run?

snadrus:
> But the test ran. Is it supposed to run?

Yes because I changed a file in that folder (the definition).

LexLuthr (approved):

PR #1253, opened 2026-05-25
pdpv0: handle PDPVerifier piece cleanup before local deletion

### LINKED ISSUE #1236: PDP: Update pdpv0 termination cleanup state machine for PDPVerifier cleanup mode

LexLuthr:
Fixes https://github.com/filecoin-project/curio/issues/1236

## Summary
- This PR updates the pdpv0 deletion pipeline for the new PDPVerifier cleanup mode.
- After deleteDataSet, Curio no longer treats the data set as locally deletable immediately. It checks PDPVerifier state, detects cleanup mode, and runs cleanupPieces(setId, maxPieces) until on-chain cleanup is finalized.
- The PR adds a cleanupPieces Harmony task and watcher, records cleanup tx hashes, retries cleanup batches, and only proceeds to local data set deletion, local piece cleanup, and IPNI/index cleanup after PDPVerifier reports cleanup complete.
- It also extends deletion recovery paths so chain sync and delete watchers can resume correctly when a data set is already in cleanup mode or cleanup has finalized externally.

- [x] Resolve https://github.com/filecoin-project/curio/issues/1236#issuecomment-4535463297


tasks/pdpv0/task_chain_sync.go
```diff
@@ -132,10 +146,24 @@ func (t *TaskChainSync) syncStaleDeletionTaskIDs(ctx context.Context) error {
 			return false, xerrors.Errorf("failed to clear stale delete data set task ids: %w", err)
 		}
 
-		if terminated > 0 || deleted > 0 {
+		cleanup, err := tx.Exec(`UPDATE pdp_delete_data_set pdds
```
snadrus: In-case it's useful: 
this can automatically be done with

ALTER TABLE pdp_delete_data_set
ADD CONSTRAINT cleanup_pieces_task_id_fkey
FOREIGN KEY (harmony_task_id)
REFERENCES harmony_task(id)
ON DELETE SET NIULL

tasks/pdpv0/task_cleanup_pieces.go
```diff
@@ -0,0 +1,325 @@
...
+		"dataSetId", dataSetID,
+		"txHash", txHash.Hex(),
+		"batchSize", batchSize,
+		"remainingPieceSlots", state.NextPieceID,
+		"refundRecipient", sender.Hex())
+
+	return true, nil
+}
+
+func (t *CleanupPiecesTask) sendCleanupPieces(ctx context.Context, sender common.Address, pdpAddress common.Address, pdpABI *abi.ABI, dataSetID int64, remainingPieces *big.Int) (common.Hash, uint64, error) {
```
snadrus, line 135: add comment

tasks/pdpv0/task_cleanup_pieces.go
```diff
@@ -0,0 +1,325 @@
...
+	}
+
+	if state.Finalized() {
+		if err := cleanupFinalizedDataSet(ctx, t.db, dataSetID); err != nil {
+			return false, err
+		}
+		return true, nil
+	}
+
+	// TODO: Should we just error out instead of resetting state? Could be the result of chain fork/re-org?
```
snadrus: you could break the tie by reading chain state?

tasks/pdpv0/task_cleanup_pieces.go
```diff
@@ -0,0 +1,325 @@
...
+			"err", err)
+		batchSize = nextBatchSize
+	}
+}
+
+func (t *CleanupPiecesTask) schedule(ctx context.Context, addTaskFunc harmonytask.AddTaskFunc) error {
+	var stop bool
+
+	for !stop {
+		addTaskFunc(func(taskID harmonytask.TaskID, tx *harmonydb.Tx) (shouldCommit bool, seriousError error) {
```
snadrus, line 192: We should FUTURE a "multi-delete" concept where the same sql does lists at a time. Especially since CPU:1, we could benefit from a deletion script that gets rid of all marked pieces.

LexLuthr: I am not sure I understand what you are saying. Can you explain?

tasks/pdpv0/task_chain_sync.go
LexLuthr: Agreed. I did not the historical reason for it to be like this. That's why I avoided changing it. I reviewed the entire code and found no valid reason so I will create FKs and indexes.

tasks/pdpv0/task_cleanup_pieces.go
LexLuthr: It is reading chain state. TODO was more internal monologue. I am keeping the code as is and removing TODO.

tasks/pdpv0/task_cleanup_pieces.go
LexLuthr: done

tasks/pdpv0/task_cleanup_pieces.go
snadrus: I'd add:
// FUTURE: # $githubIssueNumber Do deletes in bulk.

tasks/pdpv0/task_chain_sync.go
```diff
@@ -339,6 +306,70 @@ func (t *TaskChainSync) syncMissingDataSetDeleteMessageWaits(ctx context.Context
...
+			SELECT 1
+			FROM message_waits_eth mwe
+			WHERE mwe.signed_tx_hash = pdds.cleanup_pieces_tx_hash
+		  )
+		ORDER BY id
+	`); err != nil {
+		return xerrors.Errorf("failed to select cleanupPieces missing message wait rows: %w", err)
+	}
+
+	if len(missing) == 0 {
```
snadrus: unnecessary

tasks/pdpv0/task_piece_gc.go
```diff
@@ -18,39 +19,64 @@ import (
...
+		db:  db,
+		idx: idx,
+	}
+}
 
-		err := processIndexingAndIPNICleanup(ctx, db, cfg, idx)
-		if err != nil {
-			log.Warnf("Failed to process indexing and IPNI cleanup: %s", err)
-		}
+func (t *PieceGCTask) Do(taskID harmonytask.TaskID, stillOwned func() bool) (done bool, err error) {
```
snadrus, line 47: These names (db & task) are all running together. 
Can this be PcIdxGcTask or similar so a list of task names gives me a good idea of where to look for something in particular?

snadrus (approved):

tasks/pdpv0/task_chain_sync.go
```diff
@@ -99,57 +106,6 @@ func (t *TaskChainSync) Adder(taskFunc harmonytask.AddTaskFunc) {}
 var _ = harmonytask.Reg(&TaskChainSync{})
 var _ harmonytask.TaskInterface = &TaskChainSync{}
 
-// syncStaleDeletionTaskIDs clears deletion task ids that point at Harmony tasks which no longer exist.
-// It does not advance any deletion state; it only unpins rows whose
-// terminate/delete task exhausted retries so the normal schedulers can
```
ZenGround0, line 104: To make sure I'm getting this we don't need this anymore because of the SQL constraints you introduced earlier?  And many of the indexes you also introduced earlier were put in place so that those constraints once triggered could run efficiently?

tasks/pdpv0/task_chain_sync.go
```diff
@@ -40,22 +40,22 @@ func (t *TaskChainSync) Do(taskID harmonytask.TaskID, stillOwned func() bool) (d
...
-	if err := t.syncStaleDeletionTaskIDs(ctx); err != nil {
-		return false, xerrors.Errorf("syncing stale PDP deletion task ids: %w", err)
+	if err := t.syncMissingDataSetTerminationMessageWaits(ctx); err != nil {
+		return false, xerrors.Errorf("syncing missing PDP termination message waits: %w", err)
 	}
 
 	if !stillOwned() {
 		return false, nil
 	}
-	if err := t.syncMissingDataSetTerminationMessageWaits(ctx); err != nil {
```
ZenGround0, line 50: This one looks like it is not being replaced by a direct analog.  I'm trying to figure out why we don't need it anymore.

tasks/pdpv0/task_chain_sync.go
ZenGround0: Let me know if I'm right or wrong.  If I'm getting this right then I much prefer it this way.

tasks/pdpv0/task_terminate_fwss.go
```diff
@@ -110,7 +109,7 @@ func (t *TerminateFWSSTask) Do(taskID harmonytask.TaskID, stillOwned func() bool
 		}
 		data, err = fwssABi.Pack("terminateService", big.NewInt(dataSet.ID), dataSet.ExtraData)
 	} else {
-		data, err = fwssABi.Pack("terminateService", big.NewInt(dataSet.ID))
+		data, err = fwssABi.Pack("terminateService", big.NewInt(dataSet.ID), []byte{})
```
ZenGround0: We don't technically have to update it because of backwards compatibility but since contract is already upgraded and this one is listed as "deprecated" I guess this is preferrable.

tasks/pdpv0/task_delete_data_set.go
```diff
@@ -94,6 +97,7 @@ func (t *DeleteDataSetTask) Do(taskID harmonytask.TaskID, stillOwned func() bool
 		return false, xerrors.Errorf("failed to pack data: %w", err)
 	}
 
+	// deleteDataSet is nonpayable; the cleanup deposit was prepaid at creation.
```
ZenGround0: nit: I don't think this comment is necessary, I wouldn't expect us to need to pay here

tasks/pdpv0/task_cleanup_pieces.go
```diff
@@ -0,0 +1,333 @@
...
+				log.Debugw("no pending PDP data sets for piece cleanup")
+				return false, nil
+			}
+
+			pending := pendings[0]
+
+			n, err := tx.Exec(`UPDATE pdp_delete_data_set
+				SET cleanup_pieces_task_id = $1
+				WHERE id = $2
+				  AND cleanup_pieces_task_id IS NULL
```
ZenGround0, line 222: Why are all these clauses necessary?  I would have thought that since transactions are atomic they also guarantee no state changes in between Execs and so id alone would be enough to perfectly identify the row we want to set.  Is that not true?  So for each Exec we basically need to check that everything is in the state we expect at the given id

ZenGround0 (approved):

tasks/pdpv0/task_chain_sync.go
LexLuthr: Yes. I added constraint to set that column to null automatically when harmony_task row is deleted. Makes much better then us cleaning it manually later and efficient. But, there are cases where we don't want to do this because this can result in continuous retry for certain failure conditions.

tasks/pdpv0/task_cleanup_pieces.go
LexLuthr: https://github.com/filecoin-project/curio/issues/1265 I am not putting TODO in code as this affects multiple tasks. Everything can be cleaned up in a single PR. With harmony changes, I hope we don't hit these funcs too often.

tasks/pdpv0/task_piece_gc.go
LexLuthr: It is called PDPv0_PieceGC task. Internal struct does not matter much as system always spits out the task name for failures and harmony logging

tasks/pdpv0/task_chain_sync.go
LexLuthr: It is there few lines above. You are seeing removal because of how github works. It replaced the syncStaleDeletionTaskIDs line with syncMissingDataSetTerminationMessageWaits

tasks/pdpv0/task_chain_sync.go
LexLuthr: Removed

tasks/pdpv0/task_cleanup_pieces.go
LexLuthr: This is just being extra cautious and redundant even. I created a couple of bugs in past by doing bare minimum so I am extra cautious with SQL matches.

We will be reviewing these as part of https://github.com/filecoin-project/curio/issues/1265 and clean them up as well. I think an AI will shine at this job. It can easily ensure only invariants remains.

tasks/pdpv0/task_delete_data_set.go
LexLuthr: removed

tasks/pdpv0/task_terminate_fwss.go
LexLuthr: Actually, this is incorrect and it will break the compatibility. Correct version is in https://github.com/filecoin-project/curio/pull/1261

ZenGround0:
Merged #1261 in here before doing the final merge to feature branch

PR #1261, opened 2026-05-27
FWSS version gate for client-requested PDPv0 termination

### LINKED ISSUE #1260: pdp: fwss version checking to gracefully handle termination signature change
Today @LexLuthr and I discussed a minor annoyance related to merging `pdpv0-user-termination` branch to `main` after #1246 and #1253 land. fwss service is adding a new termination signature which allows passing extradata for client auth.  We should call the generated data accessor `FilecoinWarmStorageServiceCaller.VERSION` to determine if the contract has upgraded yet and if not fail the user generated termination flow.  The non-user generated non-extradata call is left backwards compatible on the contract so we don't need to touch the provider triggered termination flow.

The other alterative is not merging `pdpv0-user-termination` to main until the fwss contract updates. 

Yet another alternative is merging everything but always failing the user termination as a rule in code and only undoing that in a PR once we know the contract is upgraded.

LexLuthr:
Summary
• Adds FWSS version helpers to detect whether the deployed contract supports client-requested termination.
• Gates client termination at both the HTTP endpoint and task execution before submitting a tx.
• Updates FWSS ABI bindings for the overloaded terminateService methods.
• Uses terminateService0(uint256,bytes) for client-authorized termination and terminateService(uint256) for provider termination.
• Adds tests for FWSS version comparison and ABI overload packing.


pdp/contract/FWSS/utils.go
```diff
@@ -16,3 +22,98 @@ func EnsureServiceTermination(tx *harmonydb.Tx, dataSetID int64) error {
...
+			return true, nil
+		}
+		if current[i] < required[i] {
+			return false, nil
+		}
+	}
+	return false, nil
+}
+
+func parseVersion(version string) ([3]int, error) {
```
ZenGround0: all of this stuff looks good but I have to wonder if there is a semver library that already does this that can keep this code neatly in a dependency.

ZenGround0: yeah it looks like we've got: https://pkg.go.dev/golang.org/x/mod/semver

I'd prefer if we used that since its a golang/x package which is half a step away from being in the stdlib

ZenGround0 (approved):

PR #1262, opened 2026-05-27
friendlier reorg sql

snadrus:
The SQL used in reorg management task is hard. Adding some indexes and making it a UNION (cheap) of 2 SELECT-JOIN-WHERE queries (easy to reason about) should help this considerably.

Discovered by @beck on the latest main. Bug-to-fix in 3 hours, so no issue created.

### DIFF
--- a/harmony/harmonydb/sql/20260527-pdpv0-reorg-chk-indexes.sql
+++ b/harmony/harmonydb/sql/20260527-pdpv0-reorg-chk-indexes.sql
@@ -1,0 +1,21 @@
+-- Indexes for PDPv0 reorg check candidate selection (message_sends_eth + message_waits_eth).
+
+-- Drive branch 1 from confirmed waits past finality depth.
+CREATE INDEX IF NOT EXISTS idx_message_waits_eth_reorg_confirmed
+    ON message_waits_eth (confirmed_block_number, signed_tx_hash)
+    WHERE tx_status = 'confirmed'
+      AND tx_success = TRUE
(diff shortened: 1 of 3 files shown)


harmony/harmonydb/sql/20260527-pdpv0-reorg-chk-indexes.sql
```diff
@@ -0,0 +1,21 @@
+-- Indexes for PDPv0 reorg check candidate selection (message_sends_eth + message_waits_eth).
+
+-- Drive branch 1 from confirmed waits past finality depth.
+CREATE INDEX IF NOT EXISTS idx_message_waits_eth_reorg_confirmed
```
ZenGround0, line 4: Mostly out of curiosity as my sql knowledge is limited: index creation won't incur any big up front or recurring costs for a big unpruned table like message waits eth?

ZenGround0 (approved):
Confirming for my understanding: the issue this is resolving was query latency not query correctness right?

harmony/harmonydb/sql/20260527-pdpv0-reorg-chk-indexes.sql
LexLuthr: Index cost is on writes. But message wait is not pruned frequently so it should be fine

LexLuthr (approved):

LexLuthr:
This was needed for Curio v1.28.2-rc1 fix

Issue #1266, opened 2026-05-30
waiting for send lock Too frequent


beck-8:
~~I'm not sure whether to increase `SendLockedWait`~~ or change this log to debug
```
{"level":"info","ts":"2026-05-30T10:07:07.104+0800","logger":"curio/message","caller":"message/sender_eth.go:97","msg":"waiting for send lock","task_id":14637983,"from":"0x6c8E9cE0a324EC85de5BA9147909B289D56Bd796"}
{"level":"info","ts":"2026-05-30T10:07:07.105+0800","logger":"curio/message","caller":"message/sender_eth.go:97","msg":"waiting for send lock","task_id":14638097,"from":"0x6c8E9cE0a324EC85de5BA9147909B289D56Bd796"}
{"level":"info","ts":"2026-05-30T10:07:07.207+0800","logger":"curio/message","caller":"message/sender_eth.go:97","msg":"waiting for send lock","task_id":14639264,"from":"0x6c8E9cE0a324EC85de5BA9147909B289D56Bd796"}
{"level":"info","ts":"2026-05-30T10:07:07.207+0800","logger":"curio/message","caller":"message/sender_eth.go:97","msg":"waiting for send lock","task_id":14637983,"from":"0x6c8E9cE0a324EC85de5BA9147909B289D56Bd796"}
{"level":"info","ts":"2026-05-30T10:07:07.208+0800","logger":"curio/message","caller":"message/sender_eth.go:97","msg":"waiting for send lock","task_id":14639263,"from":"0x6c8E9cE0a324EC85de5BA9147909B289D56Bd796"}
{"level":"info","ts":"2026-05-30T10:07:07.208+0800","logger":"curio/message","caller":"message/sender_eth.go:97","msg":"waiting for send lock","task_id":14638097,"from":"0x6c8E9cE0a324EC85de5BA9147909B289D56Bd796"}
{"level":"info","ts":"2026-05-30T10:07:07.310+0800","logger":"curio/message","caller":"message/sender_eth.go:97","msg":"waiting for send lock","task_id":14639264,"from":"0x6c8E9cE0a324EC85de5BA9147909B289D56Bd796"}
{"level":"info","ts":"2026-05-30T10:07:07.310+0800","logger":"curio/message","caller":"message/sender_eth.go:97","msg":"waiting for send lock","task_id":14638097,"from":"0x6c8E9cE0a324EC85de5BA9147909B289D56Bd796"}
[...]

PR #1267, opened 2026-05-30
chore: set `waiting for send lock` to debug

beck-8:
https://github.com/filecoin-project/curio/issues/1266


beck-8:
If not satisfied with this. Another way is to define a timeout. If the lock is not obtained after the timeout, a warn will be output.

ZenGround0 (approved):

rvagg (approved):

Issue #1255, opened 2026-05-25
Scale parked_pieces cleanup by replacing anti-join ref discovery


LexLuthr:
## Problem
DropPiece cleanup polls every 5s and finds unreferenced pieces with an anti-join:

```sql
SELECT pp.id
FROM parked_pieces pp
WHERE pp.cleanup_task_id IS NULL
  AND NOT EXISTS (
    SELECT 1 FROM parked_piece_refs ppr
    WHERE ppr.piece_id = pp.id
  );
```

At large parked_pieces sizes, this becomes expensive even when there is nothing to clean up, because cleanup_task_id IS NULL matches most rows.

## Proposal
Track reference count or cleanup eligibility directly on parked_pieces, maintained from parked_piece_refs changes.
Then schedule cleanup from an indexed bounded query:

```sql
SELECT id
FROM parked_pieces
WHERE cleanup_task_id IS NULL
  AND ref_count = 0
ORDER BY id
LIMIT $n;
```

Keep the final delete guarded with NOT EXISTS to protect against concurrent references.

### Acceptance Criteria
• No periodic full-table anti-join over parked_pieces.
• Cleanup scheduling is bounded.
• Ref count or eligibility state is backfilled and maintained correctly.
• Final delete remains race-safe.
• Add tests covering parked-piece creation paths to ensure any production path that creates a parked_pieces row also creates the corresponding parked_piece_refs row in the same transaction or same SQL statement.
• Add migration/test coverage or static validation ensuring new indexes use portable PostgreSQL syntax only, with no Yugabyte-specific HASH, USING lsm, or other YB-only index forms.

ZenGround0:
Since we're seeing issues in #1248 I agree with keeping this for GA 🚀

ZenGround0:
I am a little out of my depth with the sql mechanics so double checking with @snadrus: does this approach seem on the right track?

snadrus:
so the workflow would be:
1. have a trigger on adds & deletes of parked_piece_refs to add or subtract 1 from parked_pieces (an index lookup on id so it's fast).
2. the proposed query
-    - it could use a partial index, like: [ref_count, cleanup_task_id WHERE ref_count=0] making the index tiny

This switches to zero scans for this workflow and adds a "nothing index" (usually empty) and a new column (that could be 16 bits). This looks good to me.

Zen: To use a partial index, you must match the WHERE clause (but you can 'AND' more clauses onto it)

LexLuthr:
[...]

PR #1268, opened 2026-06-01
Cleanup Parked Pieces more efficiently

ZenGround0:
Closing out #1255.


tasks/piece/task_cleanup_piece.go
```diff
@@ -34,21 +34,27 @@ func NewCleanupPieceTask(db *harmonydb.DB, sc *ffi.SealCalls, max int) *CleanupP
 	return pt
 }
 
+// cleanupCandidateBatch caps how many candidate pieces the poller pulls per
+// iteration. The bounded SELECT plus the partial index on
+// (id) WHERE cleanup_task_id IS NULL AND ref_count = 0 keeps each poll cheap
+// regardless of total parked_pieces size.
+const cleanupCandidateBatch = 256
```
ZenGround0, line 41: What do you think about this batch idea?  Nice to have or unnecessary?

harmony/harmonydb/sql/20260601-parked-piece-refcount.sql
```diff
@@ -0,0 +1,102 @@
...
+-- parked_pieces rows have cleanup_task_id IS NULL. With ref_count maintained
+-- by triggers, cleanup picks candidates from a tiny partial index.
+--
+-- Final cleanup DELETE remains guarded by a NOT EXISTS subquery so a race
+-- between candidate selection and deletion cannot drop a piece that just
+-- gained a new ref.
+
+ALTER TABLE parked_pieces ADD COLUMN IF NOT EXISTS ref_count INTEGER NOT NULL DEFAULT 0;
+
+-- Backfill ref_count from existing parked_piece_refs rows.
```
snadrus, line 14: This'll run faster as it gathers the data first then works on that gathered set

WITH ref_counts AS (
    SELECT piece_id, COUNT(*)::int AS cnt
    FROM parked_piece_refs
    GROUP BY piece_id
)
UPDATE parked_pieces pp
SET ref_count = ref_counts.cnt
FROM ref_counts
WHERE pp.id = ref_counts.piece_id
  AND pp.ref_count <> ref_counts.cnt;

tasks/piece/task_cleanup_piece.go
```diff
@@ -34,21 +34,27 @@ func NewCleanupPieceTask(db *harmonydb.DB, sc *ffi.SealCalls, max int) *CleanupP
...
+
 func (c *CleanupPieceTask) pollCleanupTasks(ctx context.Context) {
 	for {
-		// select pieces with no refs and null cleanup_task_id
+		// ref_count is maintained by triggers on parked_piece_refs, so this
+		// query is served by idx_parked_pieces_cleanup_eligible without
+		// scanning parked_piece_refs at all.
 		var pieceIDs []struct {
```
snadrus: // easier to read: 
var pieceIDs []storiface.PieceNumber

If you switch to db.Query() this won't apply anymore

tasks/piece/task_cleanup_piece.go
snadrus: since all this does is create more tasks, the only bloom is a list of IDs (64 bit int?) But this could run on lighter machines and in rare instances maybe it's a risk, so I suppose it's good to have.

snadrus: Note, there's a standard way to handle an unknown, scary count of rows by dealing with them one-by-one and letting hidden layers do the pipelining/buffering:
q, err := db.Query(....)
defer q.Close()
for q.Next() {
   var id storiface.PieceNumber
   err := q.Scan(&id)

snadrus (approved):

tasks/piece/task_cleanup_piece.go
```diff
@@ -64,16 +70,14 @@ func (c *CleanupPieceTask) pollCleanupTasks(ctx context.Context) {
 
 			// create a task for each piece
 			c.TF.Val(ctx)(func(id harmonytask.TaskID, tx *harmonydb.Tx) (shouldCommit bool, err error) {
-				// update
-				n, err := tx.Exec(`UPDATE parked_pieces pp
+				// Atomically claim the piece. ref_count = 0 must still hold;
+				// if a ref was added since the SELECT, the trigger bumped
+				// ref_count and this UPDATE becomes a no-op.
+				n, err := tx.Exec(`UPDATE parked_pieces
```
LexLuthr, line 74: Question - Why should a row have unique task? I could not find any justification for that. There is no reason for a task to not handle multiple pieces.

LexLuthr: You can use ANY($1) where to set all of them at once.

tasks/piece/task_cleanup_piece.go
LexLuthr, line 41: I think we should do much smaller batches of 10 pieces. I am talking about execution batches. Once we set those, scheduling automatically becomes batched as we run only once per poll. This would slow down clean up a bit but it would also pace the load on DB and disk.

itests/park_piece_refcount_test.go
LexLuthr, line 1: Curio does not automatically picks up the itests. You will need to manually add the files to the CI. Some itests should simply be put on other relevant files. I think this one is suitable for `park_piece_test`. Let's avoid new files unless necessary as it does cost us $$ on AWS to run each unique test for each commit in the repo.

LexLuthr (commented):
I will add some details I collected in anticipation of this PR.

1. This PR should only be run on nodes which have unique pieces i.e. https://github.com/filecoin-project/curio/pull/1163 has done it's job. Otherwise, we need to check how that task impacts the table
2. Some DB numbers for reference

```sql
EXPLAIN ANALYZE (join query)Aggregate  (cost=133.34..133.35 rows=1 width=56) (actual time=1235.778..1235.780 rows=1 loops=1)
  Buffers: temp read=9 written=32
  ->  YB Batched Nested Loop Join  (cost=105.00..131.33 rows=200 width=8) (actual time=112.601..1215.715 rows=64510 loops=1)
        Join Filter: (parked_piece_refs.piece_id = pp.id)
        Buffers: temp read=9 written=32
        ->  HashAggregate  (cost=105.00..107.00 rows=200 width=16) (actual time=98.444..130.717 rows=64510 loops=1)
              Group Key: parked_piece_refs.piece_id
              Batches: 5  Memory Usage: 8257kB  Disk Usage: 208kB
              Buffers: temp read=9 written=32
              ->  Seq Scan on parked_piece_refs  (cost=0.00..100.00 rows=1000 width=8) (actual time=1.700..66.158 rows=65803 loops=1)
        ->  Index Scan using parked_pieces_pkey on parked_pieces pp  (cost=0.00..0.13 rows=1 width=8) (actual time=14.344..15.052 rows=1024 loops=63)
              Index Cond: (id = ANY (ARRAY[parked_piece_refs.piece_id, $1, $2, ..., $1023]))
Planning Time: 44.447 ms
Execution Time: 1237.221 ms
Peak Memory Usage: 10073 kB
```

| Query | Result |
|---|---|
| parked_piece_refs | 65,803 |
| distinct_ref_piece_ids | 64,510 |
| parked_pieces | 64,510 |
| active_parked_pieces (cleanup_task_id IS NULL) | 64,510 |



```sql
QUERY PLAN
-----------------------------------------------------------------------------------------------------------------------------------------------------
 Aggregate  (cost=133.34..133.35 rows=1 width=56) (actual time=807.319..807.320 rows=1 loops=1)
   Buffers: temp read=88 written=155
   ->  YB Batched Nested Loop Join  (cost=105.00..131.33 rows=200 width=8) (actual time=82.368..800.805 rows=87300 loops=1)
         Join Filter: (parked_piece_refs.piece_id = pp.id)
         Buffers: temp read=88 written=155
         ->  HashAggregate  (cost=105.00..107.00 rows=200 width=16) (actual time=68.177..84.241 rows=87300 loops=1)
               Group Key: parked_piece_refs.piece_id
               Batches: 5  Memory Usage: 8257kB  Disk Usage: 752kB
               Buffers: temp read=88 written=155
               ->  Seq Scan on parked_piece_refs  (cost=0.00..100.00 rows=1000 width=8) (actual time=1.102..46.484 rows=87366 loops=1)
         ->  Index Scan using parked_pieces_pkey on parked_pieces pp  (cost=0.00..0.13 rows=1 width=8) (actual time=7.573..7.739 rows=1015 loops=86)
               Index Cond: (id = ANY (ARRAY[parked_piece_refs.piece_id, $1, $2, ..., $1023]))
 Planning Time: 40.369 ms
 Execution Time: 807.832 ms
 Peak Memory Usage: 10038 kB
(15 rows)
```

| Query | Result |
|---|---|
| parked_piece_refs | 87366 |
| distinct_ref_piece_ids | 87300 |
| parked_pieces | 87300 |
| active_parked_pieces (cleanup_task_id IS NULL) | 87300 |


3. Based on these number, read is not an issue but write probably will be at scale. You queries should read first then write based on condition holding true even if we are using a transaction.
4. ~I am still waiting for number from provider with 1+ million pieces to understand the impact. I think we should hold off merging till we have those numbers.~ Attached in comment below

LexLuthr (commented):
We are also missing both itests requested in the issue. 

> - Add tests covering parked-piece creation paths to ensure any production path that creates a parked_pieces row also creates the corresponding parked_piece_refs row in the same transaction or same SQL statement.
> - Add migration/test coverage or static validation ensuring new indexes use portable PostgreSQL syntax only, with no Yugabyte-specific HASH, USING lsm, or other YB-only index forms.

harmony/harmonydb/sql/20260601-parked-piece-refcount.sql
snadrus: done

tasks/piece/task_cleanup_piece.go
snadrus: done

tasks/piece/task_cleanup_piece.go
snadrus: lets leave it alone for urgency concerns.

snadrus: skipping for now

tasks/piece/task_cleanup_piece.go
snadrus: done

tasks/piece/task_cleanup_piece.go
snadrus: task-per-piece will allow us to run it on the right machine

LexLuthr:
```sql
Aggregate  (cost=273082.85..273082.86 rows=1 width=56) (actual time=29040.068..29040.070 rows=1 loops=1)
  Buffers: temp read=3925 written=7605
  ->  YB Batched Nested Loop Join  (cost=170597.50..265566.68 rows=751617 width=8) (actual time=1719.062..28727.787 rows=1096270 loops=1)
        Join Filter: (parked_piece_refs.piece_id = pp.id)
        Buffers: temp read=3925 written=7605
        ->  HashAggregate  (cost=170597.50..186643.55 rows=751617 width=16) (actual time=1701.422..2727.473 rows=1096270 loops=1)
              Group Key: parked_piece_refs.piece_id
              Planned Partitions: 16  Batches: 81  Memory Usage: 8337kB  Disk Usage: 31520kB
              Buffers: temp read=3925 written=7605
              ->  Seq Scan on parked_piece_refs  (cost=0.00..109182.40 rows=1091824 width=8) (actual time=1.803..1082.732 rows=1108031 loops=1)
        ->  Index Scan using parked_pieces_pkey on parked_pieces pp  (cost=0.00..0.11 rows=1 width=8) (actual time=21.594..22.248 rows=1024 loops=1071)
              Index Cond: (id = ANY (ARRAY[parked_piece_refs.piece_id, $1, $2, ..., $1023]))
Planning Time: 42.850 ms
Execution Time: 29043.980 ms
Peak Memory Usage: 11358 kB
```

| Query |	Result |
|---|---|
| parked_piece_refs |	1,108,032 |
| distinct_ref_piece_ids |	1,096,271 |
| parked_pieces |	1,096,271 |
| active_parked_pieces (cleanup_task_id IS NULL) |	1,096,271 |

itests/park_piece_refcount_test.go
snadrus: moving. also, adding a guard so we catch this before review time.

snadrus:
> * Add tests covering parked-piece creation paths to ensure any production path that creates a parked_pieces row also creates the corresponding parked_piece_refs row in the same transaction or same SQL statement.

I cannot test code that doesn't exist. We can't "find and run all sql that touches X".  We could:
A.  add a Postgres CHECK Constraint to ensure the refcount is numbered right. It will have runtime performance consequences. 
B. Run something at the end of every itest to validate invariants (if that new code gets tests). 

This may be asking for too much. New code has examples of how to do things right. 

> Add migration/test coverage or static validation ensuring new indexes use portable PostgreSQL syntax only, with no Yugabyte-specific HASH, USING lsm, or other YB-only index forms.

CI already runs against Postgres. Those validations already occur. We could make a unit test against a Postgres grammar, but the developer will likely experience problems in the same place: CI. We cannot shorten the time-to-discovery here.

LexLuthr (approved):
I am good with changes around clean up but I still don't see the requested itest ensuring devs are not creating park_piece without refs

snadrus:
Please see my comment. That's not viable

Andrew Jackson
President of Curio Storage, Inc. <http://curiostorage.org>

beck-8:
The reason for this discovery is that my machine's database CPU has remained at 100-150% for a long time. This is the result of Codex's investigation
  
  PR #1268 introduced the idx_parked_pieces_cleanup_eligible partial index. On YugabyteDB 2.25.1.0, this index can return stale cleanup candidates. The base table rows already have ref_count=1, but the query
  WHERE cleanup_task_id IS NULL AND ref_count=0 still returns them when it uses this index.

  As a result, the DropPiece cleanup task repeatedly selects false cleanup candidates, then UPDATE ... WHERE ref_count=0 affects 0 rows, the transaction rolls back, and the poller immediately loops again. This
  creates a high-frequency tight loop and drives yb-tserver CPU high.

PR #1271, opened 2026-06-03
fix(indexstore): add ALLOW FILTERING to GetPDPLayerIndex for Cassandra/Scylla

rvagg:
Originally came up in https://github.com/filecoin-project/curio/pull/1094 as the main piece needed for an independent CQL backend since Yugabyte accepts this but neither Cassandra or Scylla do. Now that we seem to have Postgres support back in https://github.com/filecoin-project/curio/pull/1159 I'd really like to use that an Scylla in a devnet because it's sooo much faster and they have a smaller memory footprint together.

The itests hit this path, but because Yugabyte is still in the CI image, it's talking to YCQL on port 9042. I'm not sure whether that's intentional or not, I'll leave that to you guys, I mainly just want this change.

### DIFF
--- a/market/indexstore/indexstore.go
+++ b/market/indexstore/indexstore.go
@@ -747,7 +747,9 @@ func (i *IndexStore) AddPDPLayer(ctx context.Context, pieceCidV2 cid.Cid, layer 
 
 func (i *IndexStore) GetPDPLayerIndex(ctx context.Context, pieceCidV2 cid.Cid) (bool, int, error) {
 	var layerIdx int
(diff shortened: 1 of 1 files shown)


LexLuthr:
I want to keep CQL coming from YB. We keep running into problem where Postgres specific things pass the test and then fail the production deployment on YB. With switch to Postgres, we need to be extremely careful with SQL semantics. I don't think adding one more burden on devs is a good move.

LexLuthr (approved):
This is fine since YB silently ignores the last part.

rvagg:
nice 👌

PR #1277, opened 2026-06-04
fix: handle finalized PDP cleanup and gas retries

beck-8:
Fixes two PDPv0 cleanup failure cases observed in calib.

Errors:

```text
Failed to process pending PDP piece cleanup: failed to process PDP cleanup txs: failed to read PDP cleanup state for data set 58: failed to get next piece id for data set 58: chain: message execution failed (exit=[33], revert reason=[0x9a83803e], vm error=[message failed with backtrace:
00: f0176630 (method 3844450837) -- contract reverted at 75 (33)
01: f0176630 (method 6) -- contract reverted at 6499 (33)
 (RetCode=33)])
```

```text
error: failed to send cleanupPieces transaction: failed to estimate gas: chain: message execution failed (exit=[33], revert reason=[none], vm error=[message failed with backtrace:
00: f0176630 (method 3844450837) -- contract reverted at 75 (33)
01: f0176630 (method 6) -- out of gas (7)
 (RetCode=33)])
```

Changes:

- Treat PDPVerifier `DataSetNotFound()` as finalized cleanup state in the PDPv0 deletion pipeline.
- Retry `cleanupPieces` with a smaller batch when Lotus returns `out of gas (7)` during gas estimation.

### DIFF
--- a/tasks/pdpv0/error_detection.go
+++ b/tasks/pdpv0/error_detection.go
@@ -4,6 +4,7 @@ import (
 	"encoding/hex"
 	"strings"
 
+	"github.com/filecoin-project/curio/pdp/contract"
 	"github.com/filecoin-project/curio/pdp/contract/FWSS"
 )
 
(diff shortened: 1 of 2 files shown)


LexLuthr (approved):
I think first IsPDPVerifierDataSetNotFound is enough and other 2 are mostly overkill. But, overkill is better than missing it.

PR #1278, opened 2026-06-04
fix(curio): detect FWSS client termination support correctly

rvagg:


rvagg:
sorry, should have left a description: the `>1` means the check never passes, `Compare` is a -1,0,1 check

LexLuthr (approved):

PR #1280, opened 2026-06-05
Attempt handling stale idx results

ZenGround0:
Attempt to address: https://github.com/filecoin-project/curio/pull/1268#issuecomment-4619877224

### DIFF
--- a/tasks/piece/task_cleanup_piece.go
+++ b/tasks/piece/task_cleanup_piece.go
@@ -35,77 +35,72 @@ func NewCleanupPieceTask(db *harmonydb.DB, sc *ffi.SealCalls, max int) *CleanupP
 }
 
 // cleanupCandidateBatch caps how many candidate pieces the poller pulls per
-// iteration. The bounded SELECT plus the partial index on
-// (id) WHERE cleanup_task_id IS NULL AND ref_count = 0 keeps each poll cheap
-// regardless of total parked_pieces size.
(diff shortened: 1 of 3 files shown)


tasks/piece/task_cleanup_piece.go
```diff
@@ -59,12 +59,26 @@ func (c *CleanupPieceTask) pollCleanupTasks(ctx context.Context) {
...
-		if len(pieceIDs) == 0 {
+		var pieceIDs2 []storiface.PieceNumber
+
+		// Now check the table directly to handle the stale index case
+		err = c.db.Select(ctx, &pieceIDs, `WITH checked AS MATERIALIZED (
+    			SELECT id, ref_count, cleanup_task_id
+    			FROM parked_pieces
+    			WHERE id = ANY($1::bigint[])    
+			)
+			SELECT id
```
ZenGround0: If I am getting this right then this query will be guaranteed to see the real data.  So this will filter away the stale rows from the partial index and leave us with only correct rows matching the predicate.

ZenGround0:
At this point I don't know if keeping the partial index is worth it.  If there are many many stale rows it somewhat defeats the purpose.

beck-8:
Tested well. The load is reduced. The droppiece tasks accumulated before are now available.

LexLuthr (approved):
Fixed with latest commit. Confirmed in test with @beck-8.

Issue #1251, opened 2026-05-25
PDPv0: pdp_pieceref table must have unique entry per new piece


LexLuthr:
## Problem
Current, we add a piece to Curio and then create an entry in pdp_pieceref table. This entry acts as our state machine as well as piece reference. Problem is that we create these entries uniquely for each piece but we don't use them deterministically. So, I might create a new entry but decide to use an old one. This creates chaos as Curio now has to clean up this new entry without it ever being used.
This creates a lot of DB churn and very complicated piece cleanup pipeline. 

## Solution 1
We should bind the piece lifetime to the pdp_pieceref which was created for it.
Part of this would require creating new pdp_pieceref and migrating the (dataSet,piece) table details to use this unique ref

## Solution 2
A piece is a piece. We should only have a pdp_pieceref per unique piece_cid_v2. So, we never do any changes to the DB as long as we 1 dataSet using that piece. This works when we are expecting a big overlap in data. But, this will require even more complicated state machine mechanics and extremely careful cleanup.

ZenGround0:
It seems like a problem here is that we are condensing two things: "our state machine" and "piece reference".  If we seperate them out then perhaps "even more complicated state machine mechanics and extremely careful cleanup" can go away?  The structure of pdp_piecrefs table already seems to handle cleanup very simply with a dataset refcount and is therefore suggesting the one-pieceref-per-piece design of solution 2.  I suspect this is the direction we should lean into. 

This looks like the sort of issue where some targeted thinking about the design will greatly clean up the code.  The solution we arrive at will probably be hard to guess at before we do all that thinking.

LexLuthr:
# Proposed direction for solution 2:

Make `pdp_piecerefs` the canonical PDPv0 piece row, with one row per unique PieceCIDv2.

Today the table stores `piece_cid` as PieceCIDv1 and points at a `parked_piece_refs` row. That is not enough to express the intended uniqueness because PieceCIDv2 depends on both the v1 commitment and raw size. We should store `piece_cid_v2` directly on `pdp_piecerefs` and enforce uniqueness on that.

## Required changes:

1. Schema
   - Add `pdp_piecerefs.piece_cid_v2 TEXT`.
   - Backfill it from `pdp_piecerefs.piece_cid` plus `parked_pieces.piece_raw_size`.
   - Add a unique index/constraint on `piece_cid_v2`.
   - Keep `piece_cid` as v1 for existing readers unless we want a larger cleanup later.

[...]

PR #1290, opened 2026-06-09
fix/upload-to-bad-dataset

snadrus:
This is a "stop the bleeding" fix to resolve a DDoS. 
Today's flow:
1. POST /pdp/piece  +  PUT /pdp/piece/upload/{uuid}   → succeeds (bytes parked)
2. PDPv0_Notify (async, ~2s)       → creates pdp_piecerefs
3. POST /pdp/data-sets/{id}/pieces (×22–54 datasets) → 404/409, no-op
4. GC waits 24 hours

Quick-Fix for hammering:
- Step 3 should mark the piece for the delete poller to delete 
- Ban IP from all APIs for 5 minutes if Step-3 sees 5 problems in a minute. 

FUTURE (not here): This API should be reversed: 
1. Validate the wallet & DataSet is good. Get a token
2. Do an upload that waits on a task, gets a CommP and finishes.

### DIFF
--- a/pdp/dataset_verify.go
+++ b/pdp/dataset_verify.go
@@ -44,3 +44,51 @@ func verifyDataSetForService(ctx context.Context, db *harmonydb.DB, service stri
 
 	return nil
 }
+
+// discardOrphanPiecrefsForSubPieces removes unreferenced pdp_piecerefs for the
+// given subPiece CIDs. Called when addPieces is rejected for a missing or
+// terminated data set, so notify-created piecerefs do not linger.
+func discardOrphanPiecrefsForSubPieces(ctx context.Context, db *harmonydb.DB, service string, subPieceCidV1List []string) error {
+	if len(subPieceCidV1List) == 0 {
+		return nil
+	}
+
+	_, err := db.BeginTransaction(ctx, func(tx *harmonydb.Tx) (bool, error) {
+		n, err := tx.Exec(`
+			WITH doomed AS (
+				SELECT pr.id, pr.piece_ref
+				FROM pdp_piecerefs pr
+				WHERE pr.service = $1
+				  AND pr.piece_cid = ANY($2)
+				  AND pr.data_set_refcount = 0
+				  AND NOT EXISTS (
+					SELECT 1 FROM pdp_data_set_piece_adds a
+					WHERE a.pdp_pieceref = pr.id
+					  AND a.pieces_added = FALSE
+					  AND (a.add_message_ok IS NULL OR a.add_message_ok = TRUE)
+				  )
+			),
+			deleted AS (
+				DELETE FROM pdp_piecerefs pr
+				USING doomed d
+				WHERE pr.id = d.id
+				RETURNING d.piece_ref AS piece_ref
+			)
+			DELETE FROM parked_piece_refs ppr
+			USING deleted d
+			WHERE ppr.ref_id = d.piece_ref
+			  AND NOT EXISTS (SELECT 1 FROM pdp_piecerefs pr WHERE pr.piece_ref = ppr.ref_id)
+		`, service, subPieceCidV1List)
+		if err != nil {
+			return false, fmt.Errorf("discard orphan piecerefs: %w", err)
+		}
+		if n > 0 {
+			log.Infow("discarded orphan PDP piecerefs after bad data set addPieces",
+				"service", service,
+				"subPieceCount", len(subPieceCidV1List),
+				"parkedRefsRemoved", n)
+		}
+		return true, nil
+	}, harmonydb.OptionRetry())
+	return err
+}
--- a/pdp/dataset_verify.go
+++ b/pdp/dataset_verify.go
@@ -1,0 +1,46 @@
+package pdp
+
+import (
+	"context"
+	"errors"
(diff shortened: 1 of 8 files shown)


pdp/ip_offense_throttle.go
```diff
@@ -0,0 +1,219 @@
+package pdp
+
```
ZenGround0, line 2: Wait do we need to reimplement this ourselves?  @LexLuthr had mentioned that there are a few lines of configuration code that can do something similar that he added for mkv2 http handler

pdp/dataset_verify.go
```diff
@@ -0,0 +1,94 @@
...
+	// ErrDataSetNotFound indicates the data set does not exist or does not belong to the service.
+	ErrDataSetNotFound = errors.New("data set not found")
+	// ErrDataSetTerminated indicates the data set was terminated due to unrecoverable proving failure.
+	ErrDataSetTerminated = errors.New("data set has been terminated due to unrecoverable proving failure")
+)
+
+// verifyDataSetForService checks that dataSetId exists in pdp_data_sets, belongs to service,
+// and has not been terminated due to unrecoverable proving failure.
+func verifyDataSetForService(ctx context.Context, db *harmonydb.DB, service string, dataSetId uint64) error {
+	var dataSetService string
```
ZenGround0, line 23: This is good but I would probably like to see this do a ethcall to read chain state and further fail if the dataset is not live from the chain's perspective.

pdp/dataset_verify.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	_, err := db.BeginTransaction(ctx, func(tx *harmonydb.Tx) (bool, error) {
+		n, err := tx.Exec(`
+			WITH doomed AS (
+				SELECT pr.id, pr.piece_ref
+				FROM pdp_piecerefs pr
+				WHERE pr.service = $1
+				  AND pr.piece_cid = ANY($2)
+				  AND pr.data_set_refcount = 0
```
ZenGround0, line 63: I was a little worried about a race between good addPieces and spam addPieces calls referencing the same pdp_pieceref given that inserts into `pdp_data_set_piece_adds` happen before refcount gets incremented.  But I think we are safe because different uploads of the same cid actually create multiple pdp_piecerefs.

ZenGround0: Actually I think this *is* a problem because the add handler doesn't actually take in a piece id it figures out which pdp_piecerefs to get by matching cids between the add piece info and the cid column of pdp_piecerefs.  This is a mess and cleanup of this part of the system is tracked here #1251.  For now we could just filter out any piecrefs that have successfully landed in `pdp_data_set_piece_adds`

pdp/ip_offense_throttle.go
ZenGround0: Yeah I'm seeing this here already: https://github.com/filecoin-project/curio/blob/main/market/mk20/http/http.go#L52-L53.  Might be a bit less clever but its nice to keep something like middleware ip throttling outside of our circle of maintenance if possible.

pdp/dataset_verify.go
snadrus: Yes, that's a tiny-likelihood attack on your neighbor that gets you banned quickly. On the other hand, we have a hot DDoS, so that seems like a better bet until we rewire this entirely.

pdp/dataset_verify.go
snadrus: DDoS'ers want to overwhelm your system. The goal is to not give them much reach until "authenticated". If this DB is unaware of the dataset, it should be good enough to say this piece doesn't belong in this system.

pdp/ip_offense_throttle.go
snadrus: to be fair, I built a ban, not a rate-limit.

pdp/dataset_verify.go
ZenGround0: Sure I'm just saying it would be better to narrow it down further.  I guess your argument is that this narrowing down is too expensive.  Lets start with this then.

pdp/dataset_verify.go
ZenGround0: Yes but its pretty easy to filter out right? Are you saying that filtering against `pdp_data_set_piece_adds` is too expensive ?  I feel like it is totally acceptable to enforce correctness.

ZenGround0: sorry this is my bad I don't know how I missed the pdp_data_set_piece_adds check below you already accounted for this.

pdp/ip_offense_throttle.go
```diff
@@ -0,0 +1,219 @@
...
+func defaultIPOffensePolicies() map[string]IPOffensePolicy {
+	return map[string]IPOffensePolicy{
+		OffenseBadDataSetAdd: {
+			Hits:   5,
+			Window: time.Minute,
+			Block:  5 * time.Minute,
+			Reason: "too many requests for unavailable data sets",
+		},
+	}
```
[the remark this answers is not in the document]
snadrus: desc cgange

ZenGround0 (approved):
Thanks LGTM

Issue #1291, opened 2026-06-09
pdp: PDPv0_IPNI Max(50) storms YugabyteDB with 40001 serialization conflicts on the serial ad-chain head; dropped tasks strand piecerefs unannounced (cc405c39)


TippyFlitsUK:
## What happened

After upgrading our mainnet PDP node to HEAD `cc405c39` ("more parallel IPNI tasks") and resuming a large migration ingest, `PDPv0_IPNI` began failing at ~90 errors/min - 100% YugabyteDB `SQLSTATE 40001` (`could not serialize access due to concurrent update`) on the advertisement write. Over one hour: 4,999 failed attempts vs 2,197 succeeded (~70% per-attempt failure). The failed transactions (~5k/hr) also loaded the same tablets our migration commit path uses, and commit throughput visibly stagnated until the storm was stopped.

The worse half is silent: with `MaxFailures: 3`, failing tasks are dropped from `harmony_task`, and `pdp_piecerefs.ipni_task_id` has no FK back to `harmony_task` (no `ON DELETE SET NULL`). A dropped task leaves its pieceref with `needs_ipni=TRUE` and `ipni_task_id` pointing at a dead task id. `schedule()` only claims rows `WHERE ipni_task_id IS NULL`, so those pieces are **never announced to IPNI** - we accumulated **1,223 stranded piecerefs at ~1,117/hr** while the task queue itself looked healthy (dropped tasks leave the queue, so queue depth hides the loss). Same theme as #1248: the failure mode is invisible to task-queue dashboards.

## What we found

- Every failure is the same write: `insert_ad_and_update_head_checked` advances the single per-provider advertisement-chain head, so the work is inherently serial. `cc405c39` sets `Max: taskhelp.Max(50)` directly in `TypeDetails()` (`tasks/indexing/task_pdp_v0_ipni.go`), so up to 50 writers contend on that one head row. The constructor still accepts an injected limiter (`P.max`) but `TypeDetails()` no longer uses it.
- The 40001s carry YugabyteDB's "query layer retry isn't possible because data was already sent" variant, so they cannot be auto-retried server-side - `Do()` fails outright and burns one of the 3 attempts.
- We had hit the same contention the previous day at lower concurrency (`PDPv0_IPNI` sharing `idxMax` with the indexing tasks): 15% failure rate, 9.4s avg task time vs 1.5s for neighbours.

Diagnostic queries we used:

```sql
-- stranded piecerefs (never announced, invisible to the scheduler)
[...]

PR #1293, opened 2026-06-10
pdp: fix PDPv0_IPNI throughput collapse under load

TippyFlitsUK:
Fixes the `PDPv0_IPNI` throughput collapse diagnosed in #1291 (see the [follow-up comment](https://github.com/filecoin-project/curio/issues/1291#issuecomment-4672460291) for the full investigation and measurements).

## Changes

**1. Shrink the `ipni_head` conflict window** (`tasks/indexing/task_pdp_v0_ipni.go`)
The write transaction previously read the head and then performed the entire Go-side ad build - key fetch, construction, URL resolution, signing, validation, serialisation - before calling `insert_ad_and_update_head_checked`. That held a read dependency on the hottest row in the DB for seconds per task, so any concurrency produced YugabyteDB `40001` serialization aborts (which burn `MaxFailures` lives and can strand piecerefs). The SQL function is already a complete CAS via `_expected_previous`, so this PR moves the head read, key read, build and signing outside the transaction; the tx is now just the CAS call plus the completion update. A lost race is the function's clean `FALSE` return and a cheap retry instead of an abort. Measured: 6.4% task failure rate → 0% at concurrency 5.

**2. Release the pieceref claim on the already-published early exit**
The exit returned success without clearing `ipni_task_id`; `schedule()` only claims rows where it is NULL, so the pieceref was stranded with `needs_ipni=TRUE` forever.

**3. Hot-path indexes** (`20260610-pdp-ipni-hot-path-indexes.sql`)
- `pdp_piecerefs(ipni_task_id)` and `(indexing_task_id)` partial over `IS NOT NULL`: the per-task params lookup was a full scan (measured 1,397ms at 1.4M rows, per task, for both `PDPv0_IPNI` and `PDPv0_Indexing`) → 64ms.
- `ipni(piece_cid, order_number DESC)`: the already-published pre-check grew linearly with published ads (291ms at 134k rows) → 36ms.
- `IF NOT EXISTS` so nodes that created these manually are unaffected.

## Measured result

Mainnet node draining a 47k announcement backlog: **~13/min → ~1,600/min, zero failures**, currently running in production on `1.28.2-rc2+mainnet+git_cc405c39`. Operator note for YugabyteDB: a running node may need `ANALYZE` on the touched tables before cached plans pick up the new indexes.

Not included (left as findings in #1291): the harmonytask poller full-queue scan - engine-wide, interacts with retry-wait scheduling, felt like your call on shape.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tasks/indexing/task_pdp_v0_ipni.go
+++ b/tasks/indexing/task_pdp_v0_ipni.go
@@ -95,6 +95,15 @@ func (P *PDPV0IPNITask) Do(taskID harmonytask.TaskID, stillOwned func() bool) (d
 	}
 
 	if exists && !isRm {
+		_, err = P.db.BeginTransaction(ctx, func(tx *harmonydb.Tx) (commit bool, err error) {
+			if err := P.recordCompletion(tx, taskID, task.ID); err != nil {
+				return false, xerrors.Errorf("recording IPNI task completion: %w", err)
+			}
+			return true, nil
+		})
+		if err != nil {
+			return false, err
+		}
 		ilog.Infow("IPNI task already published", "task_id", taskID, "piece_cid", task.PieceCID)
 		return true, nil
 	}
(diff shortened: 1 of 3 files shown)


LexLuthr (changes requested):
Please rebase your commit on main. This is bringing in the storacha migration tool which is not required.

tasks/indexing/task_pdp_v0_ipni.go
```diff
@@ -181,77 +193,87 @@ func (P *PDPV0IPNITask) Do(taskID harmonytask.TaskID, stillOwned func() bool) (d
...
+			return false, xerrors.Errorf("converting advertisement to node: %w", err)
+		}
 
+		ad, err := ipniculib.NodeToLink(adNode, schema.Linkproto)
+		if err != nil {
+			return false, xerrors.Errorf("converting advertisement to link: %w", err)
+		}
+
+		// Short write transaction: a single CAS-checked insert plus the pieceref
+		// completion update. Conflict window is now milliseconds.
```
LexLuthr: Remove AI comments

tasks/indexing/task_pdp_v0_ipni.go
LexLuthr: Retry is not needed with newer code

tasks/indexing/task_pdp_v0_ipni.go
```diff
@@ -181,77 +193,87 @@ func (P *PDPV0IPNITask) Do(taskID harmonytask.TaskID, stillOwned func() bool) (d
...
-			pkey, err := crypto.UnmarshalPrivateKey(privKey)
-			if err != nil {
-				return false, xerrors.Errorf("unmarshaling private key: %w", err)
-			}
+		// Read the current head and signing key, build and sign the advertisement,
+		// all OUTSIDE the write transaction. insert_ad_and_update_head_checked is
+		// CAS-safe on its own (_expected_previous): if the head moved, it returns
+		// FALSE and we loop. Holding the head read across the Go-side ad build and
+		// signature kept a seconds-long conflict window on the hottest row in the
+		// DB, which is what produced YugabyteDB 40001 serialization storms under
```
LexLuthr: This comment is problem specific. In fact, most of this is useless detail.

LexLuthr (commented):
Looks good. Needs some minor cleanup along with rebase.

TippyFlitsUK:
Apologies Lex - branched from my node's local HEAD which was sitting on the rebased migration branch rather than main. Rebuilt: now a single commit on top of current main (99c899ee), just the task file + the index migration. :+1:

tasks/indexing/task_pdp_v0_ipni.go
TippyFlitsUK: Done

tasks/indexing/task_pdp_v0_ipni.go
TippyFlitsUK: Removed from both call sites.

tasks/indexing/task_pdp_v0_ipni.go
TippyFlitsUK: Done

LexLuthr:
> Apologies Lex - branched from my node's local HEAD which was sitting on the rebased migration branch rather than main. Rebuilt: now a single commit on top of current main ([99c899e](https://github.com/filecoin-project/curio/commit/99c899ee15907ea76890acc476b4d1868d87ed8e)), just the task file + the index migration. 👍

No need to apologise. This is a pretty common mistake. 😉

LexLuthr (approved):

LexLuthr:
@TippyFlitsUK Hey! You need to fix the test by creating a downgrade SQL file.

TippyFlitsUK:
Done - downgrade file added. :+1:

PR #1297, opened 2026-06-11
fix: make StorePiece min free storage percent configurable

beck-8:
StorePiece task used a hardcoded 10% min free storage requirement, so it stops scheduling once storage is over 90% full.

Use `ParkPieceMinFreeStoragePercent` (default 5, settable to 0) instead, same as ParkPiece.


beck-8:
`ParkPieceMinFreeStoragePercent` Might this name cause ambiguity?

LexLuthr (approved):

LexLuthr:
> `ParkPieceMinFreeStoragePercent` Might this name cause ambiguity?

It is fine. Both are actually parkPiece tasks and common config name in a subsystem is expected.

ZenGround0 (approved):

PR #1299, opened 2026-06-11
pdpv0: fix handler

### LINKED ISSUE #1300: PDP: GET /pdp/piece (find-piece) returns 405 after route refactor in #1290 — breaks all synapse-sdk uploads
## Summary

#1290 refactored the PDP route registration in `pdp/handlers.go`. Before, the find-piece route was registered as:

```go
r.Get(path.Join(PDPRoutePath, "/piece/"), p.handleFindPiece)
```

`path.Join` strips the trailing slash, so the effective route was **`/pdp/piece`** (no trailing slash). After #1290 it became a literal pattern inside the `/pdp` route group:

```go
r.Get("/piece/", p.handleFindPiece)   // pdp/handlers.go:193 on main
```

chi treats `/pdp/piece/` (trailing slash) as a distinct route from `/pdp/piece`, and the only method registered at `/pdp/piece` is `POST` (pieceUploadInit). Result: `GET /pdp/piece?pieceCid=...` now returns **405 Method Not Allowed**.

## Impact

synapse-sdk (all versions, including the released 0.42.0) calls exactly that form after every upload to confirm the piece is parked ([`find-piece.ts`](https://github.com/FilOzone/synapse-sdk/blob/master/packages/synapse-core/src/sp/find-piece.ts)):

```
GET /pdp/piece?pieceCid=<cid>
```

so **every SDK upload fails** against a curio built from current main, even though the piece data is stored fine.

## Reproduction (live, calibnet SPs running current main)

```
$ curl -s -o /dev/null -w "%{http_code}\n" "https://pdp.660688.xyz:8443/pdp/piece?pieceCid=bafkzcib..."
405
$ curl -s "https://pdp.660688.xyz:8443/pdp/piece/?pieceCid=bafkzcib..."
{"pieceCid":"bafkzcib..."}   # 200 — same piece, trailing-slash form
```

LexLuthr:
Fixes https://github.com/filecoin-project/curio/issues/1300

### DIFF
--- a/pdp/handlers.go
+++ b/pdp/handlers.go
@@ -189,8 +189,8 @@ func Routes(r *chi.Mux, p *PDPService) {
 		// POST /pdp/piece
 		r.Post("/piece", instrument("pieceUploadInit", p.handlePiecePost, nil))
 
-		// GET /pdp/piece/
-		r.Get("/piece/", p.handleFindPiece)
+		// GET /pdp/piece
+		r.Get("/piece", p.handleFindPiece)
 
 		// PUT /pdp/piece/upload/{uploadUUID}
 		r.Put("/piece/upload/{uploadUUID}", instrument("pieceUpload", p.handlePieceUpload, kvUploadUUID))


BigLep:
Is there a place to add a regression test?

LexLuthr:
> Is there a place to add a regression test?

No yet. Testing these is tricky as they directly interact with smart contracts so mocking won't work. I do have a way to test PDP in a local branch but it needs to be rebased and updated.

BigLep:
> > Is there a place to add a regression test?
> 
> No yet. Testing these is tricky as they directly interact with smart contracts so mocking won't work. I do have a way to test PDP in a local branch but it needs to be rebased and updated.

Ack, ok. I realize we need to get the fix out to not block release, but I want to make sure we don't lose track of having the ability to add regression tests here. Maybe create the tracking issue for the regression test (depending on your branch) so that we have visibility for this as we do Q3 planning?

ZenGround0 (approved):

PR #1301, opened 2026-06-11
Message replacement subsystem (replace by fee)

### LINKED ISSUE #1292: Curio should be able to replace message when there is a baseFee spike
When there is a baseFee spike, messages simply get stuck in the mpool and this basically kills all pipelines in Curio.

### LINKED ISSUE #1354: pdp: eth sender flattens EIP-1559 transactions to legacy at nonce assignment - every message pays its full fee cap and strands on any baseFee rise
### Checklist

- [x] This is **not** a question or a support request.
- [x] This is **not** a new feature request.
- [x] This is **not** an enhancement request.
- [x] I **have** searched on the issue tracker and there is no existing related issue (closest are #1292 and #1313, both referenced below).
- [x] I am running the most recent RC.
- [ ] I did not make any code changes to curio. *(We are running local mitigations on both nodes, detailed below, deployed after the evidence was gathered.)*

### Curio component

- [x] Other (eth message sender - `tasks/message/sender_eth.go`)

### Curio Version

```
curio version 1.28.2-rc2+mainnet+git_c98f048c_2026-06-12T00:34:03+00:00
```

All three defects are also present in `tasks/message/sender_eth.go` on current `main` (blob 4f49253).

### Describe the Bug

To share the load and speed up the Storacha migration I spun up a second mainnet PDP node alongside our existing one. It initially did more harm than good - between them the two nodes generated enough gas demand to drive the baseFee up against their own messages, and both started stranding. Profiling why led us to three compounding defects in the eth sender:

1. **EIP-1559 transactions are silently downgraded to legacy at nonce assignment.** `Send()` builds a correct `DynamicFeeTx` (cap + tip), but the nonce-assignment step rebuilds it with `types.NewTransaction(assignedNonce, ..., tx.GasPrice(), ...)` - the legacy constructor. geth's `GasPrice()` returns the **fee cap** for a dynamic-fee tx, so the cap/tip split is collapsed into a single legacy gasPrice equal to the cap. Every PDP message on chain is `type: 0x0`.

2. **The sender always pays its full cap.** Lotus maps legacy txs as `GasFeeCap = GasPremium = gasPrice`, so effective premium = cap minus baseFee, charged on the full gas limit. Six sampled receipts show `effectiveGasPrice` 4-7% **above** the tx gasPrice (premium on unused gas limit plus overestimation burn). A baseFee dip does not save the SP money - it increases the miner tip.

3. **Zero headroom, no recovery.** The cap is priced at `baseFee + gasTipCap` at send time, so any baseFee rise strands the message, and there is no fee-bumping path (#1292). Nonces are sequential, so one stranded message blocks the SP wallet's entire lane - including PDP proving messages (#1313 is the same shared-lane design).

**What this did to us in practice** (profiled 2026-07-10, both SPs, 30s-resolution mempool and baseFee data, per-tipset gas attribution):

- Our two nodes alone were **85-86% of all tipset gas** over an 80-tipset window - a bulk workload self-generates the baseFee ramps that strand its own messages. Stopping both nodes collapsed baseFee 960k to 505k in ~7 minutes; restarting ramped it straight back.
- Stock behaviour: one lane pinned for **2 1/4 hours** behind a single stale-cap head (918,479 cap vs baseFee oscillating 825k-1.14M) with nineteen includable messages (caps to 2.19M) nonce-blocked behind it. Client timeouts cycle fresh sends in behind the stuck head - we accumulated **308 stranded messages** overnight.
- **Proving hostage, twice in one day.** Proving messages share the wallet lane. A stranded migration message left data set 1342 39 minutes from missing its proving window (recovered only by manually halting the other SP to collapse baseFee). Later, with mitigations running but deep client queues, a second 30+ minute pin formed minutes *after* data sets 1244, 1255 and 99 got their proofs out - had the pin formed ten minutes earlier, all three would have missed their windows. An unattended SP in this state has a fault exposure, not a throughput problem.

Expected behaviour: 1559 semantics preserved end to end - the cap is a ceiling, the tip is the price signal, a message survives ordinary baseFee movement, and inclusion at a lower baseFee costs less, not more.

**Mitigations we tested (running on both SPs, happy to PR):**

1. Preserve the `DynamicFeeTx` through nonce assignment (type-conditional - legacy constructions elsewhere untouched).
2. `gasFeeCap = 2 x baseFee + gasTipCap`.
3. Floor the baseFee input at the max of `eth_feeHistory` over the last 120 epochs, so a restart into a quiet chain does not price messages against a transient floor (this exact trap stranded our first patched run within 15 minutes - caps born at a 24k floor, dead by 300k).

Measured results: messages now pay baseFee + tip instead of full cap; a restart from baseFee **183** (the harshest bootstrap case - first caps ~2.39M off the history floor) ran ~2 hours with both lanes confirming continuously through a 2,000x self-ramp, where the unfloored run pinned within 15 minutes.

The limit we then hit, and the reason pricing alone is not the fix: under sustained dual full-rate load (16 in-flight messages per node), messages queue long enough that even 2x caps go stale inside the self-generated ramp - one node pinned 30+ minutes at a 435k head against an 855k baseFee while the other node's fresh caps kept clearing. We are re-running at halved client concurrency so injection matches the chain's clearing rate, which should remove the self-induced cap staleness - but the correctness fix is #1292: same-nonce replacement for underwater messages, with confirmation tracked by nonce so either hash resolves. Our data above is effectively the severity case for it.

### Logging Information

```
eth_getTransactionByHash (any AddPieces tx, both SPs):  type: 0x0, gasPrice only, no maxFeePerGas
receipts (6 consecutive confirmed AddPieces):           cap 953932 -> paid 990720 | cap 953006 -> paid 970450
                                                        cap 959064 -> paid 1019407 | cap 965199 -> paid 1023162
                                                        cap 968439 -> paid 1026390 | cap 973185 -> paid 1031710
stalled lane snapshot (nonce-ordered pending):          head cap 918479 vs baseFee 825k-1.14M; caps interleaved
                                                        852126-2186904 behind it; all GasPremium == GasFeeCap
```

Full 30s-resolution TSVs from both nodes plus per-tipset gas attribution available on request.

### Repo Steps

1. Run any PDP node with addPieces load through `SenderETH`.
2. `eth_getTransactionByHash` any resulting tx: `type: 0x0`, no `maxFeePerGas`.
3. Compare `effectiveGasPrice` in the receipt to the tx gasPrice: paid >= cap, in all baseFee conditions.
4. Sustain enough load to move baseFee upward past send-time caps: the lane strands, including proving messages, with no recovery until baseFee dips below the head's cap.

LexLuthr:
## PR Summary

This PR adds a Curio replacement watcher that runs on chain head updates and replaces Curio-originated sends that appear stuck because their current fee values are no longer sufficient.

The replacer is split into separate Filecoin message and Eth transaction lanes.

### Filecoin Messages

For Filecoin messages, the replacer:

- Looks at successful Curio Filecoin message sends from locally configured sender addresses.
- Uses the sender actor nonce to only consider pending nonce lanes.
- Checks the latest signed message for each nonce, including prior successful replacements.
- Uses Lotus gas estimation with a fixed replacement max fee of 1 FIL.
- Applies Filecoin RBF premium requirements using a 120% premium bump.
- Limits work to 10 replacements per sender per trigger.
- Records replacement attempts in `message_send_replacements`, keyed by original signed CID, replaced signed CID, and replacement signed CID.
- Stores signed replacement bytes before pushing, so incomplete sends can be retried.
- Reuses already-signed replacement claims on retry.
- Treats Lotus existing-nonce / already-in-mpool push errors as successful broadcasts.
- Deletes invalid, no-longer-fee-stuck, or stale replacement claims.
- Cleans replacement rows once the actor nonce has advanced past them.

### Eth Transactions

For Eth transactions, the replacer:

- Uses a separate `transactionReplacer` and `message_send_eth_replacements` table.
- Keeps Eth replacement in the Eth domain, tracking original, replaced, and replacement transactions by signed transaction hash.
- Does not rely on Filecoin message CID lookup for Eth transaction replacement.
- Looks at successful Curio Eth sends from local `eth_keys`.
- Uses the sender account nonce at the current chain height to only consider pending nonce lanes.
- Replaces only dynamic-fee transactions, preserving nonce, recipient, value, calldata, gas limit, chain ID, and access list.
- Recomputes fee cap from current base fee and suggested tip, then enforces a 120% bump over the latest transaction being replaced.
- Signs replacements locally from `eth_keys`.
- Stores signed replacement transaction bytes before broadcasting, so incomplete sends can be retried.
- Treats already-known / already-imported / txpool-known broadcast errors as successful replacement broadcasts.
- Leaves signed claims pending when Eth RPC send times out, because broadcast state is unknown.
- Cleans replacement rows once the account nonce has advanced past them.

### Shared Wiring

The shared `NewMessageReplacer` wiring can run either lane independently:

- Normal Curio starts the Filecoin lane and adds the Eth lane only when Eth sender support is enabled.
- PDP node starts the Eth lane.
- Both lanes are triggered from the same chain-head watcher, but keep separate replacement state and SQL tables.

### Eth Waits

Eth callers continue to wait through `message_waits_eth` using the original tx hash.

The Eth watcher resolves replacements internally:

- `message_waits_eth.signed_tx_hash` remains the caller-facing original hash.
- `message_send_eth_replacements` is used to find the latest successful replacement hash.
- The watcher checks receipt and transaction data using the replacement hash when one exists.
- The watcher updates the original `message_waits_eth` row.
- `confirmed_tx_hash` records the actual landed hash, which may be the replacement hash.
- PDP/pay/domain watchers do not need to know about replacement rows.

### Tests

Tests cover both replacement paths, including successful replacement, signed-claim retry, young sends ignored, non-stuck claims deleted, invalid signed claims deleted, stale row cleanup, Filecoin existing-nonce success handling, Eth already-known success handling, and Eth send timeout retry behavior.

### Open Follow-Up
- Test SQL queries at scale before merging

- Eth original send handling still treats too many `SendTransaction` errors as definitive failures. It needs send-result classification so unknown broadcast state leaves the signed tx retryable instead of recording a false failure.


tasks/message/replacer.go
```diff
@@ -0,0 +1,135 @@
...
+		return nil
+	}
+
+	t.triggerMu.Lock()
+	t.latestTrigger = replaceTrigger{
+		Height:    apply.Height(),
+		Timestamp: time.Unix(int64(apply.MinTimestamp()), 0).UTC(),
+		Key:       new(apply.Key()),
+	}
+	t.triggerMu.Unlock()
```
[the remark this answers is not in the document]
LexLuthr: Incorrect. Learn Go 1.26

deps/config/utils.go
```diff
@@ -119,3 +126,127 @@ func loadConfigs(ctx context.Context, db *harmonydb.DB) (map[string]string, erro
...
+		HostAndPort string
+		Layers      string
+	}
+	var machineDetails []machineDetail
+
+	// Get all layers in use
+	err := db.Select(ctx, &machineDetails, `
+				SELECT m.id, m.host_and_port, d.layers
+				FROM harmony_machines m
+				LEFT JOIN harmony_machine_details d ON m.id = d.machine_id;`)
```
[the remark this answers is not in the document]
LexLuthr: Nope. Curio cannot start with NULL layers

tasks/message/sender_eth.go
ZenGround0, line 241: I don't think as it stands this PR is going to resolve the no subpiece found error.  When we get a false negative, i.e. a non nil send_error, and record it here then all of the callsites of sender_eth simply don't propagate to the watcher.  It doesn't matter how clever replacer or watcher code is I don't think we're going to undo that false negative.

ZenGround0 (commented):
I have not read through all of this yet.  From my first scan I like the way it looks.  However I am skeptical it addresses #1180 at all.  That might be fine.  We can address the basefee replacement issues here and then do something else to work on #1180.  If you think I'm missing something please point me in the right direction.  Either way I'll get a more thorough review of this over the weekend.

tasks/message/sender_eth.go
LexLuthr: No it is not. A PR on top of this would fix that. I did not write out that part intentionally and put it in open items in summary. Retry would be the base PR and then error handling would be on top of it as it would be common between normal sender and retry sender.

ZenGround0: That makes sense.  But I don't yet see a reason to block #1180 on this PR.  I'll review this one but it looks like there is potentially some blocking trickiness when relating this replace mechanism to synapse sdk (https://filecoinproject.slack.com/archives/C08TVNKJV7C/p1784170169227429?thread_ts=1784128896.158029&cid=C08TVNKJV7C) that I don't want us to block fixing the highest priority item on our list.

ZenGround0: If this is correct I think you should move forward with a fix to #1180 before we try to land this.

ZenGround0: Basically I don't think retry helps us resolve the race in #1180 at all.  The two issues seem orthogonal.

LexLuthr: They are connected because retry mechanism has to be wired in both senders. It seems like waste of time to do change on one side and then do a separate PR to do other sender. https://github.com/filecoin-project/curio/pull/1369 should handle #1180.

deps/config/utils.go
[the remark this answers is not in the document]
ZenGround0: Ha my AI was very worried about this when I started asking about this file too

deps/config/utils.go
```diff
@@ -119,3 +126,127 @@ func loadConfigs(ctx context.Context, db *harmonydb.DB) (map[string]string, erro
...
+
+type minerLookUpApi interface {
+	StateMinerInfo(ctx context.Context, actor address.Address, tsk types.TipSetKey) (api.MinerInfo, error)
+}
+
+func GetAddressesFromConfig(ctx context.Context, db *harmonydb.DB, api minerLookUpApi) ([]address.Address, []address.Address, error) {
+	// MachineDetails represents the structure of data received from the SQL query.
+	type machineDetail struct {
+		ID          int
+		HostAndPort string
```
ZenGround0: nit: Any reason to keep selecting ID and HostAndPort here?  It doesn't look like we actually read them so we could just drop these fields from this struct.

tasks/message/replace.go
```diff
@@ -0,0 +1,638 @@
+package message
```
ZenGround0, line 1: It's hard to miss how similar this is to tasks/message/replace_eth.go.  What are the main differences between these two files that makes sharing functions untenable?

tasks/message/replace_eth.go
```diff
@@ -0,0 +1,648 @@
...
+
+func isEthTransactionAlreadyKnownError(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	msg := strings.ToLower(err.Error())
+	return strings.Contains(msg, "already known") ||
+		strings.Contains(msg, "already imported") ||
+		strings.Contains(msg, "already in the txpool")
```
ZenGround0: I don't think these error strings are valid for filecoin mpool pushes.

ZenGround0: They appear to be valid for geth implementation.  While we match API we don't match error strings.

tasks/message/replace_eth.go
```diff
@@ -0,0 +1,648 @@
...
+	var rows []transactionReplaceCandidate
+	claimID := uuid.NewString()
+	// Load old successful Eth sends for active senders whose nonce has not
+	// landed yet, resolve the current successful replacement chain tip for each
+	// send, and claim the tip for this run. Expired unsent claims are reclaimed;
+	// active claims and already-successful replacement rows are skipped. Returned
+	// rows are only the claims owned by this run, including any previously signed
+	// replacement transaction that should be retried instead of re-signed.
+	err = t.db.Select(ctx, &rows, `
+		WITH sender_nonces AS (
```
ZenGround0, line 205: I will note that this 100+ line SQL query is quite inscrutable to me.  I'd like some help from @snadrus confirming that it works as intended.  Is there any path to splitting this up and encoding some of the logic within go?  If not I want to understand the constraints i.e. atomicity/safety or expected efficiency.

tasks/message/replacer.go
```diff
@@ -0,0 +1,135 @@
...
+
+	"github.com/filecoin-project/curio/build"
+	"github.com/filecoin-project/curio/harmony/harmonydb"
+	"github.com/filecoin-project/curio/lib/chainsched"
+
+	"github.com/filecoin-project/lotus/chain/types"
+)
+
+const (
+	ReplaceStuckEpochs abi.ChainEpoch = 3
```
ZenGround0: Why 3? It seems a little short to me to expect all messages to land on chain in 3 epochs. This seems like the sort of thing that would probably be good to be configurable?

tasks/message/replace_eth.go
```diff
@@ -0,0 +1,648 @@
...
+
+	DeleteSignedClaim bool
+}
+
+type signedTransactionReplacement struct {
+	Candidate transactionReplaceCandidate
+	SignedTx  *gethtypes.Transaction
+}
+
+func (t *transactionReplacer) runTransactionReplacement(ctx context.Context, height abi.ChainEpoch, tipSetTimeStamp time.Time) error {
```
ZenGround0: nit: I would probably call this ethMessageReplacer / ethMessageReplacement.  Or perhaps ethTransactionReplacer etc.  Using Transaction to refer to eth messages is not particularly standard elsewhere in filecoin code so referencing `eth` will help make this clearer.

tasks/message/eth_tx_manager.go
```diff
@@ -34,23 +39,37 @@ func (h *HarmonyEthTxManager) AssignPendingToMachine(ctx context.Context, machin
...
+		FROM message_waits_eth mwe
+		LEFT JOIN LATERAL (
+			SELECT r.signed_hash
+			FROM message_send_eth_replacements r
+			WHERE r.original_signed_hash = mwe.signed_tx_hash
+				AND r.send_success = TRUE
+				AND r.signed_hash IS NOT NULL
+				AND NOT EXISTS (
+					SELECT 1
+					FROM message_send_eth_replacements next
```
ZenGround0, line 62: I am not sure if this clever way of finding the last replacement by finding a row in the replacements table that does not have a matching next with a replacement hash is the best thing from either clarity or effiiency standpoint.  Did you think about any other formats for the data model that can make all of this more explicit?  For example maintaining a singleton table with the latest head per sender?

ZenGround0 (commented):
One thing I want to work through with you is the case where the original message / a replacement ends up getting submitted to the chain after a later replacement has been recorded in curio tables and sent to mpool.  I want to make sure this is handled properly without locking up the sender nonce.  I suspect it is handled in here but I haven't been able to trace this out yet.

ZenGround0:
One thing I learned after talking to @rvagg is that this will require some synapse work before we release this.  But that we should be empowered to drive that forward.  It sounds like you guys brainstormed something about curio status endpoints returning the replaced tx hash to listen on.  @LexLuthr how do you feel about having a go at making this synapse change as a follow on to this PR?

tasks/message/replace_eth.go
```diff
@@ -0,0 +1,648 @@
...
+	feeStuck := latest.GasFeeCap().Cmp(gasFeeCap) < 0 ||
+		latest.GasTipCap().Cmp(gasTipCap) < 0
+	if !feeStuck {
+		return nil, false, nil
+	}
+
+	rbfTipCap := bumpInt(latest.GasTipCap(), TransactionReplaceGasBumpPercent)
+	rbfFeeCap := bumpInt(latest.GasFeeCap(), TransactionReplaceGasBumpPercent)
+	gasTipCap = maxInt(gasTipCap, rbfTipCap)
+	gasFeeCap = maxInt(gasFeeCap, rbfFeeCap)
```
beck-8: We need to add a cap here too, right? Just like FIL?

beck-8 (commented):
I would like to ask if a cap should be added to the ETH channel? 
I happened to see a PDP SP recently using a custom script without setting a strict upper limit, which cost him a lot of extra money.

tasks/message/replace_eth.go
LexLuthr: They were removed in https://github.com/filecoin-project/curio/pull/1369

tasks/message/replacer.go
LexLuthr: Increased to 10. I don't think this should be configurable. I want new SPs to have minimum knobs to turn. If users complain that this is too aggressive, we can find a more suitable value. I think we should stick to sensible default unless there really is a diverse need.

tasks/message/replace_eth.go
LexLuthr: SQL perf was tested with 1M rows per relevant table, 64 senders, 16 active senders, and 10 candidates per sender.

Before indexes, the Filecoin replacement candidate query took ~18.1s on Postgres and the Eth candidate query took ~22.4s. The Eth wait lookup by original tx hash did not complete within 60s.

After the indexes:
- Filecoin candidate claim: Postgres 6.6ms, Yugabyte 26.4ms
- Eth candidate claim: Postgres 13.8ms, Yugabyte 26.1ms
- Eth wait lookup for 10k pending waits: Postgres 616.8ms, Yugabyte 52.1ms

Cleanup was also changed from one broad unnest+OR delete to split per-sender deletes matching partial indexes. In steady state/no stale rows:
- Filecoin cleanup: Postgres ~0.18ms per sender, Yugabyte ~1.06ms per sender
- Eth cleanup: Postgres ~0.24ms per sender, Yugabyte ~1.22ms per sender

For a heavier cleanup case deleting 12k stale rows for one sender:
- Filecoin cleanup: Postgres 154.6ms, Yugabyte 659.7ms
- Eth cleanup: Postgres 213.5ms, Yugabyte 692.7ms

tasks/message/eth_tx_manager.go
LexLuthr: I considered keeping an explicit latest pointer, but I kept this as a derived lookup because the replacement rows are already the authoritative history. A separate head table would only be a denormalized read model, so every successful replacement would need to update both the history row and the head row correctly.

With the added indexes, this lookup is no longer doing a broad scan. I tested it with 1M replacement rows and 10k pending waits:
- Postgres: ~617ms
- Yugabyte: ~52ms

Given this runs in the watcher path and is bounded to 10k waits, I think the current shape is acceptable for now.

tasks/message/replace.go
LexLuthr: The similarity is intentional: both files implement the same replacement lifecycle. I kept them separate because the shared part is mostly the control-flow shape, while the correctness rules are chain-specific.

Filecoin replacement is keyed by signed CID, uses actor nonce, CBOR signed messages, Lotus gas estimation, Filecoin RBF premium rules, WalletSignMessage, and MpoolPush semantics.

Eth replacement is keyed by tx hash, uses account nonce, geth transaction decoding, EIP-1559 base fee/tip cap rules, local eth_keys signing, SendTransaction, and Eth txpool error semantics.

We split these lanes specifically to avoid carrying Filecoin assumptions into Eth replacement. A shared abstraction here would mostly be a callback/generic runner or SQL templating over table/column names, which would reduce line count but make the domain-specific parts harder to audit.

ZenGround0:
> I would like to ask if a cap should be added to the ETH channel?

This sounds like a great idea.  Could you open an issue describing this item of work and then open a separate follow on PR?

LexLuthr:
> > I would like to ask if a cap should be added to the ETH channel?
> 
> This sounds like a great idea. Could you open an issue describing this item of work and then open a separate follow on PR?

This is same as what we discussed maxFee == 0.05 Fil on both Filecoin and Eth messages. I have wired it up.

tasks/message/replacer.go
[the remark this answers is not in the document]
ZenGround0: You're being very mean to the robots @LexLuthr

tasks/message/replace_eth.go
ZenGround0: Ok for this one because I really want that one to land right away.  In future I want to push for every PR independently being mergable only if it does not introduce new bugs.

tasks/message/eth_tx_manager.go
ZenGround0: Thanks for validating the approach

tasks/message/replacer.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 const (
-	ReplaceStuckEpochs abi.ChainEpoch = 3
+	ReplaceStuckEpochs abi.ChainEpoch = 10
```
ZenGround0, line 18: This seems like the sort of thing we are going to want to make configurable.

ZenGround0 (approved):

tasks/message/replacer.go
LexLuthr: When we enable the replacement system, we can decide what should be config and what can stay internal.

PR #1309, opened 2026-06-12
pdp: accept a pieceIds array in the DeletePiece endpoint for batched removals

TippyFlitsUK:
`schedulePieceDeletions` on PDPVerifier takes a `uint256[]`, but the DELETE `/pdp/data-sets/{set}/pieces/{pieceId}` handler hardcodes a single-element array — so bulk removals cost one transaction per piece.

This change adds an optional `pieceIds` field to the request body. When present, the handler validates all ids exist in the data set, packs them into one `schedulePieceDeletions` call, and records `rm_message_hash` for every id. When absent, behaviour is unchanged (the path id is used, as today).

Context: while removing ~19,800 duplicate pieces from one data set (a client-tool re-commit bug, details in BravoNatalie/migration-support-scripts#9 and fixed in BravoNatalie/migration-support-scripts#12), per-piece removal would have been 19,800 transactions; with this change it is 40 (500 ids per call, each estimating ~2.5B gas — well within block limits). The client-side authorization already supports arrays (`signSchedulePieceRemovals` signs the id list), so only the handler needed the change.

Running in production on my mainnet node.

### DIFF
--- a/pdp/handlers.go
+++ b/pdp/handlers.go
@@ -1079,7 +1079,7 @@ func (p *PDPService) handleDeleteDataSetPiece(w http.ResponseWriter, r *http.Req
 		return
 	}
 	if foundCount != len(pieceIDsI64) {
-		http.Error(w, "Piece not found", http.StatusNotFound)
+		http.Error(w, "One or more piece not found", http.StatusNotFound)
 		return
 	}
 
--- a/pdp/handlers.go
+++ b/pdp/handlers.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"math/big"
 	"net/http"
 	"strconv"
@@ -56,6 +57,8 @@ const (
 
 	// MaxDeletePieceExtraDataSize defines the limit for extraData size in DeletePiece calls (256B).
 	MaxDeletePieceExtraDataSize = 256
+
+	MaxDeletePiecesBatchSize = 500
 )
 
 // PDPService represents the service for managing data sets and pieces
@@ -953,6 +956,25 @@ func (p *PDPService) handleGetPieceAdditionStatus(w http.ResponseWriter, r *http
 	}
 }
 
+func normalizeDeletePieceIDs(ids []uint64) ([]int64, error) {
+	if len(ids) > MaxDeletePiecesBatchSize {
+		return nil, fmt.Errorf("piece count (%d) exceeds the maximum allowed per DeletePiece call (%d)", len(ids), MaxDeletePiecesBatchSize)
+	}
+	seen := make(map[uint64]struct{}, len(ids))
+	out := make([]int64, 0, len(ids))
+	for _, id := range ids {
+		if id > math.MaxInt64 {
+			return nil, fmt.Errorf("piece ID %d is out of range", id)
+		}
+		if _, ok := seen[id]; ok {
+			continue
+		}
+		seen[id] = struct{}{}
+		out = append(out, int64(id))
+	}
+	return out, nil
+}
+
 func (p *PDPService) handleDeleteDataSetPiece(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 	// Step 1: Verify that the request is authorized using ECDSA JWT
@@ -1044,13 +1066,19 @@ func (p *PDPService) handleDeleteDataSetPiece(w http.ResponseWriter, r *http.Req
 		pieceIDs = payload.PieceIDs
 	}
 
+	pieceIDsI64, err := normalizeDeletePieceIDs(pieceIDs)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
 	var foundCount int
-	err = p.db.QueryRow(ctx, `SELECT COUNT(DISTINCT piece_id) FROM pdp_data_set_pieces WHERE data_set = $1 AND piece_id = ANY($2)`, dataSetId, pieceIDs).Scan(&foundCount)
+	err = p.db.QueryRow(ctx, `SELECT COUNT(DISTINCT piece_id) FROM pdp_data_set_pieces WHERE data_set = $1 AND piece_id = ANY($2::bigint[])`, dataSetId, pieceIDsI64).Scan(&foundCount)
 	if err != nil {
 		httpServerError(w, http.StatusInternalServerError, "Failed to query piece existence", err)
 		return
 	}
-	if foundCount != len(pieceIDs) {
+	if foundCount != len(pieceIDsI64) {
 		http.Error(w, "Piece not found", http.StatusNotFound)
 		return
 	}
@@ -1062,9 +1090,9 @@ func (p *PDPService) handleDeleteDataSetPiece(w http.ResponseWriter, r *http.Req
 		return
 	}
 
-	pieceIDArgs := make([]*big.Int, len(pieceIDs))
-	for i, id := range pieceIDs {
-		pieceIDArgs[i] = new(big.Int).SetUint64(id)
+	pieceIDArgs := make([]*big.Int, len(pieceIDsI64))
+	for i, id := range pieceIDsI64 {
+		pieceIDArgs[i] = big.NewInt(id)
 	}
 
 	// Pack the method call data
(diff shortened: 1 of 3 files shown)


pdp/handlers.go
```diff
@@ -1038,14 +1039,18 @@ func (p *PDPService) handleDeleteDataSetPiece(w http.ResponseWriter, r *http.Req
...
-	// Check if we have this piece or not
-	var found bool
-	err = p.db.QueryRow(ctx, `SELECT EXISTS(SELECT 1 FROM pdp_data_set_pieces WHERE data_set = $1 AND piece_id = $2)`, dataSetId, pieceID).Scan(&found)
+	pieceIDs := []uint64{pieceID}
+	if len(payload.PieceIDs) > 0 {
+		pieceIDs = payload.PieceIDs
+	}
+
+	var foundCount int
+	err = p.db.QueryRow(ctx, `SELECT COUNT(DISTINCT piece_id) FROM pdp_data_set_pieces WHERE data_set = $1 AND piece_id = ANY($2)`, dataSetId, pieceIDs).Scan(&foundCount)
```
LexLuthr: This is probably expensive. Can you check query plan once?

LexLuthr (commented):
We are missing a few things here.
1. We have to put limit on number of pieces per tx to avoid out of gas.
2. I don't see any impact analysis or changes to watcher side.

TippyFlitsUK:
Pushed d1d4a51 addressing the review feedback.

**@LexLuthr — is the existence check expensive? (query plan)**
Checked on my mainnet node against the largest data set (`data_set` 77, 1,003,823 pieces) with a full 500-id array:
```
Index Scan using pdp_data_set_pieces_piece_id_unique  (actual time=2.944..3.347 rows=500)
  Index Cond: (data_set = 77) AND (piece_id = ANY (...))
  Storage Table Read Requests: 1
  Storage Table Rows Scanned: 500
Execution Time: 4.76 ms
```
It rides the primary key (`data_set HASH, piece_id ASC, sub_piece_offset ASC`): `data_set` hash-locates the partition and `piece_id ASC` turns the `= ANY` into a single batched in-list probe — one storage round-trip, scanning exactly the 500 requested rows, ~5 ms. Cheap even on a million-row data set. (`COUNT(DISTINCT piece_id)` is needed because each `piece_id` spans multiple `sub_piece_offset` rows.)

**Unbounded `pieceIds` → oversized txs (cap)**
Added `MaxDeletePiecesBatchSize = 500`; oversized batches now return 400, mirroring `MaxAddPiecesBatchSize`. The value is measured rather than guessed: across 5,240 `schedulePieceDeletions` txs on my mainnet node, the largest 500-piece tx used 2.02B gas — 20.2% of the 10B block limit — with zero failures. Marginal cost is ~3.8M gas/piece on a ~125M base, so the block limit isn't threatened until ~2,000+ pieces; 500 keeps ~5x headroom.

**`[]uint64` → `bigint[]` encoding (existence check + UPDATE)**
Request ids are now converted to `[]int64` and both queries cast `ANY($n::bigint[])`, so pgx encodes them correctly against the BIGINT `piece_id` column.

**Duplicate IDs → spurious 404**
Ids are de-duplicated (and range-checked against `MaxInt64`) in a `normalizeDeletePieceIDs` helper; the existence check compares `COUNT(DISTINCT piece_id)` against the unique request count, so duplicate ids no longer 404.

**No tests**
Added `pdp/handlers_delete_test.go`:
- `TestNormalizeDeletePieceIDs` (unit): single-id passthrough, duplicates collapse, batch at the cap, batch over the cap → error, `MaxInt64` accepted, out-of-int64-range → error.
- `TestHandleDeleteDataSetPiece_MissingPiece404` (handler): a batch containing a missing id returns 404.

The success path (which reaches `SenderETH.Send`) isn't unit-tested because the sender is a concrete type with no interface to mock — happy to abstract it if you'd like that path covered too.

rvagg:
There's a limit of [2,000](https://github.com/FilOzone/pdp/blob/main/src/PDPVerifier.sol#L48) piece removals in a single proving period (1 day on mainnet), so any single call to `schedulePieceRemovals` is going to revert if it adds enough new pieces to break that limit. https://github.com/FilOzone/pdp/blob/main/src/PDPVerifier.sol#L818-L821

If you want to be decoding and surfacing that revert msg back to users (or synapse could decode it I guess) then that might be OK so the user knows why their 6th batch of 400 removals fails. Otherwise Curio could be a bit smart and call `getScheduledRemovals()` on PDPVerifier to figure out what the limit is and send something sensible back to the user.

pdp/handlers.go
```diff
@@ -1038,14 +1061,24 @@ func (p *PDPService) handleDeleteDataSetPiece(w http.ResponseWriter, r *http.Req
...
+
+	var foundCount int
+	err = p.db.QueryRow(ctx, `SELECT COUNT(DISTINCT piece_id) FROM pdp_data_set_pieces WHERE data_set = $1 AND piece_id = ANY($2::bigint[])`, dataSetId, pieceIDsI64).Scan(&foundCount)
 	if err != nil {
 		httpServerError(w, http.StatusInternalServerError, "Failed to query piece existence", err)
 		return
 	}
-	if !found {
+	if foundCount != len(pieceIDsI64) {
 		http.Error(w, "Piece not found", http.StatusNotFound)
```
LexLuthr:
```suggestion
		http.Error(w, "One or more piece not found", http.StatusNotFound)
```

pdp/handlers.go
```diff
@@ -56,6 +57,8 @@ const (
 
 	// MaxDeletePieceExtraDataSize defines the limit for extraData size in DeletePiece calls (256B).
 	MaxDeletePieceExtraDataSize = 256
+
+	MaxDeletePiecesBatchSize = 500
```
LexLuthr, line 61: This will also impact the extra_data size. Have we taken that into account?

pdp/handlers.go
```diff
@@ -953,6 +956,25 @@ func (p *PDPService) handleGetPieceAdditionStatus(w http.ResponseWriter, r *http
...
+func normalizeDeletePieceIDs(ids []uint64) ([]int64, error) {
+	if len(ids) > MaxDeletePiecesBatchSize {
+		return nil, fmt.Errorf("piece count (%d) exceeds the maximum allowed per DeletePiece call (%d)", len(ids), MaxDeletePiecesBatchSize)
+	}
+	seen := make(map[uint64]struct{}, len(ids))
+	out := make([]int64, 0, len(ids))
+	for _, id := range ids {
+		if id > math.MaxInt64 {
+			return nil, fmt.Errorf("piece ID %d is out of range", id)
+		}
```
LexLuthr, line 968: This seems unnecessary give that we are doing json unmarshal to get the value.

LexLuthr (commented):
Need some minor clean but otherwise it is good. Can you please double check that extra_data will allow 500 pieces?

TippyFlitsUK:
Thanks Mayank. Addressed the error-message wording you suggested, and answered the `extra_data` and range-check questions inline. Let me know if there's any other cleanup you'd like.

pdp/handlers.go
TippyFlitsUK: This check is guarding the `uint64` to `int64` conversion rather than the JSON parse. `json.Decode` into `[]uint64` will accept values above `math.MaxInt64` (up to `math.MaxUint64`), and those overflow to a negative number when cast to the `int64` that the `BIGINT` `piece_id` column needs. The range-check rejects them before they reach the DB, so it's doing work the unmarshal doesn't.

Happy to add a brief comment making that intent clear if useful.

pdp/handlers.go
TippyFlitsUK: Checked. The `extraData` field is independent of the piece count, it's a separate argument to `schedulePieceDeletions`, capped at 256 bytes (`MaxDeletePieceExtraDataSize`) and enforced regardless of how many pieces are in the call, so batching to 500 doesn't grow it.

What scales with 500 pieces is the `pieceIds` array in the calldata itself, which is bounded by `MaxDeletePiecesBatchSize` (500) and sized to sit within block gas limits (roughly 2.5B gas per full batch, per the PR description). So 500 is fine on both counts.

pdp/handlers.go
```diff
@@ -1008,7 +1030,8 @@ func (p *PDPService) handleDeleteDataSetPiece(w http.ResponseWriter, r *http.Req
 		return
 	}
 	type DeletePiecePayload struct {
-		ExtraData *string `json:"extraData"`
+		ExtraData *string  `json:"extraData"`
```
ZenGround0, line 1033: I think we need to do some API documentation updating for this.  I'll push a commit

ZenGround0: All set

ZenGround0:
FYI @hugomrdias this is a potential synapse change to incorporate.  It shouldn't be a breaking change because (TIL) go is smart enough with its json parsing to fill unused struct fields with 0 values.

Filed a tracking issue here: https://github.com/FilOzone/synapse-sdk/issues/863
Not required of course but potentially useful

LexLuthr (approved):

Issue #77, opened 2024-06-28
Drop WindowPost tasks if deadline is past


LexLuthr:
Right now we expect wdPost tasks related to deadline in past to not be executed. This was not seen practically in the case of a recent migration. We need to investigate the cause and design wdPost so we can drop the running task if deadline is already closed. Otherwise, 2 or more parallel running tasks cause extreme IO load on permanent storage. They can also lead to failure of current deadline wdPost due to slow read.

LexLuthr:
Fixed in https://github.com/filecoin-project/curio/pull/462

PR #1311, opened 2026-06-16
Only PDP

### LINKED ISSUE #1327: Only PDP
A PDP-oriented Curio should:

- Not need CGO, ffi
- Have considerably simpler tools & setup.
- Offer a one-touch install
- be easy to upgrade to "full Curio" if ever necessary.
- pass all synapse tests

snadrus:
`make curio-pdp` aka `make skiff`

A PDP-oriented Curio should:
- Not need CGO, Lotus (use Lantern)
- Have considerably simpler tools & setup. 
- Offer a one-touch install
- Be easy to upgrade to "full Curio" if ever necessary. 
- pass all synapse tests

Setup Differences: 
- no need for miner ID
- name folders "curio-pdp-data" before start-up (within the first 3 levels of any mountpoint) to store data.
- wallet creation via UI
- DB administration via UI

Rough edges:
- A few deps are bigger than needed, but it's overall smaller than Curio
- Web UI needs a LOT of clean-up: Next PR. 

Progress so far: 
After one 1 command and 5 minutes (from a new machine): make skiff/calibnet/up
image
(considerable adjustment is still needed on the UI)

image


Reiers:
Good to see Lantern getting used here.

For context on the chain backend this PR depends on: **Lantern** (`github.com/Reiers/lantern`, in your `go.mod` at v1.7.16) and **[Reiers/curio-core](https://github.com/Reiers/curio-core)** — a PDP-only, FFI-free, Lotus-free, single-binary Curio with Lantern embedded — have been public and running on calibration + mainnet (`f03678816`) since February. curio-core is the production runner for exactly this approach; this PR is a parallel upstream take on the same idea.

Pin notes so you don't hit walls we already cleared:
- **Hold Lantern at v1.7.16+.** First version with read+write both proven bridge-off on calibration mainnet: local FEVM `eth_call` ([lantern#43](https://github.com/Reiers/lantern/issues/43)), full local write path ([lantern#45](https://github.com/Reiers/lantern/issues/45)), Filecoin bitswap prefix fix ([lantern#50](https://github.com/Reiers/lantern/issues/50)). <= v1.7.15 still calls Glif on reads and falls back to `--vm-bridge-rpc` on writes. `pkg/daemon` is the embed surface (added v1.5.0 for this); happy to keep it semver-stable.
- Three landmines I've already mapped and fixed in curio-core, in case they help: [#77](https://github.com/Reiers/curio-core/issues/77) (`harmony_machines` keepalive -> task dispatch dies ~10m after boot), [#81](https://github.com/Reiers/curio-core/issues/81) (eth watcher registration race -> pdpv0 stalls), [#83](https://github.com/Reiers/curio-core/issues/83) (zero-Glif gap analysis - the path to unconditional `--vm-bridge-rpc-disable`).

Glad the architecture's converging upstream - happy to coordinate so the work isn't duplicated.

pdpnode/storage_discover.go
```diff
@@ -0,0 +1,206 @@
...
+
+	sort.Strings(pathsOut)
+	return pathsOut, nil
+}
+
+func defaultPDPStorageMeta() storiface.LocalStorageMeta {
+	return storiface.LocalStorageMeta{
+		ID:       storiface.ID(uuid.New().String()),
+		Weight:   defaultPDPStorageWeight,
+		CanStore: true,
```
LexLuthr, line 162: Allow can seal as well. otherwise, uploads won't work

LexLuthr (approved):

Reiers:
I've been building exactly this since February - a PDP-only, FFI-free Curio derivative (Curio Core) on a Lantern-backed chain API. As of today it's proven end-to-end on mainnet: SP registered (provider 31), dataset created, piece added, and a PDP proof landed and mined - tx 0x9566d349…, block 6136371, status 0x1. Full create→add→prove→settle cycle, on minimal hardware (mac mini), self-sustaining proving loop.

So a lot of the hard problems this PR is still working through are already solved and verified on-chain: the FFI-free/sealing-free build, the Lantern chain backend, and the upload path (I went diskstash instead of the cluster storage model, which sidesteps the CanSeal-for-uploads issue raised in review here entirely).

Happy to share what I learned so far - the proving-window scheduling drift in particular bites in a way that isn't obvious until you're live on mainnet. One note: the embedded Lantern here looks behind; current is v1.7.24 (adds the eth receipt fields you need for cost/cast receipt), and it should be straightforward to keep it bumped.

I'd genuinely rather we weren't building the same thing twice.

Screenshot 2026-06-25 at 16 37 38

ZenGround0:
I've brought it up a few times already but I have strong reservations about landing this with Lantern in the first place.  I know we disagree @snadrus and you want to go ahead like this.  If we start seeing reliability trouble with this then we're gong to need to back out of this as a first step.  Since you are confident in it ok we can try it.  Please register my concern that this is going to make debugging more muddled.

Reiers:
> I've brought it up a few times already but I have strong reservations about landing this with Lantern in the first place. I know we disagree @snadrus and you want to go ahead like this. If we start seeing reliability trouble with this then we're gong to need to back out of this as a first step. Since you are confident in it ok we can try it. Please register my concern that this is going to make debugging more muddled.

@ZenGround0 If the goal is making Filecoin storage easy - PDP on minimal hardware - then the chain backend has to fit that form factor too. Lotus and Forest are full-consensus nodes; neither fits a Mac-mini-class box, and requiring one removes the whole reason to do PDP-only. Glif keeps you remote, rate-limited, version-lagged, and unable to introspect when it stalls - which is its own debugging-muddle, just hosted. For this segment I genuinely don't see the alternative that keeps the "runs on anything" promise. So the reliability bar is the thing to raise, not the dependency to drop.

On that bar: I've been running a Lantern-backed PDP node and it lands a full mainnet proving cycles on a Mac mini. It's been debuggable in practice - on-chain reverts and task failures surface directly, the receipt path is complete. So the concern is fair, but there's at least a few real data point that it holds up under live proving. 

If there are specific reliability gaps people hit, I'd rather see them filed against Lantern so we can close them - that's a smaller, more tractable target than swapping the backend, and it helps every PDP-on-light-client effort, not just one.

ZenGround0:
Looks like we're checking in the `skiff` binary.  Is that unintentional? 

Lint is unhappy about storage config functions being unused.

ZenGround0:
@Reiers just because something doesn't land in one PR doesn't mean it is out of scope for a project. I'm not objecting to it in principle it's just a lot to be doing all at once in here.  The alternative is to do things in stages with testing in between.  This is not the way curio engineering has typically done things and perhaps for good reason.  So I'm trusting @snadrus, suspending my disbelief and not blocking on this.  But if we get hit in the face with things breaking mysteriously I'm going to demand we cut out the thing I have the least confidence in quickly rather than trying to debug the whole 10k line PR all at once.

The fact that you have been testing it out and seeing good results is encouraging.  I was not sure how much use the repo has seen yet.  

> If there are specific reliability gaps people hit, I'd rather see them filed against Lantern so we can close them 

I agree with this completely

docker/skiff/Dockerfile
```diff
@@ -0,0 +1,95 @@
...
+    XDG_CACHE_HOME="/tmp"
+
+WORKDIR /opt/curio
+
+# Pre-fetch modules. filecoin-ffi is a local `replace`; maxboom does not link the
+# native libraries. Sparse-clone default branch tip (Go module metadata only, not rust/, ~9GB).
+COPY go.mod go.sum .gitmodules ./
+
+RUN git clone --filter=blob:none --sparse --depth 1 \
+      https://github.com/filecoin-project/filecoin-ffi.git extern/filecoin-ffi \
```
ZenGround0, line 37: I thought we were getting rid of cgo in this deployment?  Why do we need ffi here?

ZenGround0: Seeing no-ffi claim [here](https://github.com/filecoin-project/curio/pull/1311/changes#diff-71fb95b1d781bb0936958cb984949e6b31a1c8020c09afc5708eddfa7b27c9d8R3) too

documentation/en/curio-pdp.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Architecture and data stores
+
+| Deployment | HarmonyDB (tasks, config, PDP state) | Piece index (multihash → offset) |
+|------------|--------------------------------------|----------------------------------|
+| **Full Curio** | Yugabyte (YSQL) | Yugabyte YCQL / Cassandra-compatible |
+| **Tests / CI** | Postgres | Scylla (CQL) |
+| **Curio-PDP (maxboom)** | **Postgres** | **Disk-based indexing** (target; see note below) |
+
+Curio-PDP is intentionally lighter: no Yugabyte requirement and no separate CQL/Scylla service for operators who only run PDP.
```
ZenGround0: Dockerfile entrypoint is still starting Yugabyte.  I think this text is incorrect.

ZenGround0: There is a future question on whether we should remove yugabyte from this but I don't think it is gone from the simple deployment yet

web/api/webrpc/sql.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"context"
+
+	"golang.org/x/xerrors"
+
+	"github.com/filecoin-project/curio/harmony/harmonydb"
+)
+
+// SQLQuery runs arbitrary SQL against the HarmonyDB for the admin console.
+func (a *WebRPC) SQLQuery(ctx context.Context, query string) (*harmonydb.AdminQueryResult, error) {
```
ZenGround0, line 12: Just confirming -- this is a totally new web rpc endpoint that doesn't exist on curio main currently?

documentation/en/curio-pdp.md
```diff
@@ -0,0 +1,124 @@
...
+
+Piece payload files live on disk under writable paths discovered under `/data` (see [Storage](#storage)). Index data on disk must be copied separately when moving hosts; it is not included in SQL dumps.
+
+{% hint style="info" %}
+**Index store note:** The target Curio-PDP stack uses disk-based piece indexing. Current builds may still open a CQL `IndexStore` via `--db-cassandra-port`; disk-only indexing is follow-up work. Plan deployments accordingly.
```
ZenGround0: Even if we end up moving away from yugabyte I thought we were planning on moving to raw Postgres rather than using disk based piece indexing?

pdpnode/config_maxboom.go
```diff
@@ -0,0 +1,163 @@
...
+
+func defaultMaxBoomBaseConfig() *config.CurioConfig {
+	cfg := config.DefaultCurioConfig()
+	cfg.Subsystems.EnablePDP = true
+	cfg.Subsystems.EnableWebGui = true
+	cfg.Subsystems.GuiAddress = "127.0.0.1:4701"
+	cfg.HTTP.Enable = false
+
+	if maxboomDockerMode() {
+		cfg.Subsystems.GuiAddress = "0.0.0.0:4701"
```
ZenGround0: I think we can reach the unrestricted SQL console through this address now right?  ocker/maxboom/docker-compose.yaml shows yugabyte setup with default credentials. Is the current setup safe?  Since this deployment is for maximally lazy SPs I want to make sure they don't need to do specific firewall setup to avoid exposing arbitrary reads and writes into their database.

.github/workflows/ci.yml
```diff
@@ -131,6 +131,49 @@ jobs:
...
+      - name: Cache synapse-sdk checkout
+        uses: actions/cache@v4
+        with:
+          path: .synapse-sdk
+          key: synapse-sdk-${{ hashFiles('scripts/ci/synapse-sdk.ref') }}
+
+      - name: Build curio-pdp without CGO
+        env:
+          CGO_ENABLED: '0'
+        run: make curio-pdp
```
ZenGround0, line 166: Especially now that we are not seriously considering supporting filone with this deployment and don't benefit as much from the metullurgy theme I don't care what name we use.  But I am seeing inconsistencies in the PR where sometimes we refer to curio-pdp and sometimes to maxboom.  We should probably be referring to only one of these throughout

pdpnode/http.go
```diff
@@ -0,0 +1,176 @@
...
+	chiRouter.Use(middleware.RealIP)
+	chiRouter.Use(middleware.Recoverer)
+	chiRouter.Use(handlers.ProxyHeaders)
+	chiRouter.Use(corsHeaders)
+
+	corsOrigin := "https://" + cfg.DomainName
+	if devURL := os.Getenv("DEV_CURIO_EXTERNAL_URL"); devURL != "" {
+		corsOrigin = devURL
+	}
+	chiRouter.Use(handlers.CORS(handlers.AllowedOrigins([]string{corsOrigin})))
```
ZenGround0: I think this maybe be a noop?  corsHeaders called on line 64 runs `w.Header().Set("Access-Control-Allow-Origin", "*")`

.github/workflows/ci.yml
snadrus: The logic is this: 
- A wide-ranging feature without a codename is difficult to work with.
- Jennifer wants a "nameless" Curio-PDP **build**
- "skiff" was nice because it appeared zero times in the code before this PR
- "maxboom" is less nice as we use 'max' a lot. 

I'm going to resolve this without code change, but with a slight will to name the code-side feature something that is less conflict-y than 'maxboom'.

docker/skiff/Dockerfile
snadrus: It's for 3 metadata files REPLACE & ffi-version-check insist on. comment added

documentation/en/curio-pdp.md
snadrus: we are keeping yugabyte in this iteration. The focus is on simplicity for users rather than sheer size reduction (it's pointless to save a few MB for a storage product).

documentation/en/curio-pdp.md
snadrus: It's true. I've removed all mentions of both to keep docs accurate to the code of YB-only.

web/api/webrpc/sql.go
snadrus: yes, open to all. Anyone with admin console access can already get this info, so simplifying DB access is fine.

snadrus: This was me "starting into" fixing-up the web interface for PDP. I realize it's going to be a lot more than just this, so I plan to put that in another PR, but reverting this seems unnecessary.

pdpnode/config_maxboom.go
snadrus: I updated docker-compose to limit the admin GUI to localhost only.
I also added info in curioo-pdp.md mentioing that's the admin GUI port and updated maxboom-binary.md to explicitly say to open 80 & 443 only.

pdpnode/http.go
snadrus: fixe. it's a remnant of merging pdpv0 branch over and having solved similar problems twice.

web/api/webrpc/sql.go
snadrus: well, it was already public in a well-designed SP. If they plumbed that interface out to people who didn't have logs & system level access, then this could be additional exposure.

.github/workflows/ci.yml
ZenGround0: Please lets just go with skiff everywhere then.  Having curio-pdp and skiff together is confusing.

ZenGround0: And I'd rather get rid of maxboom from the start in this PR to remove further sources of confusion.

documentation/en/curio-pdp.md
ZenGround0: Yes absolutely

ZenGround0 (approved):
After CI is passing LGTM for merge

ZenGround0 (changes requested):
After a lot of discussion I've decided lantern is not quite ready to be relied on as a dependency.  We should explore the lite client protocol / software space more before introducing lantern as a hard dependency.  Please revert `make skiff` to rely on an external lotus node and remove lantern dependency for this PR.  We can introduce lite client feature as another PR.

.github/workflows/ci.yml
snadrus: it's now skiff again. Docker is `make skiff/up`

ZenGround0 (approved):

Issue #1303, opened 2026-06-11
pdp: PieceGC 24h window destroys uploaded-but-uncommitted pieces when commits lag uploads


TippyFlitsUK:
**Setup:** curio 1.28.2-rc2 (mainnet, cc405c39), Storacha→FWSS bulk migration: pieces are uploaded ahead of `addPieces` commits, commit lane runs hours-to-days behind.

`PDPv0_PieceGC` deletes any pieceref with `data_set_refcount = 0` older than a hardcoded 24 hours (`task_piece_gc.go`), which drops `parked_pieces.ref_count` to 0 and queues `DropPiece` to delete the data. A bulk migration breaks the 24h assumption silently: every pause or slowdown in the commit lane left uploaded pieces aging past the window. Over three days the GC destroyed **138,503 staged pieces (~35GB)** — escalating daily (7.4k → 12.2k → 74k) as the migration scaled. Committed pieces (refcount > 0) were never at risk; I verified all 1.2M attached refs still have intact data.

Two contributing findings:

1. **No error surfaces anywhere.** The client's commits later fail with `subPiece CID not found`, long after the data is gone. Nothing logs that staged data is being expired.
2. **Pieces under 127 bytes crash the GC run**: `PieceCidV2FromV1` fails (`payloadsize must be at least 127, but was 123`) and aborts the whole run — but each pieceref deletion is its own committed transaction, so destruction up to that point persists, and the task retries forever. One run survived 4.7 hours deleting ~500 refs/min before we caught it.

**Recovery (worked):** deleted pending `DropPiece` tasks (their `cleanup_task_id` claims prevent re-scheduling); patched the interval to 14 days locally; reconciled the migration tool's tracking DB against surviving piecerefs; restored the 138k pieces from backup and re-committing now.

**Concern:** any operator running a commit pipeline more than 24h behind uploads will hit this — nothing in docs or logs warns of it. Happy to provide the timeline data or PR the interval as a config option.

ZenGround0:
Thanks for filing.  This is a known concern.  I don't expect us to want to change behavior significantly but perhaps we could track totally unnused pieces in a table and log them on drop.  It will likely be pretty noisy for a big operation doing a long commit lag.

TippyFlitsUK:
Fixed in #1316: the sub-127-byte piece now skips instead of aborting the GC run, and the cleanup window is widened 24h -> 14 days. Closing.

TippyFlitsUK:
[...]

PR #1316, opened 2026-06-24
pdp: migration-scale hardening (poller, proving task, IPNI reclaim)

### LINKED ISSUE #1291: pdp: PDPv0_IPNI Max(50) storms YugabyteDB with 40001 serialization conflicts on the serial ad-chain head; dropped tasks strand piecerefs unannounced (cc405c39)

### LINKED ISSUE #1296: pdp: un-includable cleanupPieces tx wedges sender nonce queue; ProvPeriod then fails permanently after restart
**Setup:** curio 1.28.2-rc2 (calibnet, ae0548cb), ~14k data sets on one node.

cc @rvagg — this is the message backlog you spotted on ezpdpz2 yesterday; full diagnosis below.

A data set termination test run against this node triggered a wave of piece cleanup (a "DropPiece storm"). One resulting `pdp-cleanup-pieces` transaction (data set 12920, 178 pieces) was estimated at exactly 10B gas — the full block limit. At default premium it never won block selection, and nonce ordering queued 1,000 messages behind it (Rod independently observed ~483 pending nonce slots around the same time). Gas estimation simulates all pending messages from the sender, so estimates slowed to minutes and ProvPeriod tasks began timing out.

Restarting lotus and curio made it worse: ~600 overdue ProvPeriod tasks were claimed at once (the task has no `Max` in TypeDetails), every chain call queued past the 10s `EthCallTimeout`, and `MaxFailures: 3` + re-add next epoch kept the storm fed. Result: ~1,400 failures/min, zero successes, for 3+ hours across multiple restarts.

A third problem surfaced during recovery: `SenderETH.Send` calls `EstimateGas` with no context deadline (`sender_eth.go:287`), unlike the view calls which use the 10s `EthCallOpts`. 16 tasks whose estimate responses were lost during the storm sat blocked for 30+ minutes — with a concurrency cap in place they held every slot, starving the task type until another restart.

**Did not work:** restarting curio; raising `LOTUS_FVM_CONCURRENCY` to 32.

**Worked:**
1. `Max: taskhelp.Max(16)` on NextProvingPeriodTask — the failure storm stopped immediately.
2. Replacing the stuck tx with a 0-value self-send at the same nonce — the 1,000-message queue drained at ~68 msgs/block.
3. One further curio restart to clear the 16 tasks wedged in the timeout-less `EstimateGas`.

Node fully recovered: queue 782 → 0, zero failures since.

**Remaining concern:** the data set 12920 cleanup is still unsent — any retry presumably re-estimates ≥ block limit and re-wedges the queue. With more termination tests planned, this will likely recur.

Happy to provide goroutine dumps and timeline data, or PR the one-line `Max` change.

### LINKED ISSUE #1306: harmonytask: poller fetches the entire unowned queue per task start; drain rate degrades as queues deepen
Promoting a finding from #1291's investigation to its own issue so it's trackable.

`pollerTryAllWork` selects all unowned tasks per handler per cycle with no LIMIT, and the engine starts one task per cycle. At 47k queued tasks the scan took ~1.0s, and task-start gaps tracked the scan time exactly — the backlog throttled its own drain. A `LIMIT 512` on the fetch was worth ~4x drain rate alone on my node (running locally since 10 June, no observed downsides). Caveat noted in #1291: a LIMIT interacts with retry-wait ordering, so the right bound may need care.

Observed on v1.28.2-rc2/mainnet during a 47k-task IPNI backlog; details and measurements in #1291.

TippyFlitsUK:
Three small migration-scale hardening fixes.

## harmonytask poller (#1306)

`pollerTryAllWork` selected the entire unowned queue per handler per cycle with no `LIMIT`, and the engine starts one task per cycle, so drain rate degraded as queues deepened (at 47k queued tasks the scan took ~1s and task-start gaps tracked it exactly). Adds `LIMIT 512` to the fetch (~4x drain rate on a backlogged node; running locally since 10 June).

## proving task (#1296)

- `NextProvingPeriodTask` had no `Max` in `TypeDetails`. After a restart, ~600 overdue proving tasks were claimed at once, every chain call queued past `EthCallTimeout`, and `MaxFailures` + re-add kept the storm fed. Adds `Max(16)`.
- `SenderETH.Send` called `EstimateGas` with no context deadline (unlike the view calls). Tasks whose estimate response was lost could sit blocked for 30+ minutes holding a slot. The call is now bounded by `defaultEthCallTimeout`.

## IPNI stranded piecerefs (#1291)

When a `PDPv0_IPNI` task is dropped after `MaxFailures`, its pieceref keeps a non-null `ipni_task_id` pointing at the now-deleted task. `schedule()` only claims rows `WHERE ipni_task_id IS NULL`, so those pieces are never announced. `schedule()` now nulls out `ipni_task_id` for refs whose task no longer exists, so they become schedulable again. (The concurrency half of #1291 — honouring the injected limiter instead of the `Max(50)` hardcode — is already in `main`.)

Closes #1306
Closes #1296
Closes #1291

### DIFF
--- a/tasks/indexing/task_pdp_v0_ipni.go
+++ b/tasks/indexing/task_pdp_v0_ipni.go
@@ -325,6 +325,12 @@ func (P *PDPV0IPNITask) TypeDetails() harmonytask.TaskTypeDetails {
 }
 
 func (P *PDPV0IPNITask) schedule(ctx context.Context, taskFunc harmonytask.AddTaskFunc) error {
+	if _, err := P.db.Exec(ctx, `UPDATE pdp_piecerefs SET ipni_task_id = NULL
+		WHERE needs_ipni = TRUE AND ipni_task_id IS NOT NULL
+		  AND NOT EXISTS (SELECT 1 FROM harmony_task t WHERE t.id = ipni_task_id)`); err != nil {
+		return xerrors.Errorf("reclaiming stranded PDP IPNI piecerefs: %w", err)
+	}
+
 	// schedule submits
 	var stop bool
 	for !stop {
--- a/tasks/indexing/task_pdp_v0_ipni.go
+++ b/tasks/indexing/task_pdp_v0_ipni.go
@@ -325,10 +325,14 @@ func (P *PDPV0IPNITask) TypeDetails() harmonytask.TaskTypeDetails {
 }
 
 func (P *PDPV0IPNITask) schedule(ctx context.Context, taskFunc harmonytask.AddTaskFunc) error {
-	if _, err := P.db.Exec(ctx, `UPDATE pdp_piecerefs SET ipni_task_id = NULL
+	n, err := P.db.Exec(ctx, `UPDATE pdp_piecerefs SET ipni_task_id = NULL
 		WHERE needs_ipni = TRUE AND ipni_task_id IS NOT NULL
-		  AND NOT EXISTS (SELECT 1 FROM harmony_task t WHERE t.id = ipni_task_id)`); err != nil {
+		  AND NOT EXISTS (SELECT 1 FROM harmony_task t WHERE t.id = ipni_task_id)`)
+	if err != nil {
 		return xerrors.Errorf("reclaiming stranded PDP IPNI piecerefs: %w", err)
+	}
+	if n > 0 {
+		ilog.Infow("reclaimed stranded PDP IPNI piecerefs", "count", n)
 	}
 
 	// schedule submits
--- a/tasks/pdpv0/task_piece_gc.go
+++ b/tasks/pdpv0/task_piece_gc.go
@@ -141,7 +141,7 @@ func processIndexingAndIPNICleanup(ctx context.Context, db *harmonydb.DB, cfg *c
 										    JOIN parked_piece_refs ppr ON pr.piece_ref = ppr.ref_id
 										    JOIN parked_pieces pp ON ppr.piece_id = pp.id
 										WHERE pr.data_set_refcount = 0
-										  AND pr.created_at <= TIMEZONE('UTC', NOW()) - INTERVAL '24 hours'
+										  AND pr.created_at <= TIMEZONE('UTC', NOW()) - INTERVAL '14 days'
 										  AND NOT EXISTS (
 										      SELECT 1 FROM pdp_data_set_piece_adds a
 										      WHERE a.pdp_pieceref = pr.id
@@ -182,7 +182,8 @@ func processIndexingAndIPNICleanup(ctx context.Context, db *harmonydb.DB, cfg *c
 		}
 		pcidV2, err := commcid.PieceCidV2FromV1(pcid1, uint64(piece.RawSize))
 		if err != nil {
-			return xerrors.Errorf("failed to construct piece cid v2: %w", err)
+			log.Warnw("skipping piece with invalid commp v2; continuing GC run", "piece_cid", piece.PieceCID, "raw_size", piece.RawSize, "error", err)
+			continue
 		}
 
 		var skipLoop bool
--- a/tasks/pdpv0/task_piece_gc.go
+++ b/tasks/pdpv0/task_piece_gc.go
@@ -182,8 +182,7 @@ func processIndexingAndIPNICleanup(ctx context.Context, db *harmonydb.DB, cfg *c
 		}
 		pcidV2, err := commcid.PieceCidV2FromV1(pcid1, uint64(piece.RawSize))
 		if err != nil {
-			log.Errorw("skipping piece with invalid commp v2; continuing GC run", "piece_cid", piece.PieceCID, "raw_size", piece.RawSize, "error", err)
-			continue
+			return xerrors.Errorf("failed to construct piece cid v2: %w", err)
 		}
 
 		var skipLoop bool
--- a/tasks/pdpv0/task_piece_gc.go
+++ b/tasks/pdpv0/task_piece_gc.go
@@ -182,7 +182,7 @@ func processIndexingAndIPNICleanup(ctx context.Context, db *harmonydb.DB, cfg *c
 		}
 		pcidV2, err := commcid.PieceCidV2FromV1(pcid1, uint64(piece.RawSize))
 		if err != nil {
-			log.Warnw("skipping piece with invalid commp v2; continuing GC run", "piece_cid", piece.PieceCID, "raw_size", piece.RawSize, "error", err)
+			log.Errorw("skipping piece with invalid commp v2; continuing GC run", "piece_cid", piece.PieceCID, "raw_size", piece.RawSize, "error", err)
 			continue
 		}
 
--- a/tasks/pdpv0/task_piece_gc.go
+++ b/tasks/pdpv0/task_piece_gc.go
@@ -141,7 +141,7 @@ func processIndexingAndIPNICleanup(ctx context.Context, db *harmonydb.DB, cfg *c
 										    JOIN parked_piece_refs ppr ON pr.piece_ref = ppr.ref_id
 										    JOIN parked_pieces pp ON ppr.piece_id = pp.id
 										WHERE pr.data_set_refcount = 0
-										  AND pr.created_at <= TIMEZONE('UTC', NOW()) - INTERVAL '14 days'
+										  AND pr.created_at <= TIMEZONE('UTC', NOW()) - INTERVAL '24 hours'
 										  AND NOT EXISTS (
 										      SELECT 1 FROM pdp_data_set_piece_adds a
 										      WHERE a.pdp_pieceref = pr.id
--- a/harmony/harmonytask/harmonytask.go
+++ b/harmony/harmonytask/harmonytask.go
@@ -413,7 +413,7 @@ func (e *TaskEngine) pollerTryAllWork(schedulable bool) bool {
 		err := e.db.Select(e.ctx, &allUnownedTasks, `SELECT id, update_time, retries
 			FROM harmony_task
 			WHERE owner_id IS NULL AND name=$1
-			ORDER BY update_time`, v.Name)
+			ORDER BY update_time LIMIT 512`, v.Name)
 		if err != nil {
 			log.Error("Unable to read work ", err)
 			continue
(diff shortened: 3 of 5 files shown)


tasks/indexing/task_pdp_v0_ipni.go
```diff
@@ -325,6 +325,12 @@ func (P *PDPV0IPNITask) TypeDetails() harmonytask.TaskTypeDetails {
 }
 
 func (P *PDPV0IPNITask) schedule(ctx context.Context, taskFunc harmonytask.AddTaskFunc) error {
+	if _, err := P.db.Exec(ctx, `UPDATE pdp_piecerefs SET ipni_task_id = NULL
+		WHERE needs_ipni = TRUE AND ipni_task_id IS NOT NULL
+		  AND NOT EXISTS (SELECT 1 FROM harmony_task t WHERE t.id = ipni_task_id)`); err != nil {
+		return xerrors.Errorf("reclaiming stranded PDP IPNI piecerefs: %w", err)
+	}
```
LexLuthr: We should log this instead of doing it silently.

tasks/pdpv0/task_piece_gc.go
```diff
@@ -141,7 +141,7 @@ func processIndexingAndIPNICleanup(ctx context.Context, db *harmonydb.DB, cfg *c
 										    JOIN parked_piece_refs ppr ON pr.piece_ref = ppr.ref_id
 										    JOIN parked_pieces pp ON ppr.piece_id = pp.id
 										WHERE pr.data_set_refcount = 0
-										  AND pr.created_at <= TIMEZONE('UTC', NOW()) - INTERVAL '24 hours'
+										  AND pr.created_at <= TIMEZONE('UTC', NOW()) - INTERVAL '14 days'
```
LexLuthr: We should not be increasing this. GC should be 24 hours. Otherwise, we are serving data which has not owner. There is also the issue of client data being exposed.

tasks/pdpv0/task_piece_gc.go
```diff
@@ -182,7 +182,8 @@ func processIndexingAndIPNICleanup(ctx context.Context, db *harmonydb.DB, cfg *c
 		}
 		pcidV2, err := commcid.PieceCidV2FromV1(pcid1, uint64(piece.RawSize))
 		if err != nil {
-			return xerrors.Errorf("failed to construct piece cid v2: %w", err)
+			log.Warnw("skipping piece with invalid commp v2; continuing GC run", "piece_cid", piece.PieceCID, "raw_size", piece.RawSize, "error", err)
```
LexLuthr: Why change this? I don't see any reason why we are okay with invalid data and it should only be logged.

LexLuthr (changes requested):
GC delay has some real world consequences which impact product as a whole. Currently we are testing to interval produces a lot of churn but testing pain should not change expected product behaviour.

TippyFlitsUK:
For context on this PR overall: all of these fixes were surfaced by the Storacha migration, which is a bulk-ingest use of PDP rather than normal day-to-day operation. The migration pushed the commit lane days behind uploads and drove task queues far deeper than steady state, which exposed a few latent failure modes (the poller scan cost, the proving-task restart storm, GC aborting the whole run on one bad piece, and stranded IPNI piecerefs).

None of this is meant to change steady-state behaviour as a blanket default. The aim is to harden these paths so an exceptional bulk event doesn't cause data loss or wedge a task lane, while leaving normal operation as it is today. On the GC window point below, I have kept the 24h default rather than baking in a longer value, since the migration was a transient event rather than a reason to change the default.

tasks/indexing/task_pdp_v0_ipni.go
TippyFlitsUK: It now logs the count of stranded piecerefs reclaimed each run (when non-zero) rather than running silently.

tasks/pdpv0/task_piece_gc.go
TippyFlitsUK: agreed, reverted to 24h. The 14 days was only to stop staged pieces expiring mid-run during the Storacha migration, where the commit lane ran days behind uploads, but you are right that shouldn't change the default GC behaviour, unowned or exposed data is exactly what we want to avoid. Kept it as the hardcoded 24h to match the other GC timings here rather than adding config.

tasks/pdpv0/task_piece_gc.go
TippyFlitsUK: Good point, this shouldn't be silent or look acceptable. It isn't accepting invalid data, it only stops one un-constructable commp v2 (e.g. a sub-127-byte payload) from aborting the entire GC run. I've bumped it to Errorw in c8028f2 so a bad piece surfaces loudly rather than being swallowed, while the run continues past it instead of one row killing the whole pass.

snadrus (approved):

tasks/pdpv0/task_piece_gc.go
LexLuthr: I think this should be reverted. We should not have any entry where commP cannot be generated in the table. This is one of those cases which should never happen. What we are doing here is not just avoiding the abort but we are also breaking the invariant silently.

TippyFlitsUK:
Reverted in df0e2422. Agreed on all counts. A `parked_pieces` row whose commP v2 cannot be constructed is an invariant violation, and GC is the wrong layer to absorb it; the fix belongs at ingest so the row is never created. `task_piece_gc.go` now matches `main`, so this PR no longer touches PieceGC (both the commp-v2 skip and the earlier 24h window change are out). I've reopened #1303 to track the real fix at the source. The other three changes here (poller LIMIT, proving Max(16) and the EstimateGas timeout, IPNI reclaim) are unaffected.

tasks/pdpv0/task_piece_gc.go
TippyFlitsUK: Agreed, reverted, invariant belongs at ingest

LexLuthr (approved):

TippyFlitsUK:
```
Example violation (mainnet, data set 1293):

piece id 7296  — 119 bytes — bafkzcibcbablawg7ia4h35y3fopl43chmzxqiauvrcgwyme3dfhpuupolrfryfq
piece id 7587  — 118 bytes — bafkzcibcbebiuio6adxzmpnuw3il6fxfx3preyybtd52okbhicpysa4ih5gwmcy
piece id 19948 — 102 bytes — bafkzcibcdebm4lsyh4kpmmuyytoc2rh67726ltwck4uztydhwks4h7crgfa46ly

$ curl https://main.ezpdpz.net/pdp/data-sets/1293
HTTP 400
Invalid PieceCID: failed to construct v2 from v1: payloadsize data size must be at least 127, but was 119

One sub-127-byte piece fails PieceCidV2FromV1 during response construction, so the
entire data-set listing 400s — same failure class aborts a PieceGC pass and an
import-pieces batch.
```

Issue #449, opened 2025-03-27
PDP configuration values -- contract address and eth rpc api


ZenGround0:
### Checklist

- [x] This is **not** a question or a support request. If you have any Curio related questions, please ask in the [discussion forum](https://github.com/filecoin-project/curio/discussions).
- [x] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/curio/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [x] I **have** searched on the [issue tracker](https://github.com/filecoin-project/curio/issues) and the [discussion forum](https://github.com/filecoin-project/curio/discussions), and there is no existing related issue or discussion.
- [x] I am running the [`Latest release`](https://github.com/filecoin-project/curio/releases), or the most recent RC(release candidate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.

### Curio component

- [ ] PoRep Pipeline
- [ ] Snap Deals
- [ ] Market
- [ ] IPNI provider
- [ ] Curio UI
- [x] Other

### Improvement Suggestion

The curio config should allow for specifying eth rpc api and pdp verifier contract address.  These are currently hardcoded in deps/deps.go and pdp/contract/addresses.go respectively.

magik6k:
It really should just internally serve the load-balanced api to the geth rpc, go-jsonrpc and geth both have facilities to do that entirely in process

parthshah1:
I already have a working devnet of Curio PDP, with full FOC stack deployed. I have curio patch which takes these addresses via env var during runtime. 
Happy to contribute a PR for the same.

rvagg:
we're looking for config settings here rather than environment variables
so I guess that means we also need to insert default values for them that come from the existing values for calibnet or mainnet builds

ZenGround0:
Since #509 the eth api url is irrelevant.  Any remaining work would be taking addresses.go state and externalizing it to a config.  As @rvagg points out the most obvious use case for this is making it easy to do devnet testing.

LexLuthr:
This should not be part of the config. Devnet argument does not provide justification for this. All the contracts already support ENV based overriding. That should be more than enough for devnet. If any of them are missing ENV override, we can add those.

Config is user facing. We should never push unnecessary details there.

LexLuthr:
I think this issue should be closed. We already have everything in code to override the addresses.

| Address | Env override |
| --- | --- |
[...]

PR #1317, opened 2026-06-24
Curio & Piri become friends

frrist:
## Background Reading
[Curio & Piri Unification Doc](https://docs.google.com/document/d/17Ehd4hsYOtXE4k3YQyWyyAKV6ED0K3raXgc-pITMW2U/edit?usp=sharing)

## Why                                                                                                                                                                                                                    
                                                                                                                                                                                                                         
Piri (fil-forge) reuses Curio's PDP core — the harmonytask scheduler and the pdpv0 tasks — but couldn't import it: that path drags in filecoin-ffi (CGO), a Lotus node, and Curio's concrete storage types. This PR turns those hard dependencies into interfaces so the packages build and run in a CGO-free, Lotus-free consumer. Curio's own behavior and wiring are unchanged.                                                   
                                                                                                                                                                                                                         
## What changed                                                                                                                                                                                                           
                                                                                                                                                                                                                         
**Storage & chain interfaces** - pdp: depend on storage interfaces and an eth client
- pdpv0 task constructors take small interfaces: `PieceReader`, `ProofCacheStore`, `IndexCleaner` (`tasks/pdpv0/storage.go`) — instead of `*cachedreader.CachedPieceReader` / `*indexstore.IndexStore`. Curio's concrete types satisfy them as-is.                                                                                                                                                                                                    
- `FSRegister` reads the wallet balance via `ethClient.BalanceAt` instead of a Lotus StateGetActor call, dropping the `api.FullNode` parameter. Registration now works against any EVM/FEVM endpoint (incl. anvil devnets (which motivated this change)).  
                                                                                                                                                                                                                         
**FFI-free scheduler** — harmonytask: inject a ResourceInspector
- `harmonytask.New` takes a `resources.ResourceInspector`; `resources.Register` calls it instead of probing GPUs directly.
- The sole FFI touchpoint — the GPU probe — moves to a new Curio-only package `harmony/resources/ffigpu` (keeping the existing `ffi/skiff/darwin` build variants). `harmony/resources` and `harmony/harmonytask` no longer import `filecoin-ffi` and compile with `CGO_ENABLED=0`.
- `GpuOverprovisionFactor` moves next to the probe — which also fixes it being undefined under the skiff build tag.
- Curio call sites pass `ffigpu.Inspector{}`.

**Configurable contract addresses** — pdp: allow contract addresses to be set from config
- `ContractAddresses` / `ServiceRegistryAddress` / `USDFCAddress` resolved addresses solely from the network build tag + `CURIO_DEVNET_*` env vars, which bakes network selection into the binary and doesn't fit a config-driven consumer.                                                                                                                                                                                                
  - Adds `contract.SetAddresses(Addresses)`: a consumer installs addresses from its own config at startup, and the accessors prefer it, falling back to the build-tag defaults when unset.                                                                                                                                                                       
                                                                                                                                                                                                                         
## Breaking changes                                                                                                                                                                                                       
                                                                                                                                                                                                                         
`harmonytask.New` gains a ResourceInspector argument:
```go
  // before                                                                                                                                                                                                              
  harmonytask.New(db, tasks, hostAndPort)                                                                                                                                                                                
  // after                                                                                                                                                                                                               
  harmonytask.New(db, tasks, hostAndPort, ffigpu.Inspector{}) // Curio's FFI-backed GPU probe
```
Same-PR signature changes for the same reason: `resources.Register` (+`inspector`), `contract.FSRegister` (−`api.FullNode`), and the pdpv0 task constructors (interfaces instead of concrete types).                           
                                                                                                                                                                                                                         
## Compatibility                                                                                                                                                                                                          

No behavior change for Curio: the concrete types still satisfy the new interfaces, and `ffigpu.Inspector{}` reproduces the old GPU probe exactly (including `HARMONY_OVERRIDE_GPUS / HARMONY_GPU_OVERPROVISION_FACTOR`).


ZenGround0:
I am having a decent amount of trouble understanding exactly what is being proposed in the doc.  This initial code change is ambiguous enough that it isn't helping me 

On one hand I see 
> Core PDP tasks (create, delete, add‑pieces, init/next proving period, prove): SHARED

I read that and think it makes sense.  But then I read:

> We are not proposing to share the broad PDP operations (e.g. "add a piece," "create a data set") as cross‑project library calls.

Which sounds exactly like not sharing those core tasks.  

Some thoughts:
1. The bulk of tricky protocol logic is encoded within task logic.  That is what we want to share.
2. Sharing things like chain accessors is easy but not that valuable.  Again a lot of the actual protocol implementation shows up within tasks.  For example to support piece removals well there is logic that should live in the nextProvingPeriod task.  Doing proving correctly and efficiently is handled in the proving and prove caching tasks.  Just having a shared way to call `nextProvingPeriod` on top of an eth api is a small part of the annoying logic we'd like to share.
3. The DB schemas for tasks are tightly coupled to task implementation.  Trying to separately implement two tasks but keep the same DB schema seems wrong.  You end up rewriting pretty much the same code in Piri as you would in curio.

To me the minimum viable shared library kind of has to rip away all of Piri PDP subsystem and replace it wholesale with curio task engine + curio DB schema + curio PDP tasks.  I believe there would be some thorny issues to address for such a replacement to be suitable for Piri's consumption.

frrist:
> I am having a decent amount of trouble understanding exactly what is being proposed in the doc                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                                                      
Sorry, I've muddled two different boundaries together. The things I'm trying to separate are:
- **Intake**: the front door that accepts a request (Piri's UCAN API vs Curio's HTTP handlers); the bit a client calls. I'm proposing this stays distinct per-project. It's what I'm referring to with the "broad PDP operations (e.g. 'add a piece,' 'create a data set')" statement.                                                                                                  
- **Pipeline tasks**: the workers that take queued work to chain + state (add-pieces, init/next proving period, prove, …). These we **do** want to share, and in both projects they're usually triggered via the Intake boundary.

So "we're not sharing the broad operations" means the intake / calling surface, not the tasks. The tasks are the whole point, and the part of the system I'm proposing we share. (I am also proposing we share other things to make a shared task library easier to implement like chain/contract accessors and commp calculation. Basically types and functions that operate on those types)                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                                      
---                                                                                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                                      
>The bulk of tricky protocol logic is encoded within task logic. That is what we want to share.                                                                                                                                                                                                                                                                                    
> ...
> having a shared way to call nextProvingPeriod on top of an eth api is a small part of the annoying logic we'd like to share.                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                                      
Yes, 100%. These are the most valuable parts of the system to share, in my book.                                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                                                      
---                                                                                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                                      
> To me the minimum viable shared library kind of has to rip away all of Piri PDP subsystem and replace it wholesale with curio task engine + curio DB schema + curio PDP tasks. I believe there would be some thorny issues to address for such a replacement to be suitable for Piri's consumption.                                                                               
                                                                                                                                                                                                                                                                                                                                                                                      
Yes, and this is what I attempted (rather poorly, it seems) to outline in [7.B](https://github.com/filecoin-project/curio/blob/6adfa7ded6581fefcc93bdcdd0ec90c75ae5235d/PDP-Unification-Strategy.md#pillar-b--shared-scheduler--core-tasks-highest-value-the-big-lift):                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           
  - Piri adopts Curio's concrete harmonytask and the concrete tasks; Curio's runtime is unchanged.                                                                                                                                                                                                                                                                                    
  - Finish making harmony/{harmonytask, harmonydb, resources, taskhelp} cleanly consumable.                                                                                                                                                                                                                                                                                           
  - Piri migrates domain persistence off GORM onto harmonydb — i.e. Piri runs Curio's PDP schema, not a reimplementation of it.                                                                                                                                                                                                                                                       
  - Share the core PDP tasks themselves, concretely.                                                                                                                                                                                                                                                                                                                                  

So the framing isn't extract code into a new shared library — it's make _Curio itself consumable as a library_ by Piri.                                                                                                                                                                                                                                                               

---

You also caught a real issue in the doc that I sorta glossed over: [§7.2](https://github.com/filecoin-project/curio/blob/6adfa7ded6581fefcc93bdcdd0ec90c75ae5235d/PDP-Unification-Strategy.md#72-the-shared-tasks-and-their-seams) still lists a `Store` interface that each project would implement "over its own platform." That's the schema-abstraction I think you're rejecting? An example where its "used" in Curio right now is:
- https://github.com/filecoin-project/curio/blob/6adfa7ded6581fefcc93bdcdd0ec90c75ae5235d/tasks/pdp/task_prove.go#L454
- https://github.com/filecoin-project/curio/blob/6adfa7ded6581fefcc93bdcdd0ec90c75ae5235d/tasks/pdp/task_prove.go#L572

My read of it based on Curio is: `mk20` is an accessor over `market_mk20_deal`, and the watchers mutate that table to report PDP outcome back to the deal. I suspect an interface over these methods would be quite thin and easy for Piri to implement, but I also get the sense this is a part of Curio the team might want to refactor since its mixing the PDP and Market boundry(? idk maybe not)
(I've also realized there are two pdp task packages in Curio, `pdp` and `pdp0` so maybe I am looking at the wrong package? Actually, `pdp0` doesn't have this mixing of market/pdp pipelines, so maybe this is easier than I initally thought?)

The other thorny bits you mention are real, and they're where I'd most value your read:                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                                      
1. PDP vs market coupling: some tasks reach into mk20 (e.g. prove's cache-rebuild → DealFromDB / market_piece_deal) which was my initial motivation for the `Store` interface. Which tasks are already pure-PDP (read only from pdp_*) and cleanly adoptable, and which need the market coupling factored out first?                                                                                                                         
2. The feed-in: in Curio the market pipeline populates pdp_pipeline / parked_pieces; for Piri that's the UCAN accept path. Where's the clean seam between intake and the shared pipeline?

ZenGround0:
Yeah pdp0 is the place to look, pdp package is unused

Another issue is piri's need for s3 backend storage.  We're going to need to rewrite tasks in ways I haven't fully understood yet to accomodate that.

frrist:
> Yeah pdp0 is the place to look, pdp package is unused

Ah, well good to know, I was partly looking at the wrong package. That actually makes things easier: `pdp0` operates only over the `pdp_*` tables, so the market/PDP mixing I was worried about in the other package just isn't a factor here.

> Another issue is piri's need for s3 backend storage. We're going to need to rewrite tasks in ways I haven't fully understood yet to accomodate that.

This one feels fairly tractable to me, let me lay out the task dependencies to show why.

The non-storage deps are all things Piri already has or can import directly:                                                                                                    
- `harmonydb.DB`: Piri takes a dep on harmony and runs the tasks' schema.                                                                                                        
- `chainsched.CurioChainSched`: Piri imports Curio's (it has its own port it can drop). Its only dep is a thin chain interface Piri already satisfies.                           
- `ethchain.EthClient`: an interface; Piri already has one.                                                                                                                      
- `api.Chain` (Lotus): an interface; Piri already has this.                                                                                                                      
- `message.SenderETH`: Piri has a near-identical version already; it'd just depend on Curio's.

That leaves storage. Across the tasks we'd actually share (prove / save_cache / maybe piece_gc) it's just two deps:                                                                   
- `cachedreader.CachedPieceReader`: reads raw piece bytes by CID, I think? (from Curio's sealed sectors + parked pieces, with a reader cache in front).                                    
- `indexstore.IndexStore`: the Merkle-proof-layer cache, yeah? (save_cache writes it, prove reads it, piece_gc deletes it).

And the good news: both of those already have interfaces in `lib/proof`: `proof.PieceReader` and `proof.ProofCache`. And `task_prove` already adapts the concrete `cachedreader`/`indexstore` to them internally. So making the tasks storage-agnostic is mostly "depend on the interface, not the concrete type," and a consumer can abstract its whole data store behind one seam:

```go
// everything the shared PDP pipeline tasks need from storage                                                                                                                   
  type PieceStore interface {                                                                                                                                                     
      proof.PieceReader // GetPieceReader(ctx, pieceCid) -> (reader, size)                                                                                                        
      proof.ProofCache  // GetLayerIndex / GetLayer / GetNode  (read, for prove)                                                                                                  
      AddProofLayer(ctx context.Context, piece cid.Cid, layer [][32]byte) error  // save_cache                                                                                    
      DeleteProofLayer(ctx context.Context, piece cid.Cid) error                 // piece_gc                                                                                      
  }                                                                                         
```

Curio would then satisfie this interface with `cachedreader` + `indexstore`; Piri satisfies it with S3 + whatever it wants for the cache, which is where Piri picks up the proving optimization it's after. The task logic never knows the difference.

PDP-Unification-Strategy.md
```diff
@@ -0,0 +1,317 @@
...
+### Cross‑cutting
+- **Packaging & ownership:** decide where shared code lives and who owns it (§8).
+- **Cross‑repo CI:** build both Curio and Piri against shared‑code HEAD to catch breakage early.
+- **Contract bindings:** reconcile the canonical contract version, then unify bindings (§8) — *after* the protocol layer is shared via the binding‑agnostic approach.
+
+---
+
+## 8. Decisions We Need (alignment items)
+
+1. **Where does the shared code live, and who owns it?** Options: (a) Piri imports the relevant Curio packages directly (fastest; Piri has indicated willingness); (b) a new standalone, jointly‑owned module; (c) fold the protocol layer into the existing external bindings module. Determines the dependency graph, the seam‑interface home (`Store`/`Verifier`/`ChainAPI`/`PieceReader`), and the release process.
```
alanshaw: 🤷‍♂️ It feels excessive to me for Piri to depend on the whole of Curio. So I am pro-shared library.

TIL: the harmony DB code in Curio is copied from Lotus. Perhaps a shared library has more consumers than just Curio & Piri?

ZenGround0: Harmony DB is not copied from lotus.  Where did you hear that?  There is one small piece of the storage system that came from lotus-miner which is maybe what is causing confusion.

pdp/contract/utils.go
```diff
@@ -275,19 +273,20 @@ func FSRegister(ctx context.Context, db *harmonydb.DB, full api.FullNode, ethCli
...
 		return fmt.Errorf("failed to parse 5 FIL: %w", err)
 	}
 
-	token := abi.NewTokenAmount(amount.Int64())
-
-	if ac.Balance.LessThan(big.NewInt(token.Int64())) {
+	// Check the wallet balance via the eth client (BalanceAt) instead of a
+	// Filecoin StateGetActor call, so registration works against any EVM/FEVM
+	// endpoint (incl. anvil devnets) without requiring a Lotus node. FIL and the
+	// FEVM balance share 18 decimals, so the 5 FIL deposit compares directly.
```
ZenGround0: This comment makes sense in the context of this diff but can be removed since it is not that useful for understanding the code after the fact.

pdp/contract/utils.go
```diff
@@ -236,7 +234,7 @@ func GetDataSetMetadataAtKey(ctx context.Context, listenerAddr common.Address, e
 	return out.Exists, out.Value, nil
 }
 
-func FSRegister(ctx context.Context, db *harmonydb.DB, full api.FullNode, ethClient ethchain.EthClient, name, description string, pdpOffering PDPOfferingData, capabilities map[string]string) error {
+func FSRegister(ctx context.Context, db *harmonydb.DB, ethClient ethchain.EthClient, name, description string, pdpOffering PDPOfferingData, capabilities map[string]string) error {
```
ZenGround0, line 237: Nice simplification

tasks/pdpv0/storage.go
```diff
@@ -0,0 +1,50 @@
...
+type IndexCleaner interface {
+	RemoveIndexes(ctx context.Context, pieceCidv2 cid.Cid) error
+}
+
+// PieceStorage writes and removes piece data in local (sector / piece-park)
+// storage. Satisfied by *ffi.SealCalls.
+type PieceStorage interface {
+	WriteUploadPiece(ctx context.Context, pieceID storiface.PieceNumber, size int64, data io.Reader, storageType storiface.PathType, verifySize bool) (abi.PieceInfo, uint64, error)
+	RemovePiece(ctx context.Context, id storiface.PieceNumber) error
+}
```
ZenGround0: Probably good to add in the classic go interface assertion statements. Either here or in the relevant packages implementing these interfaces.

i.e. var _ IndexCleaner = (*indexstore.IndexStore)(nil)

ZenGround0 (commented):
Still a draft so I'll hold off on a full approval.  But this looks good so far

LexLuthr (commented):
Addresses should not be config. We had a similar issue https://github.com/filecoin-project/curio/issues/449 and I would reiterate my stand on this. Config should be treated with utmost regard, it is the common surface between user and devs apart from GUI. We should always strive to make it clean and simple. This is dev side of stack bleeding into UX.

PDP-Unification-Strategy.md
alanshaw: IIRC I was looking that this commit https://github.com/filecoin-project/curio/commit/590a0bc82dbfbf0b53d3eee73b150d7752262e45 - is this the code you are referring to?

frrist:
@LexLuthr Humm, I think the commit wording ("set from config") caused a misread? 

To be concrete about what this change (9c09686) does and doesn't do: 
- there's no new TOML, GUI, env var, etc. — nothing an operator can see or touch has changed.
- Curio's own address resolution (build-tag defaults + `CURIO_DEVNET_*`) is unchanged when nothing calls the new `SetAddresses` function.
- User configuration surface and behavior are identical before and after
- mainnet curio build still cannot be pointed at a different contract without recompilation. (If anything, a code-only API is a smaller operator surface than an env var.)

The flexibility lands exclusively with a developer linking Curio as a library, which is what this PR aims to enabled via Piri. I'll acknowledge upfront that consuming Curio as an embedded library is not what Curio was designed for (as far as I know) — that's the broader conversation this PR is part of. But given that's the direction we're exploring, the existing env overrides can't serve it: they're only honored under the `2k`/`debug` build tags — and those tags also change chain parameters (4s block delay, testnet address encoding via `SetAddressNetwork`), so there's no build-tag combination that gives a production network's parameters with overridable addresses. That's why this needed a new function rather than more tags.

Would acceptable middle ground be an update to the `SetAddresses` function outlineing its intended use more clearly?

ZenGround0 (approved):
Yeah the struct and setter function seems fine to me.  I think this is good to go.  Speak now or forever hold your peace @snadrus and @LexLuthr

snadrus (approved):
This looks workable

PDP-Unification-Strategy.md
ZenGround0: I see. I think curio started getting developed in lotus repo but it is not actively used nor ever incorporated into lotus-miner.  That commit is basically a`mv`

PR #1319, opened 2026-06-28
fix: market add-url file mode

hyunmoon:
Fixes `curio market add-url --file` mode.

The previous implementation had a few issues:

- `--url` was required even when `--file` was used.
- file mode parsed rows with `strings.SplitN(line, ",", 4)` and then accessed `parts[3]` unconditionally, causing a panic for valid 3-column rows like `uuid,raw_size,url`.
- the no-header file-mode insert path had a column/value mismatch.
- file mode did not return before falling through to the single-deal path.

Changes:

- Use `encoding/csv` for file input.
- Allow `--file` mode without `--url`.
- Validate file mode and single-deal mode separately.
- Support documented CSV rows: `uuid,raw size,url,header1,header2...`.
- Insert `{}` headers when no headers are provided.
- Return immediately after processing file mode.

Testing:

- Verified that `curio market add-url --file <csv>` no longer requires `--url`.
- Verified that a 3-column CSV row does not panic.
- Verified that file-mode rows insert into `market_offline_urls`.

### DIFF
--- a/cmd/curio/market.go
+++ b/cmd/curio/market.go
@@ -1,9 +1,10 @@
 package main
 
 import (
-	"bufio"
+	"encoding/csv"
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
 	"os"
 	"strconv"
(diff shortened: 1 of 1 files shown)


ZenGround0 (approved):
LGTM

PR #1324, opened 2026-07-01
Collapse simple IAmBored schedulers into single SQL updates

### LINKED ISSUE #1265: Convert schedule functions for IAmBored to single execution
We do scheduling for IAmBored in a schedule function. Typically, this looks like below.

```go
addTaskFunc(func(taskID harmonytask.TaskID, tx *harmonydb.Tx) (shouldCommit bool, seriousError error) {
			stop = true

			var pendings []struct {
				ID int64 `db:"id"`
			}

			err := tx.Select(&pendings, `SELECT id
				FROM pdp_delete_data_set
				WHERE cleanup_pieces_task_id IS NULL
				  AND cleanup_pieces_tx_hash IS NULL
				  AND after_delete_data_set = TRUE
				  AND delete_tx_hash IS NULL
				  AND service_termination_epoch IS NOT NULL
				  AND terminated = FALSE
				ORDER BY id`)
			if err != nil {
				return false, xerrors.Errorf("failed to select pending PDP cleanup data sets: %w", err)
			}

			if len(pendings) == 0 {
				log.Debugw("no pending PDP data sets for piece cleanup")
				return false, nil
			}

			pending := pendings[0]

			n, err := tx.Exec(`UPDATE pdp_delete_data_set
				SET cleanup_pieces_task_id = $1
				WHERE id = $2
				  AND cleanup_pieces_task_id IS NULL
				  AND cleanup_pieces_tx_hash IS NULL
				  AND after_delete_data_set = TRUE
				  AND delete_tx_hash IS NULL
				  AND service_termination_epoch IS NOT NULL
				  AND terminated = FALSE`, taskID, pending.ID)
			if err != nil {
				return false, xerrors.Errorf("failed to assign PDP cleanup task: %w", err)
			}
			if n != 1 {
				return false, xerrors.Errorf("updated %d rows assigning PDP cleanup task", n)
			}

			log.Debugw("scheduled PDP cleanupPieces task", "dataSetId", pending.ID)
			stop = false
			return true, nil
		})
```

We can combine above in a single SQL execution with just setting the ID for first row with condition and then return true if n == 1. This pattern can be used for all schedule function which do not require additional processing before assigning the task.

LexLuthr:
fixes https://github.com/filecoin-project/curio/issues/1265


tasks/gc/task_cleanup_piece.go
```diff
@@ -450,23 +448,29 @@ func (p *PieceCleanupTask) schedule(ctx context.Context, taskFunc harmonytask.Ad
...
+				FROM pending
+				WHERE p.id = pending.id
+				  AND p.pdp = pending.pdp
+				  AND p.cleanup_task_id IS NULL
+				  AND p.after_cleanup = FALSE`, id)
 			if err != nil {
 				return false, xerrors.Errorf("failed to update piece_cleanup: %w", err)
 			}
+			if n == 0 {
+				return false, nil
```
ZenGround0, line 469: I guess ErrNoRows case from above would fall into this one by semantics of tx.Exec?

tasks/indexing/task_pdp_indexing.go
```diff
@@ -315,28 +315,31 @@ func (P *PDPIndexingTask) schedule(ctx context.Context, taskFunc harmonytask.Add
...
+					  AND indexing_task_id IS NULL
+					  AND indexed = FALSE
+					ORDER BY indexing_created_at ASC
+					LIMIT 1
+				)
+				UPDATE pdp_pipeline p
+				SET indexing_task_id = $1
+				FROM pending
+				WHERE p.id = pending.id
+				  AND p.aggr_index = pending.aggr_index
```
ZenGround0, line 328: I notice that this is a slight behavior change.  It looks probably more correct.  Just double checking its intentional.

ZenGround0 (approved):

tasks/gc/task_cleanup_piece.go
LexLuthr: Yes.

tasks/indexing/task_pdp_indexing.go
LexLuthr: Yeah. This is actually safer. But at this point aggr_index should always be 0.

Issue #1329, opened 2026-07-02
pdp: removal batches within allowed limits make nextProvingPeriod un-mineable; after 5 failures curio auto-terminates the client's data set


TippyFlitsUK:
**Setup:** curio self-built from recent main (includes #1270 termination handling, #1220 chain-state sync, #1309 batched DeletePiece), mainnet, ezpdpz-main. Data set 1245 — Covalent's Storacha-migration data set, ~186k live pieces (206,250 piece ids allocated), active client payment rails.

As part of the Storacha migration we needed to remove 19,780 duplicate pieces from Covalent's data set 1245. On 12 June we scheduled the first 1,503 removals via `schedulePieceDeletions` — 500 per call per the handler cap, well under the contract's ~2,000-per-proving-period limit. Every call was accepted.

Scheduled removals are executed inline by the next `nextProvingPeriod` call, and on a data set this size that combined transaction exceeds the block gas limit. Estimation reverts out of gas, so curio never submits it — no failed message on-chain, no FaultRecord, nothing for chain-event monitoring to see. Proving for this data set silently stopped on 13 June. Both caps passed, the state they permitted was un-mineable, and we found no way to un-schedule removals once accepted.

```
Next challange window scheduling failed … failed to estimate gas: chain: message
execution failed (exit=[33], … out of gas (7))   (task_next_pp.go:143)
```

On 1 July at 15:05 UTC, `task_chain_sync.go:435` reset the data set's proving-failure state (`reset PDP data set proving failure state after confirmed PDPVerifier progress`, `lastProvenEpoch: 6102318` — unchanged since 13 June), arming a clean failure counter. Five consecutive estimation failures followed with backoff (`failure_handling.go:106`, 100→800 blocks). At 03:47 UTC on 2 July the data set was marked `terminated due to repeated failures` (`failure_handling.go:87/:155`), and at 05:00 UTC curio sent and confirmed FWSS `terminateService`: tx `0x3b973bf3fb06342c71fdcb5fe3f1254736f3eba630ed902c579eec179cbea8d1` — `ServiceTerminated` + `PDPPaymentTerminated` emitted, approver = the SP session wallet. Covalent did not request termination and the operator was not involved or notified.

**Result:** a paying migration customer's data set was terminated on-chain by node automation, three weeks downstream of a duplicate-removal operation that stayed inside every advertised limit. Termination is irreversible — we are recommitting the full data set (~186k pieces) from backup.

[...]

PR #1331, opened 2026-07-02
Be 10x more conservative on the removal queue

ZenGround0:
This is a quick fix that should resolve @TippyFlitsUK's recent issue #1329 for practical gas values.

It simply stops submitting removal requests once the queue is up to 200 entries.  A proper solution will need to incorporate contract upgrades described in https://github.com/FilOzone/pdp/issues/283.  For now this should prevent us from getting into bad states at the cost of deletions being lower throughput.

### DIFF
--- a/tasks/pdp/task_delete_piece.go
+++ b/tasks/pdp/task_delete_piece.go
@@ -164,32 +164,52 @@ func (p *PDPTaskDeletePiece) TypeDetails() harmonytask.TaskTypeDetails {
 func (p *PDPTaskDeletePiece) schedule(ctx context.Context, taskFunc harmonytask.AddTaskFunc) error {
 	var stop bool
 	for !stop {
+		// Pick the next pending delete and read its data set's on-chain removal queue
+		// length *before* claiming the task, so the eth_call stays out of the DB tx.
+		var did string
+		var setID int64
+		err := p.db.QueryRow(ctx, `SELECT id, set_id FROM pdp_piece_delete
+								  WHERE task_id IS NULL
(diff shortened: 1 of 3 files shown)


tasks/pdp/task_delete_piece.go
```diff
@@ -164,32 +164,52 @@ func (p *PDPTaskDeletePiece) TypeDetails() harmonytask.TaskTypeDetails {
...
-				}
-				return false, xerrors.Errorf("failed to query pdp_piece_delete: %w", err)
-			}
-			if did == "" {
-				return false, xerrors.Errorf("no valid deal ID found for scheduling")
+		// Pick the next pending delete and read its data set's on-chain removal queue
+		// length *before* claiming the task, so the eth_call stays out of the DB tx.
+		var did string
+		var setID int64
+		err := p.db.QueryRow(ctx, `SELECT id, set_id FROM pdp_piece_delete
```
ZenGround0, line 171: I was hesitant to do the pdpverifier ethcall query within the tx function.  But if that was a stupid worry then I'll get rid of the two separate db calls.

snadrus: I agree that it's best to keep transactions light as sometimes they're replayed 10 times before they succeed

LexLuthr (approved):

tasks/pdp/task_delete_piece.go
LexLuthr: Correct place to handle this would be in Do() or in the api itself

LexLuthr: This is pdp_v1. No need to fix it or worry about it.

PR #1334, opened 2026-07-03
fix: stop panic on shutdown

LexLuthr:


cmd/curio/main.go
```diff
@@ -50,7 +50,6 @@ func setupCloseHandler() {
 	go func() {
 		<-c
 		fmt.Println("\r- Ctrl+C pressed in Terminal")
-		panic(1)
 	}()
```
snadrus, line 63: this panic may need to become os.Exit(1) or detect if we are shutting-down, but as-is this blocks ctrl-c handling

snadrus: Leave this (with comments) for CLI. 
"close" it with a function before calling MonitorShutdown.

cmd/curio/run.go
```diff
@@ -135,16 +146,20 @@ var runCmd = &cli.Command{
 		}
 		defer taskEngine.GracefullyTerminate()
 
-		err = rpc.ListenAndServe(ctx, dependencies, shutdownChan) // Monitor for shutdown.
-		if err != nil {
+		errCh := make(chan error, 1)
+		go func() {
+			errCh <- rpc.ListenAndServe(ctx, dependencies, shutdownChan) // Monitor for shutdown.
```
snadrus: This conflicts with the graceful restart PR goal of port handoff. Of-course, we could rely on third-party front-ends like Caddy to do something similar.

snadrus (approved):

snadrus:
Consider adding a test that uses Ctrl-C

Issue #897, opened 2026-01-21
pdpv0: settlement watcher should process failed settlements


ZenGround0:
This is a side quest issue to the main track of #871

Rail settlement is robust.  Given that we have a watcher which looks out for persistent settlement failures and calls terminate and given that [terminated rails always settle](https://github.com/filecoin-project/curio/issues/871#issuecomment-3776098117) we should never get stuck.  

However we have a problem: the payments watcher only runs verification on [DB inserted rail IDs](https://github.com/filecoin-project/curio/blob/pdpv0/tasks/pay/watcher.go#L43).  And only rail IDs belonging to successfully settled unterminated rails are [inserted into the DB](https://github.com/filecoin-project/curio/blob/pdpv0/lib/filecoinpayment/utils.go#L131-L136).  This is made somewhat worse by the batching mechanism since up to 9 successfully settled rails can also be ignored this way.  Because of determinism in the contract getRailsForPayeeAndToken function and in our batching logic, any failing rail will force the ignoring of a stable set which can lead to perpetual ignoring and non-settlement.

The correct solution is to use [the same chain call](https://github.com/filecoin-project/curio/blob/pdpv0/lib/filecoinpayment/utils.go#L50) used during settlement within the watcher code.  Then problem rails will be targeted for termination and they will immediately start settling and unblocking settlement of the unlucky rails grouped with them.

ZenGround0:
As I've now learned that settlement errors are not expected under any circumstances in fwss I am going to defer the main fix here to lower priority and leave this issue open.  I'll open a new issue to track the lack of message waiting in the watcher which is a much more concerning and expected to be immediately buggy issue.

LexLuthr:
@ZenGround0 Hey! Is this still relevant? If not, can we close it out?

beck-8:
I added a log to see which railiD it was, still waiting for the results
```
{"level":"error","ts":"2026-06-04T21:49:35.732+0800","logger":"filecoin-pay","caller":"filecoinpayment/utils.go:136","msg":"failed to send settle transaction","error":"failed to estimate gas: chain: message execution failed (exit=[SysErrOutOfGas(7)], vm error=[message failed with backtrace:\n00: f0176012 (method 3844450837) -- out of gas (7)\n (RetCode=7)])"}
```

In addition, I saw that this task was not counted as failure. It should not meet expectations. @LexLuthr 

Settle | 2 | 0 (0%)

LexLuthr:
[...]

PR #1337, opened 2026-07-06
fix(message): apply 1.1x gas overestimation to eth transactions

beck-8:
`SenderETH.Send` uses the raw `eth_estimateGas` result as the transaction gas limit with no safety margin. The native message path gets a 1.25x buffer via lotus `GasEstimateMessageGas` (mpool default `GasLimitOverestimation`), but the ETH path has none, so state drift between estimation and inclusion can push actual usage past the limit and the transaction fails with `SysErrOutOfGas`, burning the full gas limit.

Real example (calibnet, `settleRail` for rail 12757, same class of failures discussed in #897):

- tx `0x492ec195183b9bfa96ccc98be6a6351b339ec05f50efb479f743198bf5f14b7e` was sent with gas limit 58,253,517 (the bare estimate) and failed on chain with `SysErrOutOfGas` at epoch 3864708.
- Replaying the same message on the parent tipset via `Filecoin.StateCall` with a raised gas limit succeeds with `GasUsed = 60,869,671` — about 4.5% above the estimate.

The failed settlement only surfaces as a `Settlements_FilecoinPay` alert (all tasks report success) and the rail is retried 12h later, so each occurrence wastes the full gas limit and delays settlement.

This applies the same 1.25 factor to gas limits estimated in `SenderETH.Send`, matching the native path. All FEVM transactions (settle, PDP prove, etc.) go through this path and benefit.

### DIFF
--- a/lib/filecoinpayment/utils.go
+++ b/lib/filecoinpayment/utils.go
@@ -26,6 +26,11 @@ import (
 
 var log = logging.Logger("filecoin-pay")
 
+// gasOverestimation is applied on top of the eth_estimateGas result when
+// sending settleRail transactions, since it has been observed to fall short
+// at execution time and burn the whole gas limit.
+const gasOverestimation = 1.1
+
 func SettleLockupPeriod(ctx context.Context, db *harmonydb.DB, ethClient ethchain.EthClient, sender *message.SenderETH, from common.Address, payees []common.Address, operators []common.Address, al curioalerting.AlertingInterface, system, subsystem string) error {
 	paymentContractAddr, err := PaymentContractAddress()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


LexLuthr:
While I agree with the change in principle, I think premise might be flawed here. We use same gas estimate method for both Eth and Filecoin message underneath the interfaces. Feel free to dispute this if I am wrong.

PDP already requires way too many messages from SP. With this single decision, we just pushed the SP operational cost to by 25%. My suggestion would be to be conservative here and do 10% max.

@ZenGround0 @snadrus I would like to have your opinions before we make a decision on this.

beck-8:
I just noticed that `eth_estimateGas` already looks like a multiple of 1.25. Here, I multiplied by 1.25 again. Maybe once the 897 issue is resolved, you won't need to worry about this anymore.

beck-8:
Sorry, I took a second look and realized this was indeed wrong. It shouldn't be done.

beck-8:
I reopened this and it looks like we may still need this.

LexLuthr (commented):
Run `make fix-imports` to fix the CI issue. This is still applying the 1.1 factor to all the Eth messages. In slack we agreed to only do that for settle messages. Can you please also implement that?

beck-8:
I'll deal with it right away.

beck-8:
@LexLuthr  done 🙏

LexLuthr (approved):

PR #1338, opened 2026-07-06
Handle large Filecoin payment settlement backlogs

### LINKED ISSUE #897: pdpv0: settlement watcher should process failed settlements

LexLuthr:
fixes https://github.com/filecoin-project/curio/issues/897

## Summary:
• Add gas estimation probing for settleRail and reduce the settlement target on OOG so large backlogs can still make forward progress.
• Record the intended settle_upto_epoch with settlement tx tracking.
• Use the recorded target in the settlement watcher to verify the tx reached the intended epoch before running default handling.
• Wire alerting into the settle task path for settlement estimation/send failures.


harmony/harmonydb/downgrade/20260706-filecoin-pay-settle-epoch.sql
beck-8, line 1: Is this an empty file? @LexLuthr

LexLuthr: Yes. We don't downgrade column add and table creates

LexLuthr: Add a comment to the file

tasks/pay/watcher.go
```diff
@@ -215,12 +217,24 @@ func verifySettle(ctx context.Context, db *harmonydb.DB, ethClient ethchain.EthC
...
 
-			// For live rails, check if we are fully unsettled 1 day before the lockup period ends.
-			// If so assume payer is in default and schedule deletion
-			threshold := big.NewInt(0).Add(view.SettledUpTo, view.LockupPeriod)
-			thresholdWithGrace := big.NewInt(0).Sub(threshold, big.NewInt(builtin.EpochsInDay))
+			var railDefaulting bool
 
-			if thresholdWithGrace.Uint64() < current {
+			if settle.SettleUpToEpoch.Valid {
+				if view.SettledUpTo.Int64() < settle.SettleUpToEpoch.Int64 {
```
beck-8: If you fall behind by one height here, you will break the contract and end the path to terminating the dataset.
This is not expected.

LexLuthr: That is expected. If user is not paying, we should be terminating. Why should we not?

LexLuthr: My bad! I missed it completely. Let me fix that

LexLuthr: Fixed

lib/filecoinpayment/utils.go
```diff
@@ -172,3 +212,44 @@ func SettleLockupPeriod(ctx context.Context, db *harmonydb.DB, ethClient ethchai
...
+
+func calculateSettleUpTo(ctx context.Context, pabi *abi.ABI, paymentContractAddr *common.Address, ethClient ethchain.EthClient, id *big.Int, from common.Address, settledUpTo, currentEpoch *big.Int) (int64, error) {
+	next := currentEpoch
+	for {
+		data, err := pabi.Pack("settleRail", id, next)
+		if err != nil {
+			return 0, xerrors.Errorf("failed to pack data: %w", err)
+		}
+
+		_, err = ethClient.EstimateGas(ctx, ethereum.CallMsg{
```
snadrus, line 224: is this subtracting from a remainder? This code seems confusing.

LexLuthr: I am guessing you mean code below this. 
1. We assign current to next. 
2. Then we try to gas estimate to see if the message can be executed.
3. If it fails due to out of gas then we half the subtract settledUpto from current i.e. find how many epochs we were trying to settle.
4. Divide that number by half and try again. Keep doing halving till it is successful or we hit settledUpto or other error.

beck-8:
I'll use my nodes to test and observe in practice.

snadrus (approved):

PR #1340, opened 2026-07-06
60 days lockup grace period

### LINKED ISSUE #1330: Extended Curio retention grace before lockup termination
What's happening is that our filecoin pay discovery mechanism is not great and so customers are missing the time window to refresh their funds.  I believe this [filecoin slack thread](https://filecoinproject.slack.com/archives/C0APZ22SJ2C/p1782755042915199) which @LexLuthr has been engaging in covers this.

For the sake of users getting used to our system there is a request to extend curio's internal settlement grace period beyond the 30 day protocol grace period.  

It should be noted this is a stopgap usability fix and it does put SPs at risk for delivering service without getting compensated.  However for building up business it is a fine tradeoff

I think this should be a simple matter of increasing our threshold constant for termination eligibility.  I recommend using 60 days to double to effective grace period before service cancellation. 

[gated notion link describing this and other related issues](https://app.notion.com/p/filecoindev/Temporary-extend-retention-grace-before-lockup-termination-Curio-391dc41950c181aaa4c4ccda718695de?source=copy_link)

LexLuthr:
fixes https://github.com/filecoin-project/curio/issues/1330

### DIFF
--- a/tasks/pay/watcher.go
+++ b/tasks/pay/watcher.go
@@ -23,8 +23,9 @@ import (
 )
 
 const (
-	alertName = "FilecoinPay"
-	alertType = "Settlements"
+	alertName                     = "FilecoinPay"
+	alertType                     = "Settlements"
+	temporaryDefaultGraceInEpochs = 30 * builtin.EpochsInDay
 )
 
 type settled struct {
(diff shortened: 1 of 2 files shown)


tasks/pay/watcher.go
```diff
@@ -23,8 +23,9 @@ import (
 )
 
 const (
-	alertName = "FilecoinPay"
-	alertType = "Settlements"
+	alertName                     = "FilecoinPay"
+	alertType                     = "Settlements"
+	temporaryDefaultGraceInEpochs = 30 * builtin.EpochsInDay
```
snadrus, line 28: looks like 30 instead of 60

LexLuthr: Lockup is 30 and then we add this to lockup.

snadrus (approved):

PR #1344, opened 2026-07-08
web: simplify PDP offering fields, add storage capacity, fix payment token overwrite on update

beck-8:
## Summary
- Update PDP Offering no longer exposes Min/Max Piece Size, Storage Price, Min Proving Period, or IPNI Piece/IPFS/Peer ID — none of these are enforced anywhere in Curio or the service contracts, they're purely self-reported registry metadata. Editing them added surface area without controlling any real behavior, so the update form now only lets you change Service URL, Location, and custom Capabilities; the other required capability keys are silently round-tripped from the provider's existing on-chain values so the update transaction still satisfies the contract's required-key check. The same fields are also dropped from the read-only status view for the same reason.
- Add a `capacityTiB` capability so providers can advertise total storage capacity, settable at registration and editable via Update PDP Offering, so a fresh registration is complete without a follow-up update.
- Fix `FSUpdatePDP` silently overwriting `paymentTokenAddress` with the zero address on every update (since `updateProduct` on `ServiceProviderRegistry` does a full replace of the capability set, not a merge, so every field must be resent). It now resolves the real USDFC address the same way registration does.
- Use named fields in an existing `PDPOfferingData` test literal so it doesn't need editing every time the struct grows a field.

## Test plan
- [x] `go build` / `go vet` / `go test` on `pdp/contract` and `web/api/webrpc`
- [x] Manual round-trip check: `PDPOfferingData{CapacityTiB: 500}` → `OfferingToCapabilities` → `capabilitiesToOffering` decodes back to 500
- [x] Manually exercised Register / Update PDP Offering modals and the read-only status view against a local `web/devsrv` build (screenshots to follow in a comment)

---
02-update-pdp-offering-modal

03-register-provider-modal


beck-8:
The last commit also removes these parameters from the state display.

LexLuthr (commented):
I am assuming something changed in registry contract to trigger these changes. If yes, can you please link the issue/PR and release details. This seems very concerning that contract changed significantly without Curio team having any clue.

beck-8:
I just see these fields seem useless for SP. So I want to reduce the burden on SPs. But I looked at these fields and contracts still need them. So they were simply removed from the UI.

pdp/contract/utils.go
```diff
@@ -46,6 +46,7 @@ const (
 	CapMinProvingPeriod = "minProvingPeriodInEpochs"
 	CapLocation         = "location"
 	CapPaymentToken     = "paymentTokenAddress"
+	CapCapacityTiB      = "capacityTiB" // Optional, advertised total storage capacity in TiB
```
ZenGround0: Please PR a reference to this in comment here:
https://github.com/FilOzone/filecoin-services/blob/main/service_contracts/src/ServiceProviderRegistry.sol#L22

ZenGround0 (approved):
I support this change.  Should more sophisticated registry usage and interaction between registry values and curio configuration become important we can easily reintroduce UI toggles for these things.  Since these defaults are for now enforced internally there is no reason to confuse users without benefits.

ZenGround0:
While I am no js expert the AI review's nitpicks seem worth addressing at first glance.

beck-8:
I'll check tomorrow morning; I remember making the changes as required. Maybe I forgot to push the update locally.

PR #1345, opened 2026-07-08
fix(market): use MK20 DDO duration for deal schedule

hyunmoon:
## Problem

MK20 DDO ingestion was not preserving the duration requested by the DDO.

The code computes the deal end epoch from the DDO duration, but then replaces it with `start + alloc.TermMin` when building the piece activation manifest for an allocation-backed DDO.

In practice this makes the sector deal schedule use the allocation minimum term instead of the DDO duration. For example, a DDO with a 5 year duration and an allocation with a 180 day `TermMin` is inserted as a 180 day deal schedule.

## Root Cause

The allocation `TermMin` was being used as the deal duration.

For MK20 DDO, the DDO duration should be the requested deal duration. The allocation terms should only be used to validate that the requested duration is allowed.

## Fix

Keep the DDO duration as the deal schedule duration.

When an allocation is present, validate that the requested duration is within the allocation bounds:

`TermMin <= duration <= TermMax`

This also fixes the MK20 DDO sanity check so invalid durations are rejected before ingestion instead of reaching this path.

## Testing

- `go test ./tasks/storage-market ./market/storageingest`
  - `github.com/filecoin-project/curio/tasks/storage-market` [no test files]
  - `github.com/filecoin-project/curio/market/storageingest` [no test files]

### DIFF
--- a/market/mk20/mk20.go
+++ b/market/mk20/mk20.go
@@ -478,14 +478,14 @@ func (m *MK20) sanitizeDDODeal(ctx context.Context, deal *Deal) (*ProviderDealRe
 		if alloc.TermMin > deal.Products.DDOV1.Duration {
 			return &ProviderDealRejectionInfo{
 				HTTPCode: ErrBadProposal,
-				Reason:   "Allocation term min is less than deal duration",
+				Reason:   "Allocation term min is greater than deal duration",
 			}, nil
 		}
 
-		if alloc.TermMax > deal.Products.DDOV1.Duration {
+		if alloc.TermMax < deal.Products.DDOV1.Duration {
 			return &ProviderDealRejectionInfo{
 				HTTPCode: ErrBadProposal,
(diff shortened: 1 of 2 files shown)


LexLuthr (changes requested):
I am surprised a bad deal reached this execution (misunderstood the PR summary) while passing the sanity checks. On review, I found another bug in sanity check https://github.com/filecoin-project/curio/blob/4ef2e4d4c10b63455a9c04ebcda8457a9cc87604/market/mk20/mk20.go#L485. This should be inverted. Can you please also fix that?

hyunmoon:
Good catch, thank you. You’re right, the TermMax check was inverted.

I pushed a follow-up commit that makes the sanity check reject durations outside the allocation bounds before ingestion, and keeps the ingestion-side validation consistent with that:

TermMin <= duration <= TermMax

I also fixed the TermMin error message while touching that block.

LexLuthr (approved):

PR #1346, opened 2026-07-08
Fix: FWSS settlement target epoch selection

LexLuthr:
## Problem
FWSS settlement currently asks Filecoin Pay to settle up to the current chain epoch. That is not always a valid target for FWSS rails because the FWSS validator only accepts settlement through closed proving-period boundaries. When the target falls inside an open proving period or outside the validator’s accepted range, gas estimation can fail with InvalidEpochRange even though the rail still needs settlement.

## Summary:
• Add operator-keyed settlement target resolvers to Filecoin Pay settlement.
• Keep common payment settlement FWSS-agnostic while letting FWSS provide validator-specific settle bounds.
• Cap FWSS settlement targets to the last closed proving-period deadline before gas estimation.
• Add tests

## ~Blocker~
- [x] Fix InvalidEpochRange

### DIFF
--- a/lib/filecoinpayment/error_detection.go
+++ b/lib/filecoinpayment/error_detection.go
@@ -5,29 +5,70 @@ import (
 	"strings"
 )
 
-// ErrSelectorRailInactiveOrSettled is the hex-encoded 4-byte selector for
-// FilecoinPayV1's RailInactiveOrSettled(uint256). Reverts carrying it prove
-// the rail has been finalised and zeroed.
-var ErrSelectorRailInactiveOrSettled string
+var (
+	// ErrSelectorRailInactiveOrSettled is the hex-encoded 4-byte selector for
+	// FilecoinPayV1's RailInactiveOrSettled(uint256). Reverts carrying it prove
+	// the rail has been finalised and zeroed.
+	ErrSelectorRailInactiveOrSettled string
+
+	skippableSettleRailErrorSelectors map[string]struct{}
+)
+
+// skippableSettleRailErrorNames is intentionally Pay-only and only for
+// settleRail simulation failures. Validator-specific errors belong in the
+// validator resolver; this list is for reverts where another pass can safely
+// retry or chain state already made this settlement attempt unnecessary.
+var skippableSettleRailErrorNames = []string{
+	"CannotSettleFutureEpochs",
+	"NoProgressInSettlement",
+	"RailInactiveOrSettled",
+}
 
 func init() {
 	parsed, err := PaymentsMetaData.GetAbi()
 	if err != nil {
 		panic("failed to parse FilecoinPay Payments ABI: " + err.Error())
 	}
-	e, ok := parsed.Errors["RailInactiveOrSettled"]
-	if !ok {
-		panic("FilecoinPay Payments ABI missing RailInactiveOrSettled error")
+
+	skippableSettleRailErrorSelectors = make(map[string]struct{}, len(skippableSettleRailErrorNames))
+	for _, name := range skippableSettleRailErrorNames {
+		e, ok := parsed.Errors[name]
+		if !ok {
+			panic("FilecoinPay Payments ABI missing " + name + " error")
+		}
+
+		selector := hex.EncodeToString(e.ID[:4])
+		skippableSettleRailErrorSelectors[selector] = struct{}{}
+		if name == "RailInactiveOrSettled" {
+			ErrSelectorRailInactiveOrSettled = selector
+		}
 	}
-	ErrSelectorRailInactiveOrSettled = hex.EncodeToString(e.ID[:4])
+}
+
+func isSkippableSettleRailError(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	for selector := range skippableSettleRailErrorSelectors {
+		if errorContainsSelector(err, selector) {
+			return true
+		}
+	}
+
+	return false
 }
 
 // IsRailInactiveOrSettledError reports whether err carries the
 // RailInactiveOrSettled revert, proving the rail has been zeroed during
 // finalisation. Distinguishes finalisation from transient RPC errors.
 func IsRailInactiveOrSettledError(err error) bool {
-	if err == nil {
+	return errorContainsSelector(err, ErrSelectorRailInactiveOrSettled)
+}
+
+func errorContainsSelector(err error, selector string) bool {
+	if err == nil || selector == "" {
 		return false
 	}
-	return strings.Contains(strings.ToLower(err.Error()), "0x"+ErrSelectorRailInactiveOrSettled)
+	return strings.Contains(strings.ToLower(err.Error()), "0x"+selector)
 }
--- a/pdp/contract/FWSS/settle_resolver.go
+++ b/pdp/contract/FWSS/settle_resolver.go
@@ -88,7 +88,7 @@ func standardSettleTarget(rail filecoinpayment.PaymentsRailView, currentEpoch *b
 	if rail.EndEpoch != nil && rail.EndEpoch.Sign() > 0 && rail.EndEpoch.Cmp(target) < 0 {
 		target.Set(rail.EndEpoch)
 	}
-	if target.Cmp(rail.SettledUpTo) <= 0 {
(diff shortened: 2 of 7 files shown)


rvagg:
@LexLuthr I think that all of this complexity goes away if we properly handle the case you identified in https://github.com/FilOzone/filecoin-services/pull/545 in the contract itself:

> `toEpoch < activationEpoch` still reverts `InvalidEpochRange`

Settling rate-based rails in the pre-activation range should just settle with a zero value, in the client's favour, because there's no proving begun. We just make that explicit: you don't pay for proving if the SP hasn't said it's started proving. Then almost all of this goes away and you just need graceful handling of the edge cases. Let me get up an filecoin-services PR and we can look at the model a little more carefully as it relates to Curio.

rvagg:
https://github.com/FilOzone/filecoin-services/pull/552

ZenGround0:
@LexLuthr still catching up but does the contract PR now enable us to close this curio PR down?

rvagg:
I've written up a list of things to care about in https://github.com/filecoin-project/curio/issues/1356; I think this PR can be closed but that list will need to be worked thorugh for a new iteration of it.

LexLuthr:
> I've written up a list of things to care about in #1356; I think this PR can be closed but that list will need to be worked thorugh for a new iteration of it.

I do not agree. FWSS PR only fixes https://github.com/filecoin-project/curio/issues/1353 not the whole validation. Curio must  still validate rails before we try to settle. We cannot keep wasting gas on settling things which gives no money to SP.

This PR should be merged with 1 change around settledEpoch == EndEpoch exclusion. Then a separate PR can be done once new contracts are deployed to address #1356 and #1353 collectively.

lib/filecoinpayment/error_detection.go
```diff
@@ -5,29 +5,70 @@ import (
...
+var (
+	// ErrSelectorRailInactiveOrSettled is the hex-encoded 4-byte selector for
+	// FilecoinPayV1's RailInactiveOrSettled(uint256). Reverts carrying it prove
+	// the rail has been finalised and zeroed.
+	ErrSelectorRailInactiveOrSettled string
+
+	skippableSettleRailErrorSelectors map[string]struct{}
+)
+
+// skippableSettleRailErrorNames is intentionally Pay-only and only for
```
ZenGround0, line 17: These names are coming from the contract?  Is it possible to get these from the abi?

pdp/contract/FWSS/settle_resolver.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/filecoin-project/curio/lib/ethchain"
+	"github.com/filecoin-project/curio/lib/filecoinpayment"
+	"github.com/filecoin-project/curio/pdp/contract"
+)
+
+// SettleTargetResolver builds a Filecoin Pay settlement target resolver for
+// FWSS rails. FWSS uses multiple rail types: PDP validator rails must respect
+// proving-period boundaries, while non-validator rails can use the normal
+// Filecoin Pay settlement target.
+func SettleTargetResolver(ctx context.Context, ethClient ethchain.EthClient) (filecoinpayment.SettleTargetResolver, error) {
```
ZenGround0, line 19: I think this should be renamed FWSSSettleResolver or something with FWSS in the name

ZenGround0 (commented):
@LexLuthr I've spent a lot of time learning about all of the recent changes.  Overall I am not seeing the problem warranting this specific change.  

This is very cleanly stated in the PR description and looks like the main problem: 
> When the target falls inside an open proving period or outside the validator’s accepted range, gas estimation can fail with InvalidEpochRange even though the rail still needs settlement.

But when I look at the service contract `validatePayment` function I see only two places raising this error

> require(totalEpochsRequested > 0, Errors.InvalidEpochRange(fromEpoch, toEpoch));

and 

> require(toEpoch >= activationEpoch && toEpoch <= block.number, Errors.InvalidEpochRange(fromEpoch, toEpoch));

Neither of these are telling me that our settlement logic needs to understand anything about proving periods.  Just that we can't settle empty intervals of epochs, can't settle epochs into the future and can't settle before we've activated.

The other (reasonable) concern you are raising is:

> Curio must still validate rails before we try to settle. We cannot keep wasting gas on settling things which gives no money to SP.

How much gas are we wasting here? If we just extended the settlement frequency out another 6 hours couldn't we reduce a lot of waste without introducing the complexity of determining and encoding rules about rail types?

LexLuthr:
> @LexLuthr I've spent a lot of time learning about all of the recent changes. Overall I am not seeing the problem warranting this specific change.
> 
> This is very cleanly stated in the PR description and looks like the main problem:
> 
> > When the target falls inside an open proving period or outside the validator’s accepted range, gas estimation can fail with InvalidEpochRange even though the rail still needs settlement.
> 
> But when I look at the service contract `validatePayment` function I see only two places raising this error
> 
> > require(totalEpochsRequested > 0, Errors.InvalidEpochRange(fromEpoch, toEpoch));
> 
> and
> 
> > require(toEpoch >= activationEpoch && toEpoch <= block.number, Errors.InvalidEpochRange(fromEpoch, toEpoch));
> 
> Neither of these are telling me that our settlement logic needs to understand anything about proving periods. Just that we can't settle empty intervals of epochs, can't settle epochs into the future and can't settle before we've activated.
> 
> The other (reasonable) concern you are raising is:
> 
> > Curio must still validate rails before we try to settle. We cannot keep wasting gas on settling things which gives no money to SP.
> 
> How much gas are we wasting here? If we just extended the settlement frequency out another 6 hours couldn't we reduce a lot of waste without introducing the complexity of determining and encoding rules about rail types?

I think the key distinction is that the contract fixes remove the activation/pre-activation validation bug, but they do not remove Curio’s need to choose a settlement target that matches FWSS semantics.

FWSS validation is still proving-period based: proven closed periods pay, closed unproven periods can advance with zero payment, and open periods should not advance because a proof can still land. So targeting `currentEpoch` is still often targeting inside the currently open proving period. Extending the settlement interval reduces frequency, but it does not make `currentEpoch` a valid semantic target.

On gas: with 10,000s of rails, “try it and see” is unfair to SPs because they pay for every successful on-chain no-op or low-value settlement. Spending a few extra lines of Curio code to avoid those sends is justified if it saves operator funds. Reducing cost not matter how insignificant has been one of the primary motivators for Curio since inception.

The goal here is to keep generic Filecoin Pay settlement operator/validator agnostic, while allowing FWSS to provide FWSS-specific target selection. That also matches the direction in #1356.

I fixed the termination-boundary issue: a terminated rail with `settledUpTo == endEpoch` may still need one final `settleRail()` call to zero/finalize the rail. The resolver now allows that equality target through instead of treating it as no progress. Once the rail is actually finalized, `GetRail` returns `RailInactiveOrSettled`, which we already treat as terminal.

lib/filecoinpayment/error_detection.go
LexLuthr: These are from ABI. Same as other error in that file.

LexLuthr: Or did I misunderstood your concern?

pdp/contract/FWSS/settle_resolver.go
LexLuthr: It was originally FWSSSettleResolver. It was feeling a bit mouthful and since function is a callback it kind of read like package + function = `FWSS.FWSSSettleResolver` so I changed it to `FWSS.SettleTargetResolver`. I can change back if this is important. I will wait for your final thoughts on the matter.

lib/filecoinpayment/error_detection.go
ZenGround0: I only meant grabbing the strings as constants rather than creating them here.

pdp/contract/FWSS/settle_resolver.go
ZenGround0: Oh ok `FWSS.SettleTargetResolver` is actually better

ZenGround0 (approved):
I still don't like it but you obviously care a lot and I trust your judgement so its approved for merge.

PR #1347, opened 2026-07-08
Add Apprise alerting integration

beck-8:
## Summary
- Adds a new `Apprise` alert plugin (`alertmanager/plugin/apprise.go`) alongside the existing Slack/PagerDuty/Prometheus AlertManager plugins, so a single curio config can fan alerts out to nearly any notification service (Telegram, Discord, Matrix, ntfy, email, SMS gateways, etc.) via a self-hosted [Apprise API](https://github.com/caronc/apprise-api) server.
- Supports both Apprise's stateless `/notify` endpoint (destination URLs listed in curio config via `NotifyURLs`) and its stateful `/notify/<key>` endpoint (destinations pre-configured on the Apprise server, curio just posts body/title/type).
- New config: `[Alerting.Apprise]` with `Enable`, `APIURL`, `NotifyURLs`, `Tag` — registered in `LoadAlertPlugins`.
- Regenerated `deps/config/doc_gen.go` and `documentation/en/configuration/default-curio-configuration.md` to include the new config section.

## Test plan
- [x] `go build ./alertmanager/... ./deps/config/...`
- [x] `gofmt -l` clean on touched files
- [x] Verified outgoing JSON payload (`title`/`body`/`type`/`format`/`urls`/`tag`, `Content-Type: application/json`) against the [apprise-api](https://github.com/caronc/apprise-api) API docs, including correcting a bug where HTTP 204 (no persisted config for the given key) was originally treated as a success instead of an error
- [x] End-to-end test against a running apprise-api instance with a real destination (Telegram/Discord) — in progress by reporter

### DIFF
--- a/alertmanager/plugin/apprise.go
+++ b/alertmanager/plugin/apprise.go
@@ -23,20 +23,6 @@ func NewApprise(cfg config.AppriseConfig) Plugin {
 	return &Apprise{cfg: cfg}
 }
 
-// appriseNotifyType maps a curio alert severity onto one of Apprise's notification types.
-func appriseNotifyType(severity string) string {
-	switch strings.ToLower(severity) {
-	case "critical", "error":
-		return "failure"
-	case "warning":
-		return "warning"
-	case "info":
-		return "info"
-	default:
-		return "warning"
-	}
-}
-
 // SendAlert posts to an Apprise API server (https://github.com/caronc/apprise-api), either its
 // stateless /notify endpoint (NotifyURLs required) or a stateful /notify/<key> endpoint (NotifyURLs
 // left empty).
@@ -53,7 +39,7 @@ func (p *Apprise) SendAlert(data *AlertPayload) error {
 	payload := map[string]any{
 		"title":  data.Summary,
 		"body":   strings.TrimSpace(body.String()),
-		"type":   appriseNotifyType(data.Severity),
+		"type":   "failure",
 		"format": "markdown",
 	}
 	if len(p.cfg.NotifyURLs) > 0 {
@@ -68,7 +54,7 @@ func (p *Apprise) SendAlert(data *AlertPayload) error {
 		return xerrors.Errorf("error marshaling JSON: %w", err)
 	}
 
-	req, err := http.NewRequest("POST", p.cfg.APIURL, bytes.NewBuffer(jsonData))
+	req, err := http.NewRequest("POST", p.cfg.URL, bytes.NewBuffer(jsonData))
 	if err != nil {
 		return fmt.Errorf("error creating request: %w", err)
 	}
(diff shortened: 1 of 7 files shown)


beck-8:
image
image

deps/config/types.go
```diff
@@ -752,6 +755,23 @@ type SlackWebhookConfig struct {
 	WebHookURL string
 }
 
+type AppriseConfig struct {
+	// Enable is a flag to enable or disable the Apprise integration.
+	Enable bool
```
LexLuthr: Can we get rid of this and use APIURL as enable/disable?

deps/config/types.go
```diff
@@ -752,6 +755,23 @@ type SlackWebhookConfig struct {
...
+
+	// APIURL is the notify endpoint of a running Apprise API server (https://github.com/caronc/apprise-api).
+	// Either its stateless endpoint (e.g. "http://127.0.0.1:8000/notify", use with NotifyURLs) or a
+	// stateful, pre-configured endpoint (e.g. "http://127.0.0.1:8000/notify/curio", leave NotifyURLs empty).
+	APIURL string
```
LexLuthr:
```suggestion
	URL string
```

alertmanager/plugin/apprise.go
```diff
@@ -0,0 +1,110 @@
...
+
+	var body strings.Builder
+	for key, value := range data.Details {
+		fmt.Fprintf(&body, "**%s**\n%v\n\n", key, value)
+	}
+
+	payload := map[string]any{
+		"title":  data.Summary,
+		"body":   strings.TrimSpace(body.String()),
+		"type":   appriseNotifyType(data.Severity),
```
LexLuthr: Let's just hardcode this. Alert mechanism is not meant for anything except critical alerts. if plugin uses a different string for the same hardcode it.

LexLuthr (commented):
Great addition! It needs some minor changes/cleanup. Looks good.

LexLuthr (approved):
Approved. Please update end to end test result.

beck-8:
I forgot to update when posting the picture. Already finished this part
https://github.com/filecoin-project/curio/pull/1347#issuecomment-4914496830

beck-8:
Give me a moment, let me confirm something. I see a failure warning.

beck-8:
The problem found here is: Settlements_FilecoinPay A large number of alerts are generated at the same time, all placed under a single event. Some channels successfully push the alerts, while others do not (because the content is too large).

Tomorrow, I'll look at how to handle it elegantly, so the alert is accurate without causing too much interference.

There aren't major issues here. If you think merging this first, I can handle it later in PR without any issues.

LexLuthr:
Ahh! Yes, the classic POST limits. You can fix it in this PR.

beck-8:
testing

beck-8:
He might lose some ending information, but I think that's enough.
image
image

LexLuthr:
> He might lose some ending information, but I think that's enough. image image

@beck-8 You can break the data into multiple alerts same as what we do for slack in https://github.com/filecoin-project/curio/blob/main/alertmanager/plugin/slack_webhook.go

beck-8:
Hundreds of alerts were issued at the same time—I was undoubtedly a huge disaster. Slack's handling method also cannot solve this problem.

PR #1348, opened 2026-07-08
Make AlertSendTest actually exercise configured alert plugins

beck-8:
## Summary
- `AlertSendTest` (used by the Web UI's alert-manager "send test alert" button) previously just inserted a hardcoded row into `alert_history` with `sent_to_plugins=FALSE` and `sent_at` already set. It never reached any configured plugin (Slack/PagerDuty/Prometheus AlertManager/etc.), and could never be picked up later either, since the queued-alert dispatch query requires `sent_at IS NULL`. The UI correctly labeled it "Local Only", but there was no way to actually test delivery from that button.
- Now it loads the currently-enabled plugins (`plugin.LoadAlertPlugins(a.Deps.Cfg.Alerting)`) and sends a real test alert through each of them synchronously, returning an error describing which channel(s) (if any) failed to deliver.

## Test plan
- [x] `go build ./web/api/webrpc/...`
- [x] `gofmt -l` clean
- [x] Manual verification against a real configured channel - in progress by reporter

### DIFF
--- a/web/api/webrpc/alerts.go
+++ b/web/api/webrpc/alerts.go
@@ -122,11 +122,9 @@ func (a *WebRPC) AlertPendingCount(ctx context.Context) (int, error) {
 	return count, nil
 }
 
-// AlertSendTest sends a test alert through every currently-enabled alert plugin (Slack,
-// PagerDuty, Prometheus AlertManager, Apprise, ...) and records the outcome in alert_history so
-// it's immediately visible in the UI and sidebar. If no plugins are enabled, or all of them fail,
-// the test alert is still recorded (marked as not sent to plugins) and an error is returned
-// describing what failed.
+// AlertSendTest sends a test alert through every enabled alert plugin and records the outcome
+// in alert_history. With no plugins enabled it just records locally - an error is
+// only returned if a configured plugin fails to deliver.
 func (a *WebRPC) AlertSendTest(ctx context.Context) error {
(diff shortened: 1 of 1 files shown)


web/api/webrpc/alerts.go
```diff
@@ -120,16 +122,45 @@ func (a *WebRPC) AlertPendingCount(ctx context.Context) (int, error) {
...
-	`)
+		VALUES ('TestAlert', $1, 'web-ui', $2, NOW())
+	`, testMessage, sentToPlugins)
 	if err != nil {
 		return xerrors.Errorf("inserting test alert: %w", err)
 	}
+
+	if len(plugins) == 0 {
+		return xerrors.Errorf("test alert recorded, but no alert plugins are enabled - nothing was sent")
+	}
```
LexLuthr: This is overkill. What is user only want to test GUI?

LexLuthr (changes requested):
Earlier implementation was intentionally GUI specific. I like the extension but we should do it on best effort basis.

web/api/webrpc/alerts.go
beck-8: I configured the alarm medium, and I want to click test to actually send a test alert. It's not just the display on the interface.

LexLuthr (commented):
Please attach a screenshot of real world test from any plugin.

LexLuthr (approved):

beck-8:
image
image

Issue #1320, opened 2026-06-30
UI Overhaul


snadrus:
The Curio UI is a bunch of user-discoverable tools that are decently self explanatory and easy for automated translation to update. We want to keep the no-compile, minimal-dependency approach as this JS has the potential to drain SP funds. The following is a variety of changes that would help make it more consumable. 

Note, the `make curio-pdp` gets a reduced UI which needs to make sense for them. 

Appearance:
It should take on a "Stripe dashboard" look with component ideas from Linear. The amount of useful, dynamic information on the screen should be high. This means that "all these things are well" summaries make sense.

Menu groups:
How want info grouped. Note, many of these pages exist already, but need arranging. 

___Dashboards___
Graphs like "PDP Status", "Wallet Status", "PoREP Status", Dataflow status.
and Wizards (step-by-step walkthroughs with complete-ness verification).
Note, some of this data may need either poller tasks or OpenCensus time-series data collection. 
The priority should be on PDP as the PoREP people have been around for a while & likely have their own solutions.

___Alerts___
A dash showing the output of the alerts framework

___Detail___
DB-Schema-like: Each page show searchable tables with potential drilldowns and the most obvious buttons there. Ex: 
- Sectors, 
- PDP Datasets (which could drill down to pieces or health) 
- Configuration
- Storage location maintenance

___Tools___
Form based, transactional or kick-off tools that don't have another home. 
- Configuration
- Wallet setup
- CC schedule

PDP Getting Started: 
Making PDP easier for SPs is key. They should need only to understand Datacenter terminology, NOT Filecoin terminology. The following steps should auto-detect completeness and suggest ways the NEXT incomplete step should be accomplished (with links to docs, relevant tools, or similar):
- Connect to Chain
- Connect wallet as PDP wallet
- Fund wallet: suggest links to Binance, etc that point to funding that configured wallet
- Connect Storage Location(s)
- Add DNS to config
- Verify reachability via DNS: suggest CloudflareD tunnels
- Verify PDP flow (this should be a Curio tool)
- Register as an FOC provider: Outlink?

beck-8:
> Verify reachability via DNS: suggest CloudflareD tunnels

This is not feasible; CF has an upload limit of 100M.

LexLuthr:
I don't like the CF idea either. 
I think we missed an important invariant here. No react or other frameworks. We don't want a JS dependency nightmare. We use the existing LitElement style JS.

[...]

PR #1351, opened 2026-07-10
Style the UI to work better for PDP

snadrus:
Based off the ideas around the mock https://beck-8.github.io/pdp-sp-console/mock/#/dashboard
But not where it cannot drive an action or cause new continuous load. 

To develop properly, ReadOnly mode was added. 

Overview:
image

PDP Guide:
image

image


snadrus:
Yeah that visual regression test is way off now. I'm not sure it's helping anything

LexLuthr:
@beck-8 Hey! Can you please take the roll of active reviewer for this? You have insight as SP, TSE and engineer.

beck-8:
@LexLuthr  I think I have no problem, but I look at the current situation more from the user's perspective. I am not professional in coding

snadrus:
I'm keeping 3rd party libraries and risky changes to a minimum here. I need a user perspective most (this is UI work).

ZenGround0:
cc @TippyFlitsUK if you have time to run the UI from this branch your feedback would also be useful.

cmd/skiff/main.go
```diff
@@ -101,6 +107,13 @@ func main() {
 			}
 			return c.Set(curiodeps.FlagRepoPath, p)
 		},
+		Commands: []*cli.Command{
+			{
+				Name:  "gui",
```
ZenGround0: I'm confused on whether this does something different than the main binary.  I am not seeing different defaults getting set and the Action function is the same

pdpnode/deps.go
```diff
@@ -85,6 +85,7 @@ func Open(ctx context.Context, cctx *cli.Context) (*Deps, error) {
 		return nil, xerrors.Errorf("load config: %w", err)
 	}
 
+	applySkiffDefaults(cfg)
```
ZenGround0, line 89: I'm failing to build on this line

```
pdpnode/deps.go:88:2: undefined: applySkiffDefaults
```

ZenGround0: Im in a clean repo on `6bd590ca` and CI looks incomplete too so I think this is a problem with the branch not my setup.

ZenGround0:
@snadrus I think this closes #1320?  Is that correct?

snadrus:
Tippy reviewed this live.
I need to fix what got broken in the PoRep side before closing that Issue.

beck-8:
other: I'm a bit worried about this `SQL Console`. He really helps us perform many queries very conveniently.
But I think this might be more vulnerable, and some SPs might intentionally or unintentionally expose 4701 to the public network, which would be a huge disaster.

I want to implement some authorization policies here; by default, there is no permission to access this.

snadrus:
> other: I'm a bit worried about this `SQL Console`. He really helps us perform many queries very conveniently.
> But I think this might be more vulnerable, and some SPs might intentionally or unintentionally expose 4701 to the public network, which would be a huge disaster.
> 
> I want to implement some authorization policies here; by default, there is no permission to access this.

I'm thinking about setting the session characteristics to read only. That'll make it an analysis tool only. It'll still be possible to steal the lotus admin key, but without that particular machines' API access it won't do much good.

FilOzzy:
From sync today, @TippyFlitsUK will do a UI-review

web/api/webrpc/pdp_guide.go
```diff
@@ -0,0 +1,455 @@
...
+	}
+
+	tokenPath := filepath.Join(dir, "tunnel.token")
+	if err := os.WriteFile(tokenPath, []byte(token+"\n"), 0o600); err != nil {
+		return "", fmt.Errorf("writing tunnel token: %w", err)
+	}
+
+	if err := startCloudflaredTunnel(bin, token, dir); err != nil {
+		return "", err
+	}
```
LexLuthr: This is serious attack vector and should not be done at all. It feels very sneaky to me. Any download should be with explicit user's permission.

web/api/webrpc/pdp_pipeline.go
```diff
@@ -0,0 +1,709 @@
...
+
+			CommpTaskID    NullInt64 `db:"commp_task_id"`
+			AggrTaskID     NullInt64 `db:"agg_task_id"`
+			AddPieceTaskID NullInt64 `db:"add_piece_task_id"`
+			SaveCacheTask  NullInt64 `db:"save_cache_task"`
+			IndexingTaskID NullInt64 `db:"indexing_task_id"`
+		}
+
+		err = tx.Select(&pipelines, `SELECT piece_ref, sector, commp_task_id, agg_task_id, indexing_task_id
+			FROM market_mk20_pipeline WHERE id = $1`, id)
```
LexLuthr: Wrong table.


```suggestion
			FROM pdp_pipeline WHERE id = $1`, id)
```

web/static/pages/pdp/pipeline.mjs
LexLuthr: Can you please fix these? This is my mistake from Mk20 code. Sorry!

web/api/webrpc/chain_status.go
```diff
@@ -0,0 +1,89 @@
...
+		return "unreachable"
+	}
+	if allOk {
+		return "ok"
+	}
+	if worst == "unreachable" && reachable > 0 {
+		return "degraded"
+	}
+	return worst
+}
```
LexLuthr, line 60: Unreachable nodes set worst = "unreachable" but do not set allOk = false. If one node is unreachable and another reports ok, this returns ok

```suggestion
func rollupSyncStatus(states []RpcInfo) string {
	if len(states) == 0 {
		return "unknown"
	}

	reachable := 0
	anyUnreachable := false
	worst := "ok"

	for _, s := range states {
		if !s.Reachable {
			anyUnreachable = true
			continue
		}

		reachable++
		syncState := strings.ToLower(s.SyncState)

		switch {
		case syncState == "ok":
		case strings.HasPrefix(syncState, "behind"):
			worst = "behind"
		case strings.HasPrefix(syncState, "slow") && worst == "ok":
			worst = "slow"
		case worst == "ok":
			worst = "degraded"
		}
	}

	if reachable == 0 {
		return "unreachable"
	}
	if worst != "ok" {
		return worst
	}
	if anyUnreachable {
		return "degraded"
	}
	return "ok"
}
```

web/api/webrpc/pdp_guide.go
```diff
@@ -0,0 +1,455 @@
...
+type PDPGuideWalletStatus struct {
+	OK          bool   `json:"ok"`
+	Configured  bool   `json:"configured"`
+	Address     string `json:"address,omitempty"`
+	FilAddress  string `json:"filAddress,omitempty"`
+	Balance     string `json:"balance,omitempty"`
+	Funded      bool   `json:"funded"`
+	ActorExists bool   `json:"actorExists"`
+	Detail      string `json:"detail,omitempty"`
+}
```
LexLuthr, line 41: Unknown balance is not returned here so might have glitchy GUI if there is an RPC error.

web/api/webrpc/pdp.go
LexLuthr: Another bug from merge to main. This should use correct address.

web/api/webrpc/pdp_guide.go
```diff
@@ -0,0 +1,455 @@
...
+		if strings.Contains(err.Error(), "no PDP key") {
+			return PDPGuideRegistryStatus{Detail: "Register after configuring a funded PDP wallet and public DNS."}
+		}
+		return PDPGuideRegistryStatus{Detail: err.Error()}
+	}
+	if status == nil {
+		return PDPGuideRegistryStatus{Detail: "Not registered with the Filecoin Onchain Cloud service provider registry."}
+	}
+	detail := fmt.Sprintf("Registered as %q (provider id %d).", status.Name, status.ID)
+	if !status.Active {
```
LexLuthr, line 205: This is a failure mode and should be handled such.

web/api/webrpc/pdp_guide.go
```diff
@@ -0,0 +1,455 @@
...
+	if err != nil {
+		return false, err.Error()
+	}
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return false, err.Error()
+	}
+	defer func() { _ = resp.Body.Close() }()
+	_, _ = io.Copy(io.Discard, io.LimitReader(resp.Body, 1024))
+	if resp.StatusCode >= 200 && resp.StatusCode < 500 {
```
LexLuthr: Why not check 200?

LexLuthr: Current range is happy with a 404 which can come from an incorrectly configured proxy.

web/api/webrpc/pdp_dashboard.go
```diff
@@ -0,0 +1,415 @@
...
+	return new(big.Int).Div(sum, big.NewInt(int64(ok))), nil
+}
+
+func (a *WebRPC) pdpIncome30d(ctx context.Context) (settled *big.Int, accrued *big.Int, err error) {
+	settled = big.NewInt(0)
+	accrued = big.NewInt(0)
+
+	paymentAddr, err := filecoinpayment.PaymentContractAddress()
+	if err != nil {
+		return settled, accrued, nil
```
LexLuthr:
```suggestion
		return settled, accrued, err
```

web/static/pages/pdp-guide/guide.mjs
```diff
@@ -0,0 +1,663 @@
...
+                            <input id="reg-name" .value=${this.registerName}
+                                   @input=${(e) => { this.registerName = e.target.value; }} required>
+                            <label for="reg-desc">Description</label>
+                            <input id="reg-desc" .value=${this.registerDescription}
+                                   @input=${(e) => { this.registerDescription = e.target.value; }} required>
+                            <label for="reg-loc">Location</label>
+                            <input id="reg-loc" .value=${this.registerLocation}
+                                   @input=${(e) => { this.registerLocation = e.target.value; }}
+                                   placeholder="C=US;ST=California;L=San Francisco" required>
+                            <p class="hint">Location format: <span class="mono">C=US;ST=California;L=San Francisco</span></p>
```
LexLuthr, line 530: These inputs should be validated on JS side before we Go rejects them.

LexLuthr (commented):
I have done 1st pass and found a few issues. Couple of them are older than this PR but will have an impact so I flagged them as well.

ZenGround0 (approved):
Given all the manual review done in calls and by tippy and beck I think we can proceed with merging once @LexLuthr's comments are addressed.

beck-8:
I simplified some SP registration information in section 1344. What will we do here? (I'm not sure if this has been changed; I didn't notice it. I'll check again tomorrow.)

snadrus:
> I simplified some SP registration information in section 1344. What will we do here? (I'm not sure if this has been changed; I didn't notice it. I'll check again tomorrow.)

I don't see any conflict.

LexLuthr (approved):

ZenGround0:
Are we blocking only on getting #1355 approved or something else?  Can we merge this to main now and rebase #1355 to main to unblock?

LexLuthr (approved):

PR #1355, opened 2026-07-13
UI: PDProve, DataSets, PoRep

snadrus:
Lets clean-up PoRep UI experience. 
Also, a few PDP pages are missing: Proving & DataSets
Screenshot 2026-07-14 at 10 54 44 AM
image
Screenshot 2026-07-14 at 4 30 11 PM
PoRep (underpopulated):
image


ZenGround0 (commented):
Looking at CI builds still seem to be failing in the same way I ran into when trying to play with #1351 (https://github.com/filecoin-project/curio/pull/1351#discussion_r3573580623)

ZenGround0 (approved):
I have not looked extensively but I would like to unblock.

pdp/explore/static/explore.js
```diff
@@ -0,0 +1,286 @@
...
+    if (Number.isNaN(d.getTime())) return String(iso)
+    return d.toISOString().replace('T', ' ').replace(/\.\d+Z$/, ' UTC')
+  }
+
+  function revokeObjectUrls() {
+    for (const u of objectUrls) URL.revokeObjectURL(u)
+    objectUrls = []
+  }
+
+  function pieceUrl(cid) {
```
LexLuthr, line 59: This assumes /piece/{cid} is mounted on the same host, but PDP-only pdpnode does not mount retrieval.Router, so Preview/Download 404 on public PDP deployments. Please use the same https://github.com/filecoin-project/curio/blob/ce15c0c92209366a5523b803e9c159baa2ffb66a/cuhttp/server.go#L92 with additioanl switch for libp2p and http market path disabled when PDP only.

web/api/webrpc/pdp_proving.go
```diff
@@ -0,0 +1,382 @@
...
+		return out, xerrors.Errorf("net summary: %w", err)
+	}
+	out.HeadEpoch = net.Epoch
+	head := net.Epoch
+
+	var rows []struct {
+		ProveAtEpoch    sql.NullInt64 `db:"prove_at_epoch"`
+		ChallengeWindow sql.NullInt64 `db:"challenge_window"`
+	}
+	err = a.Deps.DB.Select(ctx, &rows, `
```
LexLuthr, line 46: This ignores next_prove_attempt_at; datasets in retry backoff keep stale prove_at_epoch, while schedulers skip them until backoff expires, so the overview can falsely report them as overdue.

web/api/webrpc/pdp_proving.go
```diff
@@ -0,0 +1,382 @@
...
+	ProveAtEpoch              *int64     `json:"proveAtEpoch,omitempty"`
+	ChallengeWindow           *int64     `json:"challengeWindow,omitempty"`
+	Reason                    string     `json:"reason"`
+}
+
+func (a *WebRPC) PDPProvingTimeline24h(ctx context.Context) ([]PDPProvingTimelineEvent, error) {
+	var events []PDPProvingTimelineEvent
+	// pdp_prove_tasks only exists while the task is in-flight (CASCADE on completion).
+	// For completed tasks, recover data_set from the error text when present.
+	err := a.Deps.DB.Select(ctx, &events, `
```
LexLuthr, line 166: Completed successful prove tasks cannot be mapped back to a dataset here: pdp_prove_tasks is deleted on completion and successful history rows have empty err, so normal successes become dataset ? or disappear.

web/api/webrpc/pdp_proving.go
```diff
@@ -0,0 +1,382 @@
...
+
+	type dsRow struct {
+		ID                        int64         `db:"id"`
+		ConsecutiveProveFailures  int           `db:"consecutive_prove_failures"`
+		NextProveAttemptAt        sql.NullInt64 `db:"next_prove_attempt_at"`
+		UnrecoverableFailureEpoch sql.NullInt64 `db:"unrecoverable_proving_failure_epoch"`
+		ProveAtEpoch              sql.NullInt64 `db:"prove_at_epoch"`
+		ChallengeWindow           sql.NullInt64 `db:"challenge_window"`
+	}
+	var unhealthy []dsRow
```
LexLuthr, line 214: This only emits failures for datasets already marked unhealthy, so recent failed prove history with a dataset id but no consecutive_prove_failures row is dropped instead of shown.

tasks/pdpv0/task_prove.go
```diff
@@ -453,7 +453,9 @@ func (p *ProveTask) Do(taskID harmonytask.TaskID, stillOwned func() bool) (done
 		if !comm {
 			return false, xerrors.Errorf("failed to commit transaction")
 		}
-		return true, nil
+		// Complete the task (no harmony retry) but retain the send error in task history
+		// as a non-failing error so the proving UI can surface what went wrong.
+		return true, sendErr
```
LexLuthr, line 456: https://github.com/filecoin-project/curio/pull/1363 will break this.

web/api/webrpc/pdp_datasets.go
```diff
@@ -0,0 +1,790 @@
...
+	if offset < 0 {
+		offset = 0
+	}
+	sortBy = normalizeDataSetSortBy(sortBy)
+
+	filter = strings.TrimSpace(filter)
+	net, _ := a.NetSummary(ctx)
+	head := net.Epoch
+
+	if looksLikeEthAddress(filter) {
```
LexLuthr, line 222: Wallet filtering reports the chain-global dataset total but displays only local DB rows, so pagination can show pages/counts that do not exist on this node.

pdp/explore_auth.go
```diff
@@ -0,0 +1,39 @@
...
+	"strings"
+)
+
+const exploreDataSetsPathPrefix = "/explore/data-sets/"
+
+// allowExploreList soft-gates the dataset explorer piece listing.
+// Membership is public on-chain and piece bytes are served at /piece/{cid};
+// this accepts either a valid service JWT or a Referer from the explorer UI
+// so the list endpoint is clearly not meant as a public catalog API.
+func (p *PDPService) allowExploreList(r *http.Request) bool {
```
LexLuthr, line 17: Clients can forge Referer: /explore/data-sets/1, and /explore is mounted outside the PDP throttle, so the piece-list API is publicly enumerable.

web/api/webrpc/pdp_datasets.go
```diff
@@ -0,0 +1,790 @@
...
+			UNION
+			SELECT LOWER(TRIM(BOTH FROM add_message_hash))
+			FROM (
+				SELECT DISTINCT add_message_hash
+				FROM pdp_data_set_piece_adds
+				WHERE data_set = $1 AND add_message_hash IS NOT NULL
+				LIMIT 15
+			) recent_adds
+		)
+		SELECT mse.signed_hash AS tx_hash,
```
LexLuthr, line 582: This uses message_sends_eth.send_success as the interaction result, but that only means broadcast success; on-chain success is message_waits_eth.tx_success, so reverted txs can show as ok.

LexLuthr (changes requested):
We are not mounting retrieval endpoints and are skipping badbit list as well.

LexLuthr (approved):
Approving to unblock

pdp/explore_auth.go
snadrus: i put it behind throttle to protect SP. Not interested in protecting dataset as it's public on FilFox.

web/api/webrpc/pdp_datasets.go
snadrus: stopped using .NetSummary for DB

pdp/explore/static/explore.js
snadrus: https://github.com/filecoin-project/curio/pull/1378 is the fix

web/api/webrpc/pdp_proving.go
snadrus: thanks! I didn't know anything about that one.

web/api/webrpc/pdp_proving.go
snadrus: ugh, to not make this too complicated, I reduced the "successful" proves to simply reflect task success. The proof could still fail on-chain AND it doesn't show WHICH datasets were successful since we lose that when we delete these rows. 

I'm still not entirely sure you're right though, because this would suggest that we don't know which datasource we proved, and I remember showing-off the proven datasets. 

In any case, it now just says 'prove: ok' for the healthy prove periods.

tasks/pdpv0/task_prove.go
snadrus: When a PR breaks these assumptions, that PR should fix (or remove) the UI it breaks. That note is there because it's a reminder that changes here affect UI.

PR #1358, opened 2026-07-14
Relax pull URL requirement

### LINKED ISSUE #1357: Relax pull url requirement
The /piece/{pieceCidV2} URL-shape requirement lives in one Curio function (ValidatePullSourceURL, pdp/pull_types.go:59-90).  We should relax this requirement as it is not providing anything value and is blocking ipfs2foc

ZenGround0:
Closes #1357

### DIFF
--- a/pdp/pull_types.go
+++ b/pdp/pull_types.go
@@ -4,10 +4,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"net"
 	"net/url"
 	"os"
-	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pdp/pull_types.go
```diff
@@ -67,18 +64,6 @@ func ValidatePullSourceURL(sourceURL string, expectedPieceCid string) error {
...
-
-	// Extract pieceCid from URL and compare with expected
-	urlPieceCid := matches[1]
-	if urlPieceCid != expectedPieceCid {
-		return fmt.Errorf("pieceCid in URL %q does not match expected %q", urlPieceCid, expectedPieceCid)
-	}
-
 	// Validate host is not a private/local address (skip in devnet mode)
 	if !pullAllowInsecure {
 		if err := validatePublicHost(parsed.Host); err != nil {
```
LexLuthr: I think we can drop this as well. ALl we need to validate is that it is a URL shape. https://github.com/filecoin-project/curio/blob/main/lib/robusthttp/ssrf.go should be our validator.

LexLuthr (commented):
We can probably cut a little deeper here.

pdp/pull_types.go
ZenGround0: sgtm

ZenGround0:
FYI @SgtPooki

LexLuthr (approved):

PR #1363, opened 2026-07-16
Handle PDPv0 proving contract reverts explicitly

LexLuthr:
## Summary

- Add explicit PDPv0 proving revert classification for FWSS and PDPVerifier errors.
- Handle known non-terminal proving errors without marking datasets unrecoverable.
- Keep generic/unknown contract reverts on an alert-and-retry path instead of local termination.
- Add focused coverage for revert classification and PDPv0 failure handling state changes.


tasks/pdpv0/error_detection.go
```diff
@@ -4,52 +4,149 @@ import (
...
+	ErrPDPVerifierDataSetNotLive             abi.Error
+	ErrPDPVerifierInsufficientChallengeDelay abi.Error
+
+	// Unexpected proving invariant errors. Curio should not produce these in
+	// normal PDPv0 initPP/nextPP/prove flow; classify them explicitly so they
+	// alert and require investigation instead of entering recovery/backoff paths.
+	ErrPDPVerifierExcessiveChallengeDelay abi.Error
+	ErrFWSSOnlyPDPVerifierAllowed         abi.Error
+	ErrFWSSInvalidChallengeCount          abi.Error
+)
```
ZenGround0, line 40: Nice error groupings

tasks/pdpv0/error_detection.go
```diff
@@ -4,52 +4,149 @@ import (
...
 	}
+	ErrFWSSOnlyPDPVerifierAllowed = onlyPDPVerifierAllowed
 
-	dataSetNotFound, ok := parsedPDPVerifier.Errors["DataSetNotFound"]
+	proofAlreadySubmitted, ok := parsedFWSS.Errors["ProofAlreadySubmitted"]
 	if !ok {
-		panic("PDPVerifier ABI missing DataSetNotFound error")
+		panic("FWSS ABI missing ProofAlreadySubmitted error")
+	}
+	ErrFWSSProofAlreadySubmitted = proofAlreadySubmitted
```
ZenGround0: Super nit: you could drop some lines by doing `var ok bool` at the top and then doing `ErrFWSSProofAlreadySubmited, ok = parsedFWSS.Errors["ProofAlreadySubmitted"]` etc

tasks/pdpv0/error_detection.go
```diff
@@ -60,8 +157,108 @@ func IsUnrecoverableError(err error) bool {
...
+	}
+	errStr := strings.ToLower(err.Error())
+	return strings.Contains(errStr, contractErrorSelector(ErrFWSSProofAlreadySubmitted)) ||
+		strings.Contains(errStr, contractErrorSelector(ErrFWSSProvingPeriodPassed)) ||
+		strings.Contains(errStr, contractErrorSelector(ErrFWSSNextProvingPeriodAlreadyCalled)) ||
+		strings.Contains(errStr, strings.ToLower(provingRevertNoChallengeScheduled))
+}
+
+// IsRefreshProvingStateError returns true when local/on-chain proving state
+// should be refreshed before trying the proving workflow again.
```
ZenGround0: Does "refresh" mean recompute the proof?

tasks/pdpv0/failure_handling.go
```diff
@@ -54,114 +35,212 @@ func MarkDatasetProvingUnrecoverable(tx *harmonydb.Tx, dataSetId int64, currentH
...
+}
 
-	// Apply exponential backoff
-	backoffBlocks := CalculateBackoffBlocks(newFailures)
-	nextAttempt := currentHeight + int64(backoffBlocks)
-
-	_, err = tx.Exec(`
-		UPDATE pdp_data_sets
-		SET consecutive_prove_failures = $2,
-			next_prove_attempt_at = $3
```
ZenGround0, line 99: If we are removing automatic deletion based on failure counts should we clean up the database schema to remove these fields?

tasks/pdpv0/failure_handling.go
```diff
@@ -2,42 +2,23 @@ package pdpv0
 
 import (
 	"context"
+	"math/big"
 
 	"golang.org/x/xerrors"
 
+	"github.com/filecoin-project/curio/alertmanager/curioalerting"
```
ZenGround0, line 8: The point of separating things into this file rather than keeping them in the individual tasks was that we had a generic framework with shared code.  I suggest removing all of the refactored task specific functions so they are colocated with their tasks.

ZenGround0: Anything generic that remains or that you introduce can remain here.

tasks/pdpv0/failure_handling.go
ZenGround0: So far I've only seen us setting consecutive_prove_failures to 0

tasks/pdpv0/failure_handling.go
```diff
@@ -54,114 +35,212 @@ func MarkDatasetProvingUnrecoverable(tx *harmonydb.Tx, dataSetId int64, currentH
...
+//   - Same-period timing errors, let harmony retry the current task
+//   - Other contract reverts, alert and retry without mutating proving state
+//   - Transient errors, return error for harmony retry
 //
 // Returns (err) where err==nil means the task should complete (not retry),
 // and err!=nil means harmony should retry the task.
-func HandleProvingSendError(tx *harmonydb.Tx, dataSetId int64, currentHeight int64, sendErr error) error {
-	// Tier 1: Known unrecoverable errors, mark as unrecoverable immediately
-	if IsUnrecoverableError(sendErr) {
+func HandleProvingSendError(ctx context.Context, tx *harmonydb.Tx, ethClient ethchain.EthClient, al curioalerting.AlertingInterface, dataSetId int64, currentHeight int64, sendErr error, source string) error {
```
ZenGround0: Related to moving non generic functions out of here I would like 2 seperate functions for next(init)pp and prove.  We don't want to be reading through this function with two distinct callers in our head and reading the comments to figure out which caller can hit which case.

tasks/pdpv0/failure_handling.go
```diff
@@ -54,114 +35,212 @@ func MarkDatasetProvingUnrecoverable(tx *harmonydb.Tx, dataSetId int64, currentH
...
+	case IsInsufficientChallengeDelayError(sendErr):
+		// Source: initPP/nextPP. The challenge epoch was too close to the
+		// current block. Retry the task so it recomputes challenge state and
+		// calldata instead of resending the same transaction.
+		// This should never happen and if it ever happens due to some extreme corner case,
+		// we should raise an alert
+		raiseAlert(al, sendErr, source)
+		log.Warnw("Retrying proving period scheduling after insufficient challenge delay",
+			"dataSetId", dataSetId, "source", source, "height", currentHeight, "error", sendErr)
+		return sendErr
```
ZenGround0: Is there any way to stop harmony task from retrying in this case?  I think that would be better behavior if we know the error can't be recovered with another try.

ZenGround0: We might want these error switcher functions to return a bool for retry to send back to the task's Do function

ZenGround0 (commented):
Got to say I'm a big fan of this approach.  The auto shutdown thing left me with a slightly bad feeling that has only gotten worse with time.  Leaving unapproved because I want to make sure I get another look after you refactor.  The diff will probably be pretty big if you agree with my comments, but the substance should be about the same.

tasks/pdpv0/failure_handling.go
ZenGround0: (init_pp can just call a function located in next_pp)

LexLuthr:
> Got to say I'm a big fan of this approach. The auto shutdown thing left me with a slightly bad feeling that has only gotten worse with time. Leaving unapproved because I want to make sure I get another look after you refactor. The diff will probably be pretty big if you agree with my comments, but the substance should be about the same.

Refactored as requested.

tasks/pdpv0/error_detection.go
ZenGround0: FYI I still like this ^ better but if you don't that's ok

tasks/pdpv0/task_next_pp.go
```diff
@@ -432,3 +405,203 @@ func (n *NextProvingPeriodTask) Adder(taskFunc harmonytask.AddTaskFunc) {
...
+			prev_challenge_request_epoch = NULL
+		WHERE id = $1
+	`, dataSetId)
+	if err != nil {
+		return xerrors.Errorf("failed to disable proving: %w", err)
+	}
+	return nil
+}
+
+var getPDPVerifierNextChallengeEpoch = func(ctx context.Context, ethClient ethchain.EthClient, dataSetId int64) (*big.Int, error) {
```
ZenGround0: Youve already got the proving schedule in both caller of preflight and send errors.  There's no reason to redundantly recalculate this, just pass int he proving schedule and call the right methods on it.

tasks/pdpv0/task_next_pp.go
```diff
@@ -432,3 +405,203 @@ func (n *NextProvingPeriodTask) Adder(taskFunc harmonytask.AddTaskFunc) {
...
+	}
+
+	challengeEpoch, err := pdpVerifier.GetNextChallengeEpoch(contract.EthCallOpts(ctx), big.NewInt(dataSetId))
+	if err != nil {
+		return nil, xerrors.Errorf("failed to get next challenge epoch: %w", err)
+	}
+	return challengeEpoch, nil
+}
+
+// skipCurrentOnChainProvingPeriod reconciles a dataset when initPP/nextPP learns
```
ZenGround0, line 353: I don't understand why we want to mess with the scheduling data if we have already scheduled a new challenge period.  I suspect most of this code can be removed.

tasks/pdpv0/task_next_pp.go
```diff
@@ -432,3 +405,203 @@ func (n *NextProvingPeriodTask) Adder(taskFunc harmonytask.AddTaskFunc) {
...
+		return xerrors.Errorf("next challenge epoch %s does not fit in int64 for data set %d", challengeEpoch.String(), dataSetId)
+	}
+
+	// prev_challenge_request_epoch normally records the epoch where Curio sent
+	// initPP/nextPP. In this recovery path that exact send epoch is not useful,
+	// because the local challenge_request_msg_hash is intentionally dropped; use
+	// the reconciliation height as a marker while prove_at_epoch drives scheduling.
+	affected, err := tx.Exec(`
+			UPDATE pdp_data_sets
+			SET challenge_request_msg_hash = NULL,
```
ZenGround0: This is not going to work as written because of how scheduling of the prove task requires challenge_request_msg_hash != NULL: https://github.com/filecoin-project/curio/blob/main/tasks/pdpv0/task_prove.go#L99

ZenGround0: If we just keep this the way that it is then it can make logs very noisy as next_pp will just keep getting scheduled and failing the entire subsequent proving period.

One approach to prevent this is to fix up next_pp scheduling to read the true value of prove_at from the chain and record so that at least next_pp scheduling won't noisily fail all proving period long and will recover correctly after the next proof has failed.

The other more correct thing is to handle the case where next_proving period message is not stored in curio in prove task scheduling.  I think all of this can be delegated to a follow on issue though

tasks/pdpv0/task_prove.go
```diff
@@ -989,3 +941,134 @@ var (
...
+
+func (p *ProveTask) disableProving(ctx context.Context, dataSetId int64) error {
+	// cleanup all proving related columns
+	// set init_ready to false so that next new piece enables proving
+	//
+	// This creates a bit of an edge case when piece deletions, additions and proving happen in the same time window:
+	// - data set gets used, pieces are added and proven
+	// - all pieces are deleted from it
+	// - nextProvingPeriod gets called on an empty dataset
+	// - a new piece gets added, it sets `init_ready = TRUE`, but it is true already
```
ZenGround0: But you are setting init_ready to FALSE here so this wouldn't be the case?

tasks/pdpv0/task_prove.go
```diff
@@ -989,3 +941,134 @@ var (
...
+	// - prove task fires, detects that proving set is empty, challenge epoch is 0 (as the proving set is empty),
+	// 		proving gets disabled
+	// Now the dataset won't get proven until one more piece gets added to set `init_ready = TRUE`.
+	// Better pattern here would be to react to events emitted in our messages from the transactions we send to PDPVerifier.
```
ZenGround0: I think you are saying this can still happen with the code here.  It sounds like a bug.  Please file an issue describing this so we can track it and close it down.

tasks/pdpv0/task_next_pp.go
ZenGround0: I guess to be clear what you are doing here will actually reduce the noise from next_pp.  What it won't do is allow for proving to be scheduled within a proving period where next_proving_period transaction is not stored in curio db.

ZenGround0 (approved):
This is pretty much good to go.  Even the `challenge_request_msg_hash = NULL,` within skipCurrentOnChainProvingPeriod shouldn't make things strictly worse.

tasks/pdpv0/task_prove.go
LexLuthr: Tracked in https://github.com/filecoin-project/curio/pull/1366, which addresses the empty-dataset/add race by reconciling confirmed proving-period effects before prove scheduling.

PR #1364, opened 2026-07-16
Fix: faster Piece HEAD & GET 

snadrus:
Our common-case TTFB is leaving lots of performance on the table. 

### THIS PR
**Disk improvements**
- Head reads just 512b (was 512 B + 1 MB)
- GET too, then no seek
- ReadAhead never supercedes range (start or end) nor 1MB

Bets with fallbacks:
 -  they gave a Parent CID
 -  the sector is on this machine

SQL Optimizations:
- The triple SQL RT of [deal, exists, parked_pieces] gets unified. 

Left Unoptimized:
- HEAD could avoid reads by serving 'application/octet-stream': browsers don't mind. Concern here: do we have custom clients that care?
- avoid rereading that 512b a second time for serving (it's likely in OS cache)
- a streaming form of the mime sniffer could read less, but the savings is below 1 disk sector read

### New Flow
1 YSQL for [deal+exists+parked_pieces]
\--> not_found then tries with YCQL to find parent & go again (1% non-PDP case)
Try-local (Disk)
\--> not_found does YSQL for find-sector (1x, not 2 because of try-local-first on other node).
open
sniff(512b) (return here for HEAD)
read only what's requested for RANGE, max 1MB.

### Previous Flow
Aggregate → (YCQL)
market_piece_deal (miss) → (YSQL)
pdp_piecerefs → (YSQL)
parked_pieces → (YSQL)
find-sector → (YSQL). (done 2x if remote)
open → (DISK/HTTP)
sniff → (DISK/HTTP) 
SEEK(0)
ReadAt for 1MB
first body byte.

### DIFF
--- a/lib/cachedreader/cachedreader.go
+++ b/lib/cachedreader/cachedreader.go
@@ -528,49 +528,53 @@ func (cpr *CachedPieceReader) GetSharedPieceReader(ctx context.Context, pieceCid
 		defer close(r.ready)
 
 		reader, size, err := cpr.getPieceReaderFromPDPPark(readerCtx, pieceCid)
-		if err != nil {
+		var finalErr error
+		switch {
+		case err == nil:
+			// PDP park hit
+		case errors.Is(err, errPDPParkNotFound):
 			log.Debugw("pdp park miss, trying aggregate", "piececid", pieceCid.String(), "err", err)
+			pdpMiss := err
 
 			reader, size, err = cpr.getPieceReaderFromAggregate(readerCtx, pieceCid, retrieval)
 			if err != nil {
 				log.Debugw("failed to get piece reader from aggregate", "piececid", pieceCid.String(), "err", err)
-
-				aerr := err
+				aggErr := err
 
 				reader, size, err = cpr.getPieceReaderFromMarketPieceDeal(readerCtx, pieceCid, retrieval)
 				if err != nil {
 					log.Debugw("failed to get piece reader", "piececid", pieceCid, "err", err)
-					finalErr := fmt.Errorf("failed to get piece reader from pdp park, aggregate, sector or piece park: %w, %w", aerr, err)
-
-					// Record error metric
-					_ = stats.RecordWithTags(context.Background(), []tag.Mutator{
-						tag.Upsert(reasonKey, "piece_not_found"),
-					}, CachedReaderMeasures.ReaderErrors.M(1))
-
-					// Cache the error in the error cache
-					cpr.pieceErrorCacheMu.Lock()
-					cpr.pieceErrorCache.Set(pieceCid, &cachedError{err: finalErr, pieceCid: pieceCid})
-					// Record error cache size
-					_ = stats.RecordWithTags(context.Background(), []tag.Mutator{
-						tag.Upsert(cacheTypeKey, "piece_error"),
-					}, CachedReaderMeasures.CacheSize.M(int64(cpr.pieceErrorCache.Count())))
-					cpr.pieceErrorCacheMu.Unlock()
-
-					// Remove the failed reader from the main cache
-					cpr.pieceReaderCacheMu.Lock()
-					cpr.pieceReaderCache.Remove(pieceCid)
-					// Record updated cache size
-					_ = stats.RecordWithTags(context.Background(), []tag.Mutator{
-						tag.Upsert(cacheTypeKey, "piece_reader"),
-					}, CachedReaderMeasures.CacheSize.M(int64(cpr.pieceReaderCache.Count())))
-					cpr.pieceReaderCacheMu.Unlock()
-
-					r.err = finalErr
-					readerCtxCancel()
-
-					return nil, 0, finalErr
+					finalErr = fmt.Errorf("failed to get piece reader from pdp park, aggregate, sector or piece park: %w, %w, %w", pdpMiss, aggErr, err)
 				}
 			}
+		default:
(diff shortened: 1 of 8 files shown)


lib/cachedreader/cachedreader.go
```diff
@@ -376,6 +376,71 @@ func (s SubPieceReader) ReadAt(p []byte, off int64) (n int, err error) {
 	return s.sr.ReadAt(p, off)
 }
 
+// errPDPParkNotFound means the piece is not a PDP park parent piece.
+// Callers should fall back to aggregate / market deal resolution.
+var errPDPParkNotFound = errors.New("piece not found in pdp piece park")
+
+// getPieceReaderFromPDPPark resolves parent pieces in one YSQL round-trip
+// (pdp_piecerefs ⋈ parked_piece_refs ⋈ parked_pieces). This is the common
```
ZenGround0, line 384: For my education how do you go about deciding whether a series of JOINs is going to be fast? ?  When I was last running into a query planner caused slow down it scared me away somewhat from JOINing a lot for performance reasons.

lib/cachedreader/cachedreader.go
```diff
@@ -456,50 +521,60 @@ func (cpr *CachedPieceReader) GetSharedPieceReader(ctx context.Context, pieceCid
 
 		cpr.pieceReaderCacheMu.Unlock()
 
-		// We just added a cached reader, so get its underlying piece reader
+		// We just added a cached reader, so get its underlying piece reader.
```
ZenGround0: nit: "We just..." can probably be removed.

lib/paths/local.go
```diff
@@ -244,6 +246,31 @@ func (p *path) sectorPath(sid abi.SectorID, fileType storiface.SectorFileType) s
...
 }
 
+// ExistingLocalFile returns a local filesystem path for sid/ft if the file
+// exists under any registered local storage path that allows ft. It does not
+// consult the sector index (no DB round-trip). Callers should fall back to
+// AcquireSector / StorageFindSector when this returns false.
+//
+// When multiple local roots are registered, Stats run sequentially on each path.
+// The random order should tend toward less disk activity than purely-parallel searching.
+func (st *Local) ExistingLocalFile(sid abi.SectorID, ft storiface.SectorFileType) (string, bool) {
```
ZenGround0, line 256: Is it possible that this could take longer than DB on a big filesystem ?  Should we allow for parallel searching between sector index and local ?

ZenGround0 (approved):
😎

lib/cachedreader/cachedreader.go
snadrus: Joins in Curio are always a column ${table name}_id to "id" (the literal) in the next table, which is always indexed, so (Check 1): 
- 1-to-1 rows will be basically free: O(n log n) streaming
- 1-to-N are dangerous ground

(Check 2)
It's important to have the "first" table produce an obviously small number of results for its size and WHERE filtering and to have that filter match an index. That's because Databases remember the likely hit count per item in an index (heuristics) and Planners try to find the fastest size collapse with 1 index (and ID indexes), or else a chain of *fancy* indexes, or sometimes multiple index collusion on 1 table (rare), or even just the whole table scan if it's tiny like harmony_task.


Ex: An easy case: 
So if you're joining a tiny row-count table to a bunch of others through simple foreign key joins, the planner will do fine. 

Ex: even easier case 
If it's quick to find 1 row and it's LIMIT 1 then you are done. 
My code here has a good collapse with an index that matches only some of the clauses, which is ok too, then limit 1 stops the search.

lib/paths/local.go
snadrus: With the directory certainly at-least partially in RAM, time goes to getting the few disk-sectors of the directory that need. If that's slower than a database roundtrip, I'd be surprised. 
As for parallelism, that has an effect on throughput. I was trying to improve velocity but not at the expense of throughput.

lib/paths/local.go
```diff
@@ -244,6 +246,31 @@ func (p *path) sectorPath(sid abi.SectorID, fileType storiface.SectorFileType) s
...
+	st.localLk.RLock()
+	for _, p := range st.paths {
+		if p.Local == "" || !ft.Allowed(p.AllowTypes, p.DenyTypes) {
+			continue
+		}
+		candidates = append(candidates, p.sectorPath(sid, ft))
+	}
+	st.localLk.RUnlock()
+	for _, spath := range candidates {
+		if fi, err := os.Stat(spath); err == nil && !fi.IsDir() {
```
magik6k, line 271: There can be 100s-1000s of those per machine in big deployments. Really most of this patch could be additional lru cache for metadata in r.local.AcquireSector, 90%+ of this PR is not really needed imo

LexLuthr: This would be really bad for PDP SPs. Much worse than PoRep actually. We are looking at millions of files on a single FS.

LexLuthr (commented):
So we starting serving bytes as soon as we have them with 0 checks?
All SQL checks remain.

lib/paths/local.go
snadrus: since this is a shortcut anyway, lets avoid all IO with:
if len(candidates) > 5 { return}

ZenGround0: Nice

snadrus:
TODO: Head should be simpler (no disk review).

LexLuthr (approved):

PR #1366, opened 2026-07-17
Fix: PDPv0 proving-period reconciliation for piece removals

LexLuthr:
Built on https://github.com/filecoin-project/curio/pull/1363

## Summary
• Adds a PDPv0 proving-period watcher that reconciles confirmed proving-period effects before prove scheduling runs.
• Moves piece-delete finalization out of the immediate nextPP send path and waits for the on-chain state to reflect removal.
• Handles empty-dataset proving periods by clearing stale local proving state and re-enabling init when pieces are present

### DIFF
--- a/tasks/pdpv0/task_next_pp.go
+++ b/tasks/pdpv0/task_next_pp.go
@@ -2,14 +2,12 @@ package pdpv0
 
 import (
 	"context"
-	"database/sql"
 	"errors"
 	"fmt"
 	"math/big"
 	"time"
 
 	"github.com/ethereum/go-ethereum/core/types"
-	"github.com/samber/lo"
 	"github.com/yugabyte/pgx/v5"
 	"golang.org/x/xerrors"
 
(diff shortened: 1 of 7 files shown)


tasks/pdpv0/watch_proving_period.go
```diff
@@ -0,0 +1,283 @@
...
+
+type pendingPieceDelete struct {
+	DataSetID int64          `db:"data_set"`
+	PieceID   int64          `db:"piece_id"`
+	TxHash    string         `db:"rm_message_hash"`
+	TxStatus  sql.NullString `db:"tx_status"`
+	TxSuccess sql.NullBool   `db:"tx_success"`
+}
+
+// processPendingPieceDeletes reconciles local piece-removal rows after the
```
ZenGround0: This phrasing is confusing.  This function reconciles local piece removal after nextProvingPeriod has removed the pieces on chain.

tasks/pdpv0/watch_proving_period.go
```diff
@@ -0,0 +1,283 @@
...
+)
+
+const alertNameProvingPeriod = "ProvingPeriod"
+
+// NewProvingPeriodWatcher reconciles confirmed proving-period side effects
+// before the prove watcher runs. nextProvingPeriod is what finally applies
+// scheduled piece removals and can also clear PDPVerifier's next challenge when
+// the dataset becomes empty, so both local states must be fixed before prove
+// task scheduling looks at the dataset.
+func NewProvingPeriodWatcher(w *Watcher) {
```
ZenGround0, line 32: There is something worse about this approach than trying to tie this to nextProvingPeriod.  We are now scanning all pieces scheduled for deletion on every tipset rather than lazily only running processing the pieces scheduled for deletion from a dataset that has just run nextProvingPeriod.

ZenGround0: In a world with many deletions this will be a lot less efficient.

tasks/pdpv0/task_next_pp.go
```diff
@@ -253,119 +251,12 @@ func (n *NextProvingPeriodTask) Do(taskID harmonytask.TaskID, stillOwned func()
 		return false, xerrors.Errorf("failed to perform database transaction: %w", err)
 	}
 
-	// For all `schedulePieceDeletions` messages relevant to this dataset, mark these pieces as removed
-	err = n.processPendingPieceDeletes(ctx, dataSetId)
```
ZenGround0, line 257: To make sure I am understanding the problem: doing it this way we are doing the removal processing before nextProvingPeriod lands.  So if it does not land for some reason then we are in a bad state?

tasks/pdpv0/watch_proving_period.go
```diff
@@ -0,0 +1,283 @@
...
+	DataSetID int64  `db:"id"`
+	TxHash    string `db:"challenge_request_msg_hash"`
+}
+
+// processEmptyProvingPeriods reconciles datasets whose confirmed initPP/nextPP
+// message left no next challenge on-chain. This happens when nextProvingPeriod
+// removes the final piece: PDPVerifier clears the challenge, but Curio may have
+// already stored the now-stale prove_at_epoch.
+//
+// Dropping that local period before piece-delete reconciliation is intentional.
```
ZenGround0: Is this comment paragraph describing the below function?  It seems like it is describing the function above and should be moved there.

ZenGround0 (approved):
A few comments but looks great.  Thanks for handling this.

tasks/pdpv0/watch_proving_period.go
ZenGround0: Perhaps we could run this once every proving period ?  Then we would be sure to run through the previous batch of deletions.  I don't think removal processing needs to happen on a quick deadline.

tasks/pdpv0/task_next_pp.go
LexLuthr: Yes. We are assuming not matter what it will not fail including a gas error. That seems way too optimistic and will produce bugs which will be very hard to figure out.

tasks/pdpv0/watch_proving_period.go
LexLuthr: Yes, I have been thinking about this. Scheduling is a little heavy handed here. Maybe watcher can be triggered by nextPP instead of always watching.

LexLuthr: Fixed

ZenGround0:
Did you rebase and squash?  In the future please keep new commits cleanly separate so I don't get lost in review.

LexLuthr:
> Did you rebase and squash? In the future please keep new commits cleanly separate so I don't get lost in review.

Only rebased because base PR was also changed based on the review. No squash. 
Although there was a conflict resolution so commit would have changed.

tasks/pdpv0/task_next_pp.go
```diff
@@ -309,7 +310,8 @@ func resetDatasetToInitPP(ctx context.Context, db *harmonydb.DB, dataSetId int64
              SET challenge_request_msg_hash = NULL,
                      prove_at_epoch = NULL,
                      init_ready = TRUE,
-                     prev_challenge_request_epoch = NULL
+                     prev_challenge_request_epoch = NULL,
+                     pp_reconcile_needed = FALSE
```
ZenGround0: I don't think we should be setting this FALSE in any of these error handling functions.  These functions area always called before the mwe db transaction which sets pp_reconcile_needed to TRUE.  So this set to FALSE should always be a NOOP.  A more comprehensible and minimal design which is just as correct is to only set this value to FALSE after processing in the watch proving period task

tasks/pdpv0/watch_proving_period.go
```diff
@@ -52,33 +95,68 @@ type confirmedProvingPeriod struct {
...
-// then decides whether InitProvingPeriodTask should pick the dataset back up.
-func processEmptyProvingPeriods(ctx context.Context, db *harmonydb.DB, ethClient ethchain.EthClient) error {
-	var confirmed []confirmedProvingPeriod
-	err := db.Select(ctx, &confirmed, `
+func selectProvingPeriodsNeedingReconcile(ctx context.Context, db *harmonydb.DB, limit int) ([]confirmedProvingPeriod, error) {
+	var periods []confirmedProvingPeriod
+	err := db.Select(ctx, &periods, `
 		SELECT pds.id,
-		       pds.challenge_request_msg_hash
+		       COALESCE(pds.challenge_request_msg_hash, '') AS challenge_request_msg_hash
```
ZenGround0: Why are you naming this with AS and not referring to it anywhere later in the query?

tasks/pdpv0/watch_proving_period.go
```diff
@@ -52,33 +95,68 @@ type confirmedProvingPeriod struct {
...
 		FROM pdp_data_sets pds
-		INNER JOIN message_waits_eth mwe ON mwe.signed_tx_hash = pds.challenge_request_msg_hash
-		WHERE pds.challenge_request_msg_hash IS NOT NULL
-		  AND mwe.tx_status = 'confirmed'
-		  AND mwe.tx_success = TRUE
+		LEFT JOIN message_waits_eth mwe ON mwe.signed_tx_hash = pds.challenge_request_msg_hash
+		WHERE pds.pp_reconcile_needed = TRUE
 		  AND pds.unrecoverable_proving_failure_epoch IS NULL
+		  AND (
+		      pds.challenge_request_msg_hash IS NULL
```
ZenGround0: I don't understand why this piece of the query is there.  Isn't the challenge_request_msg_hash being NULL case never matched by the left join above?

ZenGround0 (approved):
Approved conditioned on addressing comments, particularly the first one

tasks/pdpv0/watch_proving_period.go
LexLuthr: AS gives it the name which matches the Select function's expectation i.e. &periods. This basically allows us to pass empty string as column value instead of NULL.

LexLuthr: I will switch to sql.NullString to make this more obvious

tasks/pdpv0/task_next_pp.go
LexLuthr: Make sense. I will remove the unnecessary FALSE.

tasks/pdpv0/watch_proving_period.go
LexLuthr: LEFT JOIN does not help for the NULL-hash case because there is no message_waits_eth row to join to. When pds.challenge_request_msg_hash IS NULL, the join condition:
```sql
mwe.signed_tx_hash = pds.challenge_request_msg_hash
```
cannot match anything. The mwe.* columns will be NULL, so a condition like:
```sql
mwe.tx_status = 'confirmed' AND mwe.tx_success = TRUE
```
will not pass.
That is why the NULL-hash case has to be handled separately from the confirmed-message case. It represents a dataset where the watcher already cleared the local message hash during empty-period reconciliation, but pp_reconcile_needed is still true because later reconciliation work did not finish.

Query was still inefficient as planner kept using mwe as main table so I am not forcing it use the shape I want.

Issue #1362, opened 2026-07-16
pdp: addPieces returns 500 on duplicate pdp_data_set_piece_adds PK (SQLSTATE 23505) after the on-chain add already succeeded


TippyFlitsUK:
### Checklist

- [x] This is **not** a question or a support request.
- [x] This is **not** a new feature request.
- [x] This is **not** an enhancement request.
- [x] I **have** searched the issue tracker and discussion forum. Related: #805 (_pdpv0: make add pieces and create-and-add idempotent_), but no existing issue reports this specific runtime failure and the resulting on-chain vs. node-local divergence.
- [x] Running a released/RC build (see version below; observed on a third-party node).

### Curio component

Other (PDP / pdpv0)

### Curio Version

```
curio version 1.28.2-rc2+calibnet+git_31073fa3_2026-06-11T11:05:54-05:00
```

(Observed on a third-party node: **infrafolio-calib**, provider ID 4, calibnet. Version taken from the node's own `curio --version` heartbeat in its logs, running continuously through the time of the incident.)

### Describe the Bug

A client `addPieces` request produced a **successful on-chain add** but returned **HTTP 500** to the client, with a duplicate-key DB error on the node.

**Context**
- Network: **calibnet**
- SP: **infrafolio-calib** (provider ID 4), `https://caliberation-pdp.infrafolio.com`
- Dataset: **19497**, source `filecoin-pin`, 1 piece added (piece id 0, ~90.78 MB)

**On-chain result** (verified via chain + FOC observer indexer):
- tx `0xb99558e9573560f8f2165e7479dccc64ec24f048ca71dce435a059d1772e1aab`
- `addPieces` (selector `0x9afd37f2`) → PDPVerifier `0x85e366Cf9DD2c0aE37E963d9556F5f4718d6417C`
- Block **3894299**, **2026-07-15 22:42:30 UTC**, **Exit 0**
- Piece added to set 19497; dataset live; **no fault** recorded

**Node behaviour**
1. Within the same request, Curio sends the on-chain tx first (`sender_eth.go:410`, `send_error: null`), then inserts the local tracking row (`handlers_add.go:430`).
2. The tracking insert fails: `duplicate key value violates unique constraint "pdp_data_set_piece_adds_pk" (SQLSTATE 23505)` (`handlers_add.go:480`).
3. The handler returns HTTP `500` to the client (`handlers.go:42`, `eid=a6182165-2783-4c86-9213-aea080630963`).

[...]

PR #1367, opened 2026-07-17
fix(pdpv0): restore sub-piece component in piece_adds primary key

SgtPooki:
## What changed

`pdp_data_set_piece_adds` stores one row per sub-piece, but its primary key `(add_message_hash, add_message_index)` only distinguishes pieces. Any piece with 2+ sub-pieces violates the PK on its own second row (SQLSTATE 23505), after the AddPieces tx was already broadcast. The client gets a 500 and the piece lands on chain with no local tracking: never proven, never indexed. Full root-cause writeup in #1362.

- New migration restores a sub-piece component: PK `(add_message_hash, add_message_index, sub_piece_offset)`. `data_set` stays out of the key; it is nullable since 20251015 for create-and-add.
- `handleGetPieceAdditionStatus` compared confirmed piece ids (distinct pieces) against sub-piece row count, so a confirmed multi-sub-piece add would 500 on status polling once its rows can exist. Now compares distinct pieces.
- Tests: `TestInsertPieceAdds_MultiSubPieceAggregate` reproduces the collision (fails on main with the exact error from #1362), `TestInsertPieceAdds_MultiPieceBatch` locks in the batch shape that already works.

## How to verify

```
go test ./pdp/ -run TestInsertPieceAdds -v -count=1
```

On main, the aggregate test fails with `duplicate key value violates unique constraint "pdp_data_set_piece_adds_pk" (SQLSTATE 23505)`. With this branch both pass; `./pdp/` (131) and `./tasks/pdpv0/` (48) also pass.

## Notes / risks

- 20251010 replaced the original `(data_set, add_message_hash, sub_piece_offset)` key because multi-piece batches collide on it (every piece's sub-piece sits at offset 0). The new key keeps both discriminators, so both shapes insert cleanly.
- Validated on Postgres 14. Not tested on Yugabyte; the 20260414 lsm-shaped PK is dropped by the same constraint name, but a YB run would be good before merge.
- This prevents new orphans only. Pieces already stranded on chain with no local rows still need backfill or on-chain removal.

### DIFF
--- a/harmony/harmonydb/downgrade/20260717-pdp-v0-piece-adds-subpiece-pk.sql
+++ b/harmony/harmonydb/downgrade/20260717-pdp-v0-piece-adds-subpiece-pk.sql
@@ -1,0 +1,19 @@
+-- Revert the piece_adds primary key to (add_message_hash, add_message_index).
+-- Note: this fails if the table holds a piece with multiple sub-pieces (rows
+-- that are only distinguished by sub_piece_offset); remove those rows first.
+DO $$
+BEGIN
+  IF EXISTS (
+    SELECT 1
+    FROM pg_constraint
+    WHERE conname = 'pdp_data_set_piece_adds_pk'
+      AND conrelid = to_regclass('pdp_data_set_piece_adds')
+  ) THEN
+    ALTER TABLE pdp_data_set_piece_adds
+      DROP CONSTRAINT pdp_data_set_piece_adds_pk;
+  END IF;
+
+  ALTER TABLE pdp_data_set_piece_adds
+    ADD CONSTRAINT pdp_data_set_piece_adds_pk
+    PRIMARY KEY (add_message_hash, add_message_index);
+END $$;
--- a/harmony/harmonydb/sql/20260717-pdp-v0-piece-adds-subpiece-pk.sql
+++ b/harmony/harmonydb/sql/20260717-pdp-v0-piece-adds-subpiece-pk.sql
@@ -1,0 +1,28 @@
+-- pdp_data_set_piece_adds holds one row per SUB-piece, but the primary key
+-- (add_message_hash, add_message_index) only distinguishes pieces: every
+-- sub-piece row of an aggregated piece shares the piece's add_message_index,
+-- so adding a piece with two or more sub-pieces violates the PK on the second
+-- row. The handler has already broadcast the AddPieces transaction at that
+-- point, so the insert failure strands an on-chain piece with no local
+-- tracking (never proven, never indexed).
+--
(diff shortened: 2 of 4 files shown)


SgtPooki:
Live validation: this branch is deployed on calib2.ezpdpz.net (calibnet provider 2), and the failing shape now completes end to end.

Yesterday on the same node, pre-patch ([data set 19954](https://calib2.ezpdpz.net/pdp/data-sets/19954)): a 2-sub-piece aggregate add returned 500 with this PR's target error, the tx landed on chain anyway ([0xd899a7b3](https://calibration.filfox.info/en/tx/0xd899a7b34870e6f740be82c1528e2dcb858d0884f07c19f95658fdb9c93e9132), piece id 0), and [the add-status endpoint for that tx](https://calib2.ezpdpz.net/pdp/data-sets/19954/pieces/added/0xd899a7b34870e6f740be82c1528e2dcb858d0884f07c19f95658fdb9c93e9132) still returns "Piece addition not found". The piece is absent from the node's piece list.

Today, post-patch ([data set 19976](https://calib2.ezpdpz.net/pdp/data-sets/19976), same aggregate root): add returned 201 ([0x3dbc214d](https://calibration.filfox.info/en/tx/0x3dbc214d5a8f814178fc54238bc46ef72079a3f0a4cd8b4e82a53e8676a95477), piece id 0). Status polling returned `txStatus=confirmed`, `addMessageOk`, `piecesAdded`, `confirmedPieceIds=[0]`.. with 2 sub-piece rows and 1 confirmed piece id, the pre-patch comparison at `handlers.go:910` would have 500'd here, so both fixes in this PR were exercised. The [19976 piece list](https://calib2.ezpdpz.net/pdp/data-sets/19976) shows piece 0 with both sub-pieces at offsets 0 and 131072 ([explorer view](https://pdp.vxb.ai/calibration/dataset/19976)), and IPNI lists calib2 as provider for the source IPFS CIDs ([bafybeia222di..](https://filecoinpin.contact/cid/bafybeia222di4kgrows5yijrxhfncqbzqa43k2dmbxradelbdg4lt7alzm), [bafybeia2222m..](https://filecoinpin.contact/cid/bafybeia2222m75lxmc3ex7g3whuwwua76c2atzsi6uwkonnwrbm75loapa)); the first only exists on chain inside this aggregate, so that announce came from this add's indexing.

LexLuthr:
I’m converting this PR back to draft. The impact extends beyond the scope of this PR, as the proposed changes introduce a significant design shift and revise several existing assumptions. The Curio team needs time to assess the broader implications and align on how to proceed with this sub-piece approach before the PR is ready for further review.

SgtPooki:
@LexLuthr i'm going to close this just to remove it from your queue. feel free to reopen

PR #1369, opened 2026-07-21
Handle ambiguous Eth send errors with exact transaction lookup

### LINKED ISSUE #1180: pdp: no subpiece found proving error

LexLuthr:
built on https://github.com/filecoin-project/curio/pull/1301

## Summary
This PR changes Eth sends to treat unclear send errors as unknown until Curio can verify the exact signed transaction hash. If the hash is visible, the send is recorded as successful. If it is not visible, the signed transaction stays pending and Harmony retries it instead of recording a false failure.
The same handling is applied to Eth replacement sends, so replacement claims are only finalized when the replacement broadcast result is known.
Tests cover the sender and replacement paths for successful lookup, missing lookup, nonce assignment, signed retry reuse, finalized-row skip, and definitive failure recording.

### DIFF
--- a/tasks/message/sender_eth.go
+++ b/tasks/message/sender_eth.go
@@ -206,8 +206,7 @@ func (s *SendTaskETH) Do(taskID harmonytask.TaskID, stillOwned func() bool) (don
 				"hash", signedTx.Hash().Hex(),
 				"send_error", sendErr,
 				"lookup_error", lookupErr)
-			sendResult = ethSendDefinitiveError
-			sendErr = multierr.Combine(
+			return false, multierr.Combine(
 				xerrors.Errorf("eth transaction send state unknown for %s after settle delay: %w", signedTx.Hash().Hex(), sendErr),
 				xerrors.Errorf("looking up eth transaction %s after unknown send: %w", signedTx.Hash().Hex(), lookupErr),
 			)
@@ -220,37 +219,14 @@ func (s *SendTaskETH) Do(taskID harmonytask.TaskID, stillOwned func() bool) (don
 				"error", sendErr)
 			sendResult = ethSendAccepted
 		} else {
-			pendingNonce, nonceErr := s.checkPendingNonce(fromAddress)
-			if nonceErr != nil {
-				log.Warnw("eth transaction send state unknown; failed to check pending nonce after settle delay",
-					"task_id", taskID,
-					"from", dbTx.FromAddress,
-					"nonce", signedTx.Nonce(),
-					"hash", signedTx.Hash().Hex(),
-					"send_error", sendErr,
-					"nonce_error", nonceErr)
-				sendResult = ethSendDefinitiveError
-				sendErr = multierr.Combine(
-					xerrors.Errorf("eth transaction send state unknown for %s after settle delay: %w", signedTx.Hash().Hex(), sendErr),
-					xerrors.Errorf("checking pending nonce for %s after unknown send: %w", dbTx.FromAddress, nonceErr),
-				)
-			} else {
-				log.Warnw("eth transaction not found after unknown send error",
-					"task_id", taskID,
-					"from", dbTx.FromAddress,
-					"nonce", signedTx.Nonce(),
-					"pending_nonce", pendingNonce,
-					"hash", signedTx.Hash().Hex(),
-					"error", sendErr)
-				sendResult = ethSendDefinitiveError
-				if pendingNonce > signedTx.Nonce() {
-					sendErr = xerrors.Errorf("eth transaction send state unknown for %s: tx not found after %s and sender nonce advanced from %d to %d: %w",
-						signedTx.Hash().Hex(), ethUnknownSendSettleDelay, signedTx.Nonce(), pendingNonce, sendErr)
-				} else {
-					sendErr = xerrors.Errorf("eth transaction send state unknown for %s: tx not found after %s and sender nonce is %d: %w",
-						signedTx.Hash().Hex(), ethUnknownSendSettleDelay, pendingNonce, sendErr)
-				}
-			}
+			log.Warnw("eth transaction not found after unknown send error",
+				"task_id", taskID,
+				"from", dbTx.FromAddress,
+				"nonce", signedTx.Nonce(),
+				"hash", signedTx.Hash().Hex(),
+				"error", sendErr)
+			return false, xerrors.Errorf("eth transaction send state unknown for %s: tx not found after %s: %w",
+				signedTx.Hash().Hex(), ethUnknownSendSettleDelay, sendErr)
 		}
 	}
 
@@ -290,13 +266,6 @@ func (s *SendTaskETH) checkInTransactionPool(hash common.Hash) (bool, error) {
 	}
 
 	return false, nil
-}
-
-func (s *SendTaskETH) checkPendingNonce(from common.Address) (uint64, error) {
-	ctx, cancel := context.WithTimeout(context.Background(), defaultEthCallTimeout)
-	defer cancel()
-
-	return s.client.PendingNonceAt(ctx, from)
 }
 
 func (s *SendTaskETH) signTransaction(ctx context.Context, fromAddress common.Address, tx *types.Transaction) (*types.Transaction, error) {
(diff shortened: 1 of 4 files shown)


tasks/message/sender_eth.go
```diff
@@ -441,3 +502,137 @@ func (s *SenderETH) send(ctx context.Context, fromAddress common.Address, tx *ty
...
+		}
+	}
+	return false
+}
+
+// Source: Lotus chain/messagepool/messagepool.go, ErrExistingNonce.
+// The same signed Filecoin message is already in the local mpool.
+var errMessageWithNonceExists = "message with nonce already exists"
+
+var ethSendAmbiguousErrors = []string{
```
ZenGround0, line 561: The two main substrings I've found that create false negatives from the RPC layer are: 
1. `read tcp :<port-num>-><yyy.xxx.x.xxx>:1234: i/o timeout"`
2. context canceled

Your existing factoring should still catch these as they don't match the whitelist.  But if you want to keep the ambiguous errors list substrings from these errors, probably `i/o timeout` and `context canceled`, should be on this list.

tasks/message/sender_eth.go
```diff
@@ -441,3 +502,137 @@ func (s *SenderETH) send(ctx context.Context, fromAddress common.Address, tx *ty
...
+	if ethSendErrorContainsAny(msg, ethSendDefinitiveErrors) {
+		return ethSendDefinitiveError
+	}
+
+	return ethSendUnknown
+}
+
+func ethSendErrorContainsAny(msg string, parts []string) bool {
+	for _, part := range parts {
+		if strings.Contains(msg, part) {
```
ZenGround0, line 550: This is a probably an interesting thing to speed up if we see this becoming a bottleneck.  The list of errors we're matching is already long enough that it might warrant doing something more clever.  

AI search tells me ~20 different matches is where loop starts getting outperformed and you can use some library like https://github.com/petar-dambovaliev/aho-corasick.  

Probably you'd get better speedup by doing some simple matching of first byte(s) of elements of `parts` over `msg` and then only looping over matches.

Anyway don't worry about any of this until its becoming a problem.

tasks/message/sender_eth.go
```diff
@@ -186,13 +192,46 @@ func (s *SendTaskETH) Do(taskID harmonytask.TaskID, stillOwned func() bool) (don
...
+				"hash", signedTx.Hash().Hex(),
+				"send_error", sendErr,
+				"lookup_error", lookupErr)
+			return false, multierr.Combine(
+				xerrors.Errorf("eth transaction send state unknown for %s: %w", signedTx.Hash().Hex(), sendErr),
+				xerrors.Errorf("looking up eth transaction %s: %w", signedTx.Hash().Hex(), lookupErr),
+			)
+		}
+		if !known {
+			log.Warnw("eth transaction send state unknown; leaving signed transaction for retry",
```
ZenGround0: My biggest open question is what exactly happens to ambiguous error cases that are true negatives.  

My thoughts so far are 
* It looks like they will continually be available for replacement and therefore could end up wasting resources on retry or blocking the nonce queue
* I need to read the base branch to understand this better.  I'm guessing the base branch addresses these concerns and more
* I don't expect there to be any such ambiguous error cases that are true negatives thought I wouldnt be that surprised if there were some

tasks/message/sender_eth.go
```diff
@@ -186,13 +192,46 @@ func (s *SendTaskETH) Do(taskID harmonytask.TaskID, stillOwned func() bool) (don
...
+		known, lookupErr := s.checkInTransactionPool(signedTx.Hash())
+		if lookupErr != nil {
+			log.Warnw("eth transaction send state unknown; failed to look up transaction before retry",
+				"task_id", taskID,
+				"from", dbTx.FromAddress,
+				"nonce", signedTx.Nonce(),
+				"hash", signedTx.Hash().Hex(),
+				"send_error", sendErr,
+				"lookup_error", lookupErr)
+			return false, multierr.Combine(
```
ZenGround0, line 209: One thing I'm a bit worried about: when we return done here we only have 5 tries for the task to fail before it gives up, fails the task and never marks the send row?  Does that leave message sender subsystem in a position where it can hang and callers will never see it resolve?  i.e. in sender_eth `Send` we poll waiting on send success getting set.

I can pretty easily imagine this sort of thing happening because when RPC io errors are happening it is likely that calling the client's `TransactionByHash` client method can also get io timeouted on RPC.

ZenGround0 (commented):
I love the look of this and am seeing why replacement coming first makes sense.  I've left a few discussion points worth reading and talking through.  I'll do another pass after reviewing the base PR

tasks/message/sender_eth.go
LexLuthr: https://github.com/filecoin-project/curio/pull/1369/changes#diff-ec3c7b822dff0c59f0ef6ee7cfe3e25c0d7ef451b8f7c1e634bb3a2e5a89f5d6R530 already covers ctx cancellation. I will add io timeout here.

tasks/message/sender_eth.go
LexLuthr: It is 1000 retries but yes, there is a worst case scenario where caller can be left hanging. Let me fix that.

LexLuthr: Thanks for catching it. I don't know how I missed it.

tasks/message/sender_eth.go
LexLuthr: That's a fair question. I kept thinking about if for a while and then arrived at a conclusion - No matter how much I try to make this PR cover all cases, it will likely get something incorrect. We are going learn a few false positive and false negative as we use this. Those can be fixed later as identified. This also impacts https://github.com/filecoin-project/curio/pull/1369/changes#r3627314686

ZenGround0: While I agree we want to ship something good enough without waiting for something perfect it is worth at least thinking all of these cases through and writing them up here.  At the very least this makes future debugging easier because we'll recognize problems as belongning to these patterns as soon as we see them.

ZenGround0: In this particular case I think the precense of true negatives classified as ambiguous negatives is that we see a lot of retries for this true negative blocking one of our addresses from making progress.

ZenGround0: And in this particular case I think it is ok to merge as is because it would be surprising for errors classified as ambiguous to stay that way in a healthy system.  I expect them to resolve to unambiguous errors or successes when lotus api is healthy.

tasks/message/sender_eth.go
LexLuthr: Fixed

tasks/message/sender_eth.go
```diff
@@ -199,35 +199,59 @@ func (s *SendTaskETH) Do(taskID harmonytask.TaskID, stillOwned func() bool) (don
...
+					"nonce", signedTx.Nonce(),
+					"pending_nonce", pendingNonce,
+					"hash", signedTx.Hash().Hex(),
+					"error", sendErr)
+				sendResult = ethSendDefinitiveError
+				if pendingNonce > signedTx.Nonce() {
+					sendErr = xerrors.Errorf("eth transaction send state unknown for %s: tx not found after %s and sender nonce advanced from %d to %d: %w",
+						signedTx.Hash().Hex(), ethUnknownSendSettleDelay, signedTx.Nonce(), pendingNonce, sendErr)
+				} else {
+					sendErr = xerrors.Errorf("eth transaction send state unknown for %s: tx not found after %s and sender nonce is %d: %w",
```
ZenGround0, line 250: I'm missing what's going on with sendErr.  It looks like all these sendErrs never get used and just get overwritten with "" on line 258

tasks/message/sender_eth.go
```diff
@@ -199,35 +199,59 @@ func (s *SendTaskETH) Do(taskID harmonytask.TaskID, stillOwned func() bool) (don
...
+			log.Infow("eth transaction found after unknown send error",
 				"task_id", taskID,
 				"from", dbTx.FromAddress,
 				"nonce", signedTx.Nonce(),
 				"hash", signedTx.Hash().Hex(),
 				"error", sendErr)
-			return false, xerrors.Errorf("eth transaction send state unknown for %s: %w", signedTx.Hash().Hex(), sendErr)
+			sendResult = ethSendAccepted
+		} else {
+			pendingNonce, nonceErr := s.checkPendingNonce(fromAddress)
```
ZenGround0, line 223: It looks like this is just for error message purposes.  But all errors using this appear to be overwritten.

ZenGround0 (approved):

PR #1372, opened 2026-07-23
docs: update piece removal batch size, be precise about mechanism

rvagg:
Based on changes in https://github.com/filecoin-project/curio/pull/1331

Thinking about this from the appropriate config in synapse in https://github.com/FilOzone/synapse-sdk/pull/898

I do wonder though whether the intention of #1331 was to have a total cap of 200, the current code applies the caps separately - incoming cap of 200, then a check if on-chain is at 200. So you can get it up to 399 with the current rules. Maybe it should have been on-chain + incoming <= 200?

### DIFF
--- a/pdp/README.md
+++ b/pdp/README.md
@@ -719,9 +719,11 @@ When you initiate an upload with the `notify` field specified, the PDP Service w
 
 - **Fields:**
     - `extraData`: *(Optional)* Hex-encoded additional data for the contract call (max 256 bytes decoded).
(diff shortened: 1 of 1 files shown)


rvagg:
Ahhhh .. failure below 400 now: https://github.com/FilOzone/pdp/issues/283#issuecomment-5049574695

so we should probably rework the feature to be 200 all-counted hard cap

LexLuthr (approved):

Issue #1359, opened 2026-07-15
PDP Proving Errors


ZenGround0:
# Background

A [recent operational fire](https://github.com/FilOzone/filecoin-services/issues/546) has us focused on eliminating proving errors.

Proving errors both on mainnet and calibration net over the last 7 days fall into exactly 3 categories
1. #1180 
2. Failure to send proving message because of i/o timeouts on gas estimation 
3. Failure to send proving message because of context deadline exceeded on gas estimation 

Over the last 7 days better stack pdp-warp-speed logs show 2228 instances of proof task errors. 1425 logged errors are type 1 787 logged errors are type 2 and 16 are type 3.

The general pattern for type 1 errors is to retry 5 times (the prove task retry limit) before not submitting a proof.  The pattern for type 2 and 3 is to retry sending until timeout errors do not persist.  I have not yet found any counterexamples of these behaviors.  This means that the only errors not repairable by retries are those of type 1 and there are actually 1425 / 5 = 285 errors over 7 days across calibration and mainnet.

# Preventing new errors

Closing out #1180 should totally resolve this.  

# Fixing old errors 

Even after #1180 lands we will want to fix up the broken schema that the bug has left behind.  We can track the work to repair existing datasets here.  First we must find all of the unaccounted for pieces somehow, either with an expensive sweep or in response to logged errors.  Adding back the subpiece info to the table should be a straightforward database migration.

PR #1373, opened 2026-07-23
Feat: Recover broken create and add 

snadrus:
PDPv0 CreateDataSet / create-and-add can succeed on-chain while Curio never sees a usable eth receipt (pruned history, watcher gaps, restarts).

Using durable chain-state, we verify the completion of these processes. For Create-Dataset we use clientDatasetID from message_waits_eth (exists pre-migration) or the dataset table to get the dataset id from the contract. 

Fake receipts alleging confirmation are added to move the processes forward so that complexities (ex: create-and-add) can proceed on their normal pathways. 

Safety 1: it only operates on receipts before the current proving period. 

RISK 1: Uses extra-data (presumes FWSS for PDP). Alternatives Considered: use external explorers (FilFox). 

Expected First-run behavior: There could be 1000s of receipts needing repair. This will operate one-by-one.


ZenGround0:
Should close #1359

LexLuthr:
@snadrus Can we merge main to rebase this on latest main. Commit history is confusing.

pdp/utils.go
```diff
@@ -51,37 +35,100 @@ func decodeFWSSCreatePayload(extraData []byte) (*fwssCreatePayload, error) {
...
+	if !ok {
+		return nil, fmt.Errorf("createPayload is not bytes")
+	}
+	return createPayload, nil
+}
+
+// decodeFWSSCreatePayload decodes combined create-and-add extraData:
+//
+//	(bytes createPayload, bytes addPayload)
+func decodeFWSSCreatePayload(extraData []byte) (*FWSSCreateIdentity, error) {
```
ZenGround0, line 109: nit: this name doesn't quite work.  "Payload" vs "Data" seems unrelated to this method being specific to create and add.  Maybe `decodeFWSSCreateAndAddData`

tasks/pdpv0/receipt_wait_sync.go
```diff
@@ -0,0 +1,800 @@
...
+	return strings.ToLower(strings.TrimSpace(txHash))
+}
+
+// provingPeriodsPerDay is how many MaxProvingPeriod windows fit in one day,
+// derived from contract PDP config (epochs/day ÷ MaxProvingPeriod).
+func provingPeriodsPerDay(maxProvingPeriod uint64) uint64 {
+	if maxProvingPeriod == 0 {
+		return 1
+	}
+	epochsPerDay := uint64((24 * time.Hour / time.Second) / time.Duration(build.BlockDelaySecs))
```
ZenGround0: nit: you can just say 2880 here

tasks/pdpv0/receipt_wait_sync.go
```diff
@@ -0,0 +1,800 @@
...
+	if err != nil {
+		return nil, false, xerrors.Errorf("DataSetLive(%d): %w", dataSetId, err)
+	}
+	if !live {
+		return nil, false, nil
+	}
+	return matchPieceAddsOnChain(ctx, db, verifier, dataSetId, txHash)
+}
+
+func rebroadcastStuckReceipt(ctx context.Context, eth ethchain.EthClient, wait OutstandingReceipt, signedTx *types.Transaction) error {
```
ZenGround0: I'm not sure if we want to be doing explicit rebroadcast in light of recent developments in message replacement.  If we can unify these both into one mechanism that would be better.  

@LexLuthr we could use your help with a suggestion on what to do here.  Perhaps there is a good entrypoint into the replacement system you could suggest.  Naively I'm considering the route of just turning this into a noop

ZenGround0 (approved):

ZenGround0:
While this is a good change  I don't see how this closes down #1359.  We are only adding extra data now so we will be unable to rely on it for previously failed receipts.  Those previous failures are exactly what continues to cause the proving errors tracked in #1359.  So unless I'm missing something we need an alternative solution.

--edit--

Maybe I'm overreacting here.  Let's sync up on this.

pdp/utils.go
snadrus: I think the old name made better sense: We are extracting data we "shouldn't yet know" from a payload that is not guaranteed to serve our needs. It's sent with the CreateDataset command

snadrus:
Looks like I missed the "find all existing holes and fix them" part of this as my fix only looking at the reciept-based create&add problems, not what is already broken.

tasks/pdpv0/receipt_wait_sync.go
snadrus: I do like math. moved to const

tasks/pdpv0/receipt_wait_sync.go
snadrus: rm for now

PR #1378, opened 2026-07-24
Fix/ Forgot PDP piece router

snadrus:
`make curio-pdp` did not have a way to read out pieces! It was a big change with lots of moving parts and this got missed. Thankfully this is still from Calibration-testing-only phase so no users are actually running it.


ZenGround0 (approved):
Good find

pdp/handlers.go
```diff
@@ -91,7 +96,7 @@ func NewPDPService(
 	stor paths.StashStore,
 	ec ethchain.EthClient,
 	fc PDPServiceNodeApi,
-	sn *message.SenderETH,
+	sn ETHTxSender,
```
LexLuthr, line 103: I am guessing this is for testing?

LexLuthr (approved):
Looks good. nit: can we move the itest into itest directory? It would be great if all itests live together.

pdp/handlers.go
snadrus: yes

PR #1370, opened 2026-07-22
fix: stop PDPv0 save cache scheduler spin

beck-8:
`killalll curio` After that, 4 million lines of logs were generated instantly.

```
{"level":"info","ts":"2026-07-22T12:51:04.743+0800","logger":"curio/message","caller":"message/sender_eth.go:426","msg":"sent transaction","hash":"0x55a67b63e26ce8b9da33ac7e264f294c0e78ff5582f2bc401c9f50e5e74f3
02e","task_id":79062632,"send_error":null,"poll_loops":4}
{"level":"info","ts":"2026-07-22T12:51:04.752+0800","logger":"pdpv0","caller":"pdpv0/task_next_pp.go:290","msg":"Next challenge window scheduled","epoch":"3912536","dataSetId":11651}
{"level":"info","ts":"2026-07-22T12:51:04.995+0800","logger":"curio/message","caller":"message/sender_eth.go:426","msg":"sent transaction","hash":"0x3d2c68e3b29bfe7c315dfc8ae4aa8cb1a648c7c154629d80fc8f25ed68bfe
229","task_id":79062631,"send_error":null,"poll_loops":5}
{"level":"info","ts":"2026-07-22T12:51:04.996+0800","logger":"pdpv0","caller":"pdpv0/task_prove.go:463","msg":"PDP Prove Task: transaction sent","txHash":"0x3d2c68e3b29bfe7c315dfc8ae4aa8cb1a648c7c154629d80fc8f2
5ed68bfe229","dataSetId":12359,"taskID":79061992}
{"level":"info","ts":"2026-07-22T12:51:30.566+0800","logger":"curio/message","caller":"message/watch_eth.go:115","msg":"Assigned orphaned pending transactions","count":1,"machineID":701}
{"level":"info","ts":"2026-07-22T12:51:30.700+0800","logger":"curio/message","caller":"message/watch_eth.go:126","msg":"Processing pending transactions","count":453,"machineID":701}
{"level":"info","ts":"2026-07-22T12:51:31.208+0800","logger":"curio/message","caller":"message/watch_eth.go:126","msg":"Processing pending transactions","count":453,"machineID":701}
{"level":"info","ts":"2026-07-22T12:51:31.984+0800","logger":"harmonytask","caller":"harmonytask/task_type_handler.go:248","msg":"Beginning work on Task","id":79062853,"from":"poller","name":"PDPv0_Prove","sect
or":null}
[...]

PR #1383, opened 2026-07-29
fix: stop IAmBored scheduler spin on shutdown across all schedule() loops

beck-8:
https://github.com/filecoin-project/curio/pull/1370 
its complement.
I hadn't noticed this in many places before. Today when I was killing, I encountered a large amount of input logs from notify.


LexLuthr (commented):
You missed one in tasks/snap/task_submit.go. Please update that. I am approving to unblock merging once that is also updated.

LexLuthr (approved):

ZenGround0 (approved):

PR #1385, opened 2026-07-30
Feat/UI 2nd pass

snadrus:
Minor UI nits:
- USDFC—>FIL LINK to Secured Finance
- Beck's asks from: https://github.com/filecoin-project/curio/issues/1320#issuecomment-5105726073 
- note: time weirdness caught a TIMESTAMP that I needed to convert to TIMESTAMPZ for consistent handling.


deps/deps.go
```diff
@@ -173,6 +173,7 @@ type Deps struct {
 	ServeChunker      *chunker.ServeChunker
 	EthClient         *lazy.Lazy[ethchain.EthClient]
 	Sender            *message.Sender
+	EthSender         *message.SenderETH // set when PDP ETH send task is registered; nil otherwise
```
ZenGround0, line 179: nit: idk if we need to comment here given other deps fields don't explain conditions for nil / non nil

beck-8:
sorry , I updated it in the original comment, so maybe you didn’t see this at the time, so I’m mentioning it here.

0730 update
## 6. "Proving failures" panel shows all-time unrecoverable failures with no time bound, and can silently drop recent ones past a 200-row cap

The "unrecoverable proving failure" rows on the PDP Proving page have no time filter at all — any dataset ever marked unrecoverable stays in the list permanently, including ones from many months ago.

`web/api/webrpc/pdp_proving.go:197-202`:

```go
SELECT id, unrecoverable_proving_failure_epoch, prove_at_epoch, challenge_window
FROM pdp_data_sets
WHERE unrecoverable_proving_failure_epoch IS NOT NULL
ORDER BY id ASC
LIMIT 200
```

snadrus:
@beck-8 Anything more should become new tickets. The UI is close-enough and is free to be improved

lib/usdfcswap/swap.go
```diff
@@ -0,0 +1,391 @@
...
+	"math/big"
+	"strings"
+	"time"
+
+	"github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"golang.org/x/xerrors"
+
```
ZenGround0: Since sushiswap isn't working on calibrationnet as far as I know what are we doing about the UI for calibnet running nodes? We must disable this feature on calibnet UI.  Is that what we are doing?

pdp/contract/addresses.go
```diff
@@ -180,6 +180,17 @@ func ServiceRegistryAddress() (common.Address, error) {
 const USDFCAddressMainnet = "0x80B98d3aa09ffff255c3ba4A241111Ff1262F045"
 const USDFCAddressCalibnet = "0xb3042734b608a1B16e9e86B374A3f3e389B4cDf0"
 
+// SushiSwap V3 (Filecoin mainnet) addresses for USDFC → WFIL → FIL conversion.
+// SwapRouter: https://github.com/sushiswap/v3-periphery/blob/master/deployments/filecoin/SwapRouter.json
+// QuoterV2: https://github.com/sushiswap/v3-periphery/blob/master/deployments/filecoin/QuoterV2.json
+// WFIL is the Sushi-pool wrapped FIL (not an official Filecoin-issued WFIL).
+const (
+	WFILAddressMainnet            = "0x60E1773636CF5E4A227D9AC24F20FECA034EE25A"
```
ZenGround0: Having tested out filecoin exchange on sushi ~ a year ago and revisiting the UI right now I am struggling to understand why we are swapping through WFIL and not directly via FIL.  Going to UI I see the possibility for direct swap:

Image

Was there an intention behind routing through WFIL or is this just an AI recommendation?  If the later please investigate dropping the extra step.  We need a good reason i.e. market is way more liquid.

pdp/contract/addresses.go
```diff
@@ -180,6 +180,17 @@ func ServiceRegistryAddress() (common.Address, error) {
...
+	SushiSwapRouterAddressMainnet = "0x0389879e0156033202C44BF784ac18fC02edeE4f"
+	SushiQuoterV2AddressMainnet   = "0x9B3fF703FA9C8B467F5886d7b61E61ba07a9b51c"
+	SushiUsdfcWfilPoolFeeMainnet  = 500 // 0.05% USDFC/WFIL pool
```
ZenGround0: Note I'm seeing default at 10x higher at 0.5%.  Not sure how much it matters but this may block transactions landing

pdp/contract/addresses.go
```diff
@@ -180,6 +180,17 @@ func ServiceRegistryAddress() (common.Address, error) {
...
 const USDFCAddressCalibnet = "0xb3042734b608a1B16e9e86B374A3f3e389B4cDf0"
 
+// SushiSwap V3 (Filecoin mainnet) addresses for USDFC → WFIL → FIL conversion.
+// SwapRouter: https://github.com/sushiswap/v3-periphery/blob/master/deployments/filecoin/SwapRouter.json
+// QuoterV2: https://github.com/sushiswap/v3-periphery/blob/master/deployments/filecoin/QuoterV2.json
+// WFIL is the Sushi-pool wrapped FIL (not an official Filecoin-issued WFIL).
+const (
+	WFILAddressMainnet            = "0x60E1773636CF5E4A227D9AC24F20FECA034EE25A"
+	SushiSwapRouterAddressMainnet = "0x0389879e0156033202C44BF784ac18fC02edeE4f"
+	SushiQuoterV2AddressMainnet   = "0x9B3fF703FA9C8B467F5886d7b61E61ba07a9b51c"
```
ZenGround0: Why V2 here and V3 protocol referenced in above comments?

ZenGround0: I think this is V3 quoter.

pdp/contract/addresses.go
```diff
@@ -180,6 +180,17 @@ func ServiceRegistryAddress() (common.Address, error) {
 const USDFCAddressMainnet = "0x80B98d3aa09ffff255c3ba4A241111Ff1262F045"
 const USDFCAddressCalibnet = "0xb3042734b608a1B16e9e86B374A3f3e389B4cDf0"
 
+// SushiSwap V3 (Filecoin mainnet) addresses for USDFC → WFIL → FIL conversion.
+// SwapRouter: https://github.com/sushiswap/v3-periphery/blob/master/deployments/filecoin/SwapRouter.json
+// QuoterV2: https://github.com/sushiswap/v3-periphery/blob/master/deployments/filecoin/QuoterV2.json
+// WFIL is the Sushi-pool wrapped FIL (not an official Filecoin-issued WFIL).
+const (
+	WFILAddressMainnet            = "0x60E1773636CF5E4A227D9AC24F20FECA034EE25A"
+	SushiSwapRouterAddressMainnet = "0x0389879e0156033202C44BF784ac18fC02edeE4f"
```
ZenGround0: I'm surprised by this particular flow.  The sushi UI entrypoint is through this address `0xac4c6e212a361c968f1725b4d055b47e63f80b75`.  I see 0x038 is a valid address with some use.  And I *think* it is part of the sushi contract complex. But it looks like 100x less volume.  I'd prefer us to use the more common pathway rather than be creative unless you have a good reason for using a weird internal sushi api.

ZenGround0 (changes requested):
This might be fine but I'm seeing enough red flags with the sushi swap integration that I want us to slow down on this one.  

Primarily I want us to (1) make sure we have a good mental model of exactly what we are doing and how it relates to normal sushi swap usage.  (2) make sure the lack of calibnet swapping is reflected in the UI probably with both a warning when curio is running calibnet and definitely with full disabling in calibnet.

This is one of the few places where we can very directly lose people money.  At the very least we need to test this out

ZenGround0:
The other thing is that we 100% must execute test swaps on a testing mainnet node before we consider merging this.  This is a good candidate motivator for a small internal curio testing node.

pdp/contract/addresses.go
ZenGround0: I think once you start calling the standard entrypoint (like this UI does) then this WFIL complexity goes away

ZenGround0: I think once you start calling the standard entrypoint (like this UI does) then this WFIL complexity goes away

deps/deps.go
snadrus: it's uncommon to have nil deps in Curio

lib/usdfcswap/swap.go
snadrus: The UI shows a message that it's disabled and there are no controls visible to use it. I don't know that we have a test plan for this in prod.

pdp/contract/addresses.go
snadrus: AI. It's been changed to:
`    multicall([exactInputSingle(recipient=0), unwrapWETH9(min, wallet)])
`
which keeps the WFIL in the router and only returns FIL to the wallet.
This matches the abi of the [SushiSwap V3 router.](https://github.com/sushiswap/v3-periphery/blob/master/deployments/filecoin/SwapRouter.json)

pdp/contract/addresses.go
snadrus: The number is from this pool: 
https://www.geckoterminal.com/filecoin/pools/0x4e07447bd38e60b94176764133788be1a0736b30

pdp/contract/addresses.go
snadrus: That's the [UI aggregator](https://docs.sushi.com/contracts/red-snwapper) which makes it to the same pool through some offchain bytes addition. 
We are using the "public Uniswap V3 periphery interface" as deployed by SushiSwap, as that appears more standardized than the former.

pdp/contract/addresses.go
snadrus: yes, and it had a revision since V3. Renamed to SushiV3PeripheryQuoterV2AddressMainnet for clarity.

snadrus:
I agree completely, we need a place to test this code.

beck-8:
I was just about to test this feature. I suddenly realized that we were missing something: withdraw usdfc from the pay account, otherwise SP would still need to withdraw money through external tools.

snadrus:
@beck-8 It was decided that it's safer for funds to be handled by linking to Secured Finance's swap rather than using the contracts here.

LexLuthr (approved):

ZenGround0 (approved):

PR #1387, opened 2026-07-30
pdp: price eth sender fee caps with feeHistory floor and 2x baseFee headroom

TippyFlitsUK:
Send-time fee-cap pricing hardening from [#1354](https://github.com/filecoin-project/curio/issues/1354), as discussed with @ZenGround0 in #foc-wg. Two changes in `SenderETH.send`:

1. **Floor the baseFee at the max over the last 120 epochs of `eth_feeHistory`** before pricing the cap. Without this, a restart into a quiet chain prices caps at the momentary low, and they strand minutes later once load ramps baseFee back up.
2. **Price `gasFeeCap` as `2x baseFee + tip`** instead of `baseFee + tip`, so a baseFee rise between send and inclusion does not strand the message. With EIP-1559 semantics the sender still pays only `baseFee + tip` at inclusion, so the extra headroom costs nothing when unused.

Motivation: with zero headroom and no fee-bump path, any baseFee rise strands the head message, and sequential nonces then block the wallet's whole lane, including proving messages (two proving near-misses documented in #1354).

These mitigations have been running in production on two mainnet PDP SPs since 2026-07-10. During the recent congestion event ([filecoin-services#546](https://github.com/FilOzone/filecoin-services/issues/546)), the patched idle SP held 100% dealbot deal success over 12h while stock-sender SPs failed on confirmation timeouts in the same window.

Out of scope: same-nonce replacement / RBF for external spikes that outrun a fixed cap. That is [#1292](https://github.com/filecoin-project/curio/issues/1292), addressed by #1301. Per Zen this PR covers the send-time pricing leg #1301 does not.

### DIFF
--- a/tasks/message/sender_eth.go
+++ b/tasks/message/sender_eth.go
@@ -395,6 +395,14 @@ func (s *SenderETH) send(ctx context.Context, fromAddress common.Address, tx *ty
 			return common.Hash{}, fmt.Errorf("base fee not available; network might not support EIP-1559")
 		}
 
+		if feeHist, fhErr := s.client.FeeHistory(ctx, 120, nil, nil); fhErr == nil {
+			for _, histFee := range feeHist.BaseFee {
+				if histFee != nil && histFee.Cmp(baseFee) > 0 {
+					baseFee = histFee
+				}
+			}
+		}
+
 		// Set GasTipCap (maxPriorityFeePerGas)
 		gasTipCap, err := s.client.SuggestGasTipCap(ctx)
 		if err != nil {
@@ -402,7 +410,7 @@ func (s *SenderETH) send(ctx context.Context, fromAddress common.Address, tx *ty
 		}
 
 		// Calculate GasFeeCap (maxFeePerGas)
-		gasFeeCap := new(big.Int).Add(baseFee, gasTipCap)
+		gasFeeCap := new(big.Int).Add(new(big.Int).Mul(baseFee, big.NewInt(2)), gasTipCap)
 
 		chainID, err := s.client.NetworkID(ctx)
 		if err != nil {
--- a/tasks/message/sender_eth.go
+++ b/tasks/message/sender_eth.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"math/big"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/ethereum/go-ethereum"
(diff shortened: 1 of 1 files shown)


tasks/message/sender_eth.go
```diff
@@ -395,14 +395,22 @@ func (s *SenderETH) send(ctx context.Context, fromAddress common.Address, tx *ty
 			return common.Hash{}, fmt.Errorf("base fee not available; network might not support EIP-1559")
 		}
 
+		if feeHist, fhErr := s.client.FeeHistory(ctx, 120, nil, nil); fhErr == nil {
```
ZenGround0:
```suggestion
                 // Measure current basefee as the max over the last 120 epochs (1 hour) 
		if feeHist, fhErr := s.client.FeeHistory(ctx, 120, nil, nil); fhErr == nil {
```

ZenGround0 (approved):

ZenGround0:
@LexLuthr would be good for you to sanity check this but I'm good to merge asap

ZenGround0:
Looks like it needs gofmt and make gen

LexLuthr (commented):
This is fine. You also need to update the same in https://github.com/filecoin-project/curio/blob/066bab633e4f0ce02eea20f03f7ef828311ea063/tasks/message/replace_eth.go#L444-L477. Ask your AI to take full context into account for replacement message. Intent is to handle continuous upward spike with minimum Fee.

LexLuthr:
> This is fine. You also need to update the same in
> 
> https://github.com/filecoin-project/curio/blob/066bab633e4f0ce02eea20f03f7ef828311ea063/tasks/message/replace_eth.go#L444-L477
> 
> . Ask your AI to take full context into account for replacement message. Intent is to handle continuous upward spike with minimum Fee.

I think you can ignore it for now. I will add it later when I enable it back. Fix the fmt and make gen

tasks/message/sender_eth.go
```diff
@@ -395,14 +395,23 @@ func (s *SenderETH) send(ctx context.Context, fromAddress common.Address, tx *ty
 			return common.Hash{}, fmt.Errorf("base fee not available; network might not support EIP-1559")
 		}
 
+                 // Measure current basefee as the max over the last 120 epochs (1 hour) 
+		if feeHist, fhErr := s.client.FeeHistory(ctx, 120, nil, nil); fhErr == nil {
```
LexLuthr: This should be cached. It is too heavy on Lotus side.

LexLuthr: Maybe even put a timeout

LexLuthr (commented):
While intent is correct, I think we should take into account what this call with do the Lotus side perf at scale.

PR #1389, opened 2026-07-30
Default pdp config

### LINKED ISSUE #1153: PDP config default layer
Lets enshrine the beautiful PDP config recommendations from the experimental features docs into a default config that ships with the node.

ZenGround0:
Closes #1153


LexLuthr:
We probably don't need this as config is now fully automated in https://github.com/filecoin-project/curio/blob/main/pdpnode/config_skiff.go
Users who are running Porep and PDP together won't need this either as it will require custom config with more params. 
Essentially issue is handled even without this PR.

PR #1390, opened 2026-07-30
Fix: piece IPNI index should be respected despite temporary chain outages

### LINKED ISSUE #1381: pdp: transient chain-read failure during AddPieces permanently disables IPFS indexing for the whole batch
### Summary

`CheckIfIndexingNeeded` (`pdp/indexing.go`) performs two live chain reads (`GetDataSetListener`, `GetDataSetMetadataAtKey`) inline in every AddPieces request. When the metadata read fails transiently (lotus timeout under load), the error branch returns `false, nil`, recording the batch as "indexing not wanted." The resulting state (`needs_indexing = false, indexed_at IS NULL`) is indistinguishable from a deliberate opt-out, is invisible to the indexing scheduler's only query (`WHERE needs_indexing = TRUE`), and is never revisited by any reconciler. The pieces are permanently unservable by IPFS root CID while being live, billable, and challengeable on-chain.

Observed on 1.28.2-rc2-based mainnet builds; the code path is unchanged on current `main`: https://github.com/filecoin-project/curio/blob/main/pdp/indexing.go#L36-L42

### Evidence (two of our mainnet nodes)

```
WARN pdpv0 pdp/indexing.go:40 Failed to get data set metadata, skipping indexing
  {"error": "chain: read tcp 127.0.0.1:...->127.0.0.1:1234: i/o timeout", "dataSetId": 1293}
```

The warn line is the only signal emitted. There is no failed task row, no pending flag, and nothing for harmonytask's retry machinery to pick up.

- Node A: 124 warn occurrences since 17 Jun (still occurring today, 28 Jul); node B: 39 occurrences.
- On node B the correlation is exact: 29 warns for data set 1344 -> 1,160 never-indexed pieces (29 x 40); 10 warns for data set 1363 -> 400 never-indexed pieces (10 x 40).
- Damage is always whole-batch (40 consecutive pieceIds per event) and clusters on chain-congestion days (24-25 Jun, 7-10 Jul, 21 Jul).
- Surfaced externally: a client-side `verify` found 400 committed root CIDs returning not-found from the SP's `/ipfs/` endpoint; all 400 mapped 1:1 to the never-indexed pieces (verified by set equality on pieceIds).
- Across our two nodes: ~51,000 affected pieces in data sets that demonstrably have `withIPFSIndexing` set (sibling pieces in the same sets indexed normally). A second operator's node shows the same warn pattern in shared log streaming, so this is not site-specific.

### Operator detection query

```sql
SELECT p.data_set, COUNT(*) FROM pdp_data_set_pieces p
JOIN pdp_piecerefs r ON r.id = p.pdp_pieceref
WHERE r.indexed_at IS NULL AND r.needs_indexing = FALSE AND NOT p.removed
GROUP BY p.data_set;
```

Interpret against whether each data set actually has `withIPFSIndexing`; sets that never requested indexing legitimately show here.

### Root cause

The error branch comment states it directly: "Hard to differentiate between unsupported listener type OR internal error - So we log and skip indexing attempt." A transient RPC failure is thereby converted into a permanent per-batch indexing opt-out, silently, at warn level.

snadrus:
This stores the dataset index preference on the database (rather than doing a chain lookup). A migrator ensures that eventually every dataset has this. The repair task from [feat/create-and-add-recovers](https://github.com/filecoin-project/curio/pull/1373) also calls this to fix the index. 
This fixes https://github.com/filecoin-project/curio/issues/1381

Note: both these PRs expect FWSS to front PDP. 

This MAY be a good start for the upcoming work to remove some PDP metadata from the chain.

It's built atop [feat/create-and-add-recovers](https://github.com/filecoin-project/curio/pull/1373)

Superceded by #1392  for the correct merge target


LexLuthr:
@snadrus Can we switch base from main to feat/create-and-add-recovers to get the correct diff?

PR #1391, opened 2026-07-31
Bump extradata size limits to accommodate richer auth (programmable-acls)

JAG-UK:
Bump extradata size limits to accommodate richer auth

ONLY NEEDED IF https://github.com/FilOzone/filecoin-services/pull/536 MERGES!

Rationale: we already know that passkeys and WebAuthn are targets for identified service users. WebAuthn envelope with P256 on its own is around 700 bytes so rounding up to a KiB gives space for usable payload without going overboard.

For TerminateService and SchedulePieceRemovals we just set the limit to 1KiB as the only thing that was in there before was a legacy signature, and we make it either/or. For AddPieces we increase by 1KiB so as not to reduce the total capacity for the metadata that's also carried in there. I guess in principle we could take 160b off this to remove the previous SKR entry but I'm not sure how worthwhile that is with an already large structure.

Note that there's a small inconsistency here: Curio has addPieces max as 8192, and I've stayed faithful to that here, but the contracts have it currently at 4096 and the ACL proposal only bumps to 5120. Worth fixing?

### DIFF
--- a/cmd/pdptool/main.go
+++ b/cmd/pdptool/main.go
@@ -45,8 +45,9 @@ func validateExtraData(extraDataHexStr string) error {
 	if err != nil {
 		return fmt.Errorf("failed to decode hex in extra-data: %w", err)
 	}
-	if len(decoded) > 2048 {
-		return fmt.Errorf("decoded extra-data exceeds maximum size of 2048 bytes (decoded length: %d)", len(decoded))
+	// 8192 bytes is the size of the largest extradata (for addPieces). Individual smaller limits are checked later
+	if len(decoded) > 8192 {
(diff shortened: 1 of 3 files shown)


LexLuthr (approved):
LGTM!

jennijuju:
the PR title is a bit misleading imho, this PR bumps the limit but actually doesn't any add any write or read perm checks, is that right?

jennijuju:
> the PR title is a bit misleading imho, this PR bumps the limit but actually doesn't any add any write or read perm checks, is that right?

I guess for write, we dont need to add anything cuz the signature validation would fail upon addPiece and deletePiece calls that has an authorizer;

LexLuthr:
@JAG-UK @jennijuju If there are no objections, we can merge this. It has not negative impact on Curio side so we should be fine.

PR #1392, opened 2026-07-31
Fix: piece IPNI index should be respected despite temporary chain outages

### LINKED ISSUE #1381: pdp: transient chain-read failure during AddPieces permanently disables IPFS indexing for the whole batch

snadrus:
This stores the dataset index preference on the database (rather than doing a chain lookup). A migrator ensures that eventually every dataset has this. The repair task from https://github.com/filecoin-project/curio/pull/1373 also calls this to fix the index.
This fixes https://github.com/filecoin-project/curio/issues/1381

Note: both these PRs expect FWSS to front PDP.

This MAY be a good start for the upcoming work to remove some PDP metadata from the chain.

It's built atop https://github.com/filecoin-project/curio/pull/1373


cmd/curio/internal/translations/locales/ko/out.gotext.json
```diff
@@ -0,0 +1,4 @@
+{
```
ZenGround0, line 1: Where do these new translation files come from ?

harmony/harmonydb/sql/20260730-pdp-v0-dataset-ipni.sql
```diff
@@ -0,0 +1,17 @@
+-- Cache withIPFSIndexing intent on the data set so AddPieces does not depend on
+-- live chain metadata reads. NULL means unresolved (backfill / first resolve).
+DO $$
+BEGIN
+    IF NOT EXISTS (
+        SELECT 1
+        FROM information_schema.columns
+        WHERE table_name = 'pdp_data_sets'
+          AND column_name = 'ipni'
```
ZenGround0: nit: I'm not sure if this is the best name.  Though the two concepts overlap completely for now IPFSIndexing is distinct from ipni.  It includes a specific CAR data format assumption and includes cassandra indexing along with IPNI

tasks/pdpv0/task_dataset_idx.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	var resolveErrs int
+	for _, row := range rows {
+		if !stillOwned() {
+			return false, nil
+		}
+		mustIndex, err := pdp.ResolveDatasetShouldIPNI(ctx, t.db, t.ethClient, row.ID)
+		if err != nil {
+			log.Warnw("dataset idx fixup failed for data set", "dataSetId", row.ID, "error", err)
```
ZenGround0: nit: maybe its clear from the err message but maybe highlight in log message that the failing step is the resolution

pdp/indexing.go
```diff
@@ -2,74 +2,169 @@ package pdp
...
+	}
+	if !ipni.Valid {
+		return nil, nil
+	}
+	v := ipni.Bool
+	return &v, nil
+}
 
-	payload, err := decodeFWSSCreatePayload(extraData)
+func fetchFromChain_shouldIPNI(ctx context.Context, ethClient ethchain.EthClient, dataSetId uint64) (bool, error) {
```
ZenGround0: nit: I'm not so familiar with underscore method names since the common recommendation is camel case everything.  What does _ indicate in go naming convention?

ZenGround0: Also should this be symmetric to the above name?  i.e. fetchFromChain_DatasetShouldIPNI

tasks/pdpv0/dataset_create_watch.go
```diff
@@ -108,12 +110,21 @@ func processDataSetCreate(ctx context.Context, db *harmonydb.DB, psc DataSetCrea
...
+
 	comm, err := db.BeginTransaction(ctx, func(tx *harmonyquery.Tx) (commit bool, err error) {
 		// Insert a new entry into pdp_data_sets
 		n, err := tx.Exec(`
-        INSERT INTO pdp_data_sets (id, create_message_hash, service, proving_period, challenge_window)
-        VALUES ($1, $2, $3, $4, $5)
-    `, dataSetId, psc.CreateMessageHash, psc.Service, provingPeriod, challengeWindow)
+        INSERT INTO pdp_data_sets (id, create_message_hash, service, proving_period, challenge_window, ipni)
+        VALUES ($1, $2, $3, $4, $5, $6)
+    `, dataSetId, psc.CreateMessageHash, psc.Service, provingPeriod, challengeWindow, ipniArg)
```
ZenGround0: Theres an edge case worth thinking about.

What about in flight `pdp_dataset_creates` rows with nil extra data in the db but not on chain and ipfsIndexing set to true?  I think ipniArg will then be set to false.  And it won't be resolved by task_dataset_ids.

ZenGround0: Looking at the PR this is based on I see this column was just added.  This kind of issue is going to exist if there isn't a migration explicitly setting extraData for all outstanding creates.

ZenGround0: I suppose these rows could be manually fixed without much trouble too

ZenGround0 (approved):
Overall this looks great.

I don't think the boundary race condition is blocking.  But worth thinking about a bit to see if there's an easy way to avoid that case entirely.

pdp/indexing.go
snadrus: I needed a way for "should" to modify "dataset" and not "fetch". Open to suggestions.

cmd/curio/internal/translations/locales/ko/out.gotext.json
snadrus: I ask ChatGPT to translate them. We should have a language discussion as the current mix is possibly off-target.

harmony/harmonydb/sql/20260730-pdp-v0-dataset-ipni.sql
snadrus: rn ipfsIndexing

pdp/indexing.go
snadrus: the naming is better now with ipfsIndexing

tasks/pdpv0/dataset_create_watch.go
snadrus: i'm changing it in-place in the file since nobody's ran it.

tasks/pdpv0/task_dataset_idx.go
snadrus: improved

PR #1396, opened 2026-08-04
Repair missing pdp piece rows 

### LINKED ISSUE #1359: PDP Proving Errors

ZenGround0:
Closes #1359 This is heavily ai assisted and a rough draft but I think it does the basic job.

cc @LexLuther if you could take this over when you are back online we can get this in for thursday release.

### DIFF
--- a/tasks/pdpv0/task_repair_missing_pieces.go
+++ b/tasks/pdpv0/task_repair_missing_pieces.go
@@ -1,0 +1,533 @@
+package pdpv0
+
+import (
+	"context"
+	"encoding/hex"
+	"errors"
+	"math/big"
+	"strings"
+	"time"
+
+	"github.com/ipfs/go-cid"
+	"github.com/yugabyte/pgx/v5"
+	"golang.org/x/xerrors"
+
+	"github.com/filecoin-project/go-padreader"
+	"github.com/filecoin-project/go-state-types/abi"
+
+	"github.com/filecoin-project/curio/harmony/harmonydb"
+	"github.com/filecoin-project/curio/harmony/harmonytask"
+	"github.com/filecoin-project/curio/harmony/resources"
+	"github.com/filecoin-project/curio/harmony/taskhelp"
+	"github.com/filecoin-project/curio/lib/ethchain"
+	"github.com/filecoin-project/curio/lib/passcall"
+	"github.com/filecoin-project/curio/lib/storiface"
+	"github.com/filecoin-project/curio/pdp"
+	"github.com/filecoin-project/curio/pdp/contract"
+	"github.com/filecoin-project/curio/tasks/tasknames"
+)
+
+// RepairMissingPiecesTask reconciles one data set against PDPVerifier and
+// rebuilds pdp_data_set_pieces rows for pieces the contract reports as live but
+// which Curio has no local record of.
+//
+// # Why this exists
+//
+// A piece could land on chain without its pdp_data_set_pieces row ever being
+// written (filecoin-project/curio#1359). Two things then go wrong. Proving
+// fails with "no subpiece found" whenever a challenge lands inside that piece,
+// because provePiece resolves the challenged leaf through that table. And
+// pdp_piecerefs.data_set_refcount, which is maintained by an insert trigger on
+// that same table, stays at 0, so the piece GC is free to drop the pieceref and
+// eventually the bytes.
+//
+// # Scope
+//
+// The add-piece path is fixed in the release that carries this task, so only
+// data sets that predate the upgrade can be damaged. The work queue,
+// pdp_data_set_piece_scrub, is seeded once by the migration from the
+// pdp_data_sets rows present at that moment. Nothing enqueues a data set later.
+// Once every seeded row is complete the task goes permanently idle.
+//
+// # What a repair writes
+//
+// PDPVerifier knows the piece id and the PieceCID v2. Everything else in the
+// row follows from treating the piece as its own single sub-piece: sub_piece is
+// the piece, sub_piece_offset is 0, and sub_piece_size is the piece's padded
+// size derived from the v2 CID. There is no add message to reference, so
+// add_message_hash gets a sentinel and add_message_index gets 0.
+//
+// # When data is gone
+//
+// Before writing a row the task confirms the bytes are actually reachable: a
+// complete long-term parked_pieces row for the CID, which the piece store can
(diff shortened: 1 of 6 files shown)


tasks/pdpv0/task_repair_missing_pieces.go
```diff
@@ -0,0 +1,535 @@
...
+		if pieceRefID == 0 {
+			var refID int64
+			err := tx.QueryRow(`
+				INSERT INTO parked_piece_refs (piece_id, long_term)
+				VALUES ($1, TRUE)
+				RETURNING ref_id
+			`, loc[0].ParkedPieceID).Scan(&refID)
+			if err != nil {
+				return false, xerrors.Errorf("failed to insert parked_piece_refs for parked piece %d: %w", loc[0].ParkedPieceID, err)
+			}
```
LexLuthr: I think AI went a little overboard here. Typical AI extra guards. pdp_pieceref cannot exist without parked_piece_ref.

tasks/pdpv0/task_repair_missing_pieces.go
```diff
@@ -0,0 +1,535 @@
...
+
+			err = tx.QueryRow(`
+				INSERT INTO pdp_piecerefs (service, piece_cid, piece_ref, created_at, needs_save_cache)
+				VALUES ($1, $2, $3, NOW(), $4)
+				RETURNING id
+			`, service, pieceCidV1, refID, needsSaveCache).Scan(&pieceRefID)
```
LexLuthr: We probably don't need to insert another ref. Currently `pdp_pieceref` are not really one to one mapping to pieces. Might be better to just reuse the existing one.

tasks/pdpv0/task_repair_missing_pieces.go
```diff
@@ -0,0 +1,535 @@
...
+		if err != nil {
+			return false, err
+		}
+
+		// Resume past the last piece this page covered. Piece ids are returned
+		// in ascending order, so the next cursor is one past the tail.
+		last := page.PieceIds[len(page.PieceIds)-1]
+		if !last.IsInt64() {
+			return false, xerrors.Errorf("piece id %s on data set %d is out of int64 range", last, dataSetID)
+		}
```
LexLuthr, line 148: Seems redundant if we are measuring len before.

tasks/pdpv0/task_repair_missing_pieces.go
```diff
@@ -0,0 +1,535 @@
...
+			stop = false
+			return true, nil
+		})
+		if stop {
+			return nil
+		}
+	}
+
+	return nil
+}
```
LexLuthr, line 464: We probably don't need this. `SingletonTaskAdder` would make more sense for this type of task.

tasks/pdpv0/task_repair_missing_pieces.go
```diff
@@ -0,0 +1,535 @@
...
+	return harmonytask.TaskTypeDetails{
+		Max:  taskhelp.Max(2),
+		Name: tasknames.PDPv0_FixPiece,
+		Cost: resources.Resources{
+			Cpu: 0,
+			Gpu: 0,
+			Ram: 64 << 20,
+		},
+		MaxFailures: 3,
+		IAmBored: passcall.Every(time.Minute, func(taskFunc harmonytask.AddTaskFunc) error {
```
LexLuthr: Too frequent for a repair task.

tasks/pdpv0/task_repair_missing_pieces.go
LexLuthr: Nevermind! I guess we are never going to schedule once a dataSet is complete.

tasks/pdpv0/task_repair_missing_pieces.go
LexLuthr: Even when we finish processing all the dataSets, this will keep running every minute and cause a DB read.

LexLuthr (commented):
Design is good.

tasks/pdpv0/task_repair_missing_pieces.go
ZenGround0: It's just an annoying safety check when casting bigInt into int.  I guess piece ids can technically have big gaps.

ZenGround0:
Closing and reopening to attempt to trigger ci

tasks/pdpv0/task_repair_missing_pieces.go
```diff
@@ -0,0 +1,489 @@
...
+	if err := reader.Close(); err != nil {
+		log.Warnw("failed to close parked piece reader after repair check",
+			"dataSetId", dataSetID, "pieceId", pieceID, "parkedPieceId", loc[0].ParkedPieceID, "error", err)
+	}
+
+	comm, err := t.db.BeginTransaction(ctx, func(tx *harmonydb.Tx) (bool, error) {
+		pieceRefID := loc[0].PieceRefID
+		// The chain CID describes the whole piece, so the piece is its own only
+		// sub-piece: offset 0, size the full padded size.
+		n, err := tx.Exec(`
```
snadrus, line 297: this could run out of transaction

snadrus (approved):
looks good

PR #1403, opened 2026-08-06
chore: release v1.28.3

LexLuthr:
## Summary
- Update Lotus version
- Update FFI version
- Update other direct deps

### Replace deprecated and spoofable client-IP handling with an explicit trust model ([flagged by linter](https://github.com/filecoin-project/curio/actions/runs/31101759293/job/92616884758))
This prevents clients from evading limits—or exhausting another client’s bucket—by forging IP headers.
Non-loopback proxies are intentionally treated as the client unless trusted proxy CIDRs are added in the future.
- Remove middleware.RealIP and handlers.ProxyHeaders, which trusted client-controlled forwarding headers and mutated RemoteAddr.
- Resolve direct clients from the TCP peer address.
- Trust one-hop X-Forwarded-For only when the immediate peer is loopback; missing or malformed XFF falls back to the peer address.
- Use the resolved address consistently for MK12/MK20 rate limiting and PDP offense throttling.
- Canonicalize IPv6 addresses to /64 to prevent address rotation from bypassing limits.


snadrus:
Back when it was about arbitrary logging, the old "best effort" IP system was good enough. I could imagine some configurations not using loopback for x-forwarded-for (ex: Docker). We may need a "trusted list" eventually there.

snadrus (approved):

ZenGround0 (approved):

PR #1406, opened 2026-08-07
fix: fix crash when a machine has no harmony_machine_details row

beck-8:
> in skiff

```
{"level":"error","ts":"2026-08-07T13:17:46.041+0800","logger":"harmonytask","caller":"harmonytask/task_type_handler.go:345","msg":"Do() returned error","type":"AlertManager","id":"82763501","error":"getting addresses: getting config layers for all machines: scanning: scanning: doing scan: scanFn: scany: scan row into struct fields: can't scan into dest[0] (col: layers): cannot scan NULL into *string","errorVerbose":"getting addresses:\n    github.com/filecoin-project/curio/alertmanager.(*AlertTask).Do\n        /root/workspace/curio/alertmanager/task_alert.go:194\n  - getting config layers for all machines:\n    github.com/filecoin-project/curio/deps/config.GetAddressesFromConfig\n        /root/workspace/curio/deps/config/utils.go:147\n  - scanning: scanning: doing scan: scanFn: scany: scan row into struct fields: can't scan into dest[0] (col: layers): cannot scan NULL into *string"}
{"level":"info","ts":"2026-08-07T13:17:46.060+0800","logger":"harmonytask","caller":"harmonytask/task_type_handler.go:286","msg":"Beginning work on Task","id":82763501,"from":"poller","name":"AlertManager","sector":null}
```

### DIFF
--- a/deps/config/utils.go
+++ b/deps/config/utils.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"context"
+	"database/sql"
 	"reflect"
 	"slices"
 	"strings"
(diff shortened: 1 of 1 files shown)


deps/config/utils.go
LexLuthr, line 149: Let's return early if it is nil instead of using the length in loop.

LexLuthr (approved):
Needs 1 tiny change.

PR #1410, opened 2026-08-10
add disable SingletonTaskAdder

### LINKED ISSUE #1400: Allow SingletonTaskAdder to stop scheduling tasks
We have a bunch of one time repair tasks which utilise SingletonTaskAdder. We should add a disabled table. We can add tasks to this table once our repair is complete. Then SingletonTaskAdder should not schedule any task from this disabled table. It makes for a much cleaner one off repair task system.

beck-8:
Closed: https://github.com/filecoin-project/curio/issues/1400

### DIFF
--- a/harmony/harmonydb/downgrade/20260807-singleton-task-disabled.sql
+++ b/harmony/harmonydb/downgrade/20260807-singleton-task-disabled.sql
@@ -1,0 +1,1 @@
+DROP TABLE IF EXISTS harmony_task_singleton_disabled;
--- a/harmony/harmonydb/downgrade/20260807-singleton-task-disabled.sql
+++ b/harmony/harmonydb/downgrade/20260807-singleton-task-disabled.sql
@@ -1,1 +1,0 @@
-DROP TABLE IF EXISTS harmony_task_singleton_disabled;
--- a/itests/harmonytask/singleton_disabled_test.go
+++ b/itests/harmonytask/singleton_disabled_test.go
@@ -1,0 +1,38 @@
+package harmonytask
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/filecoin-project/curio/harmony/harmonytask"
+)
+
+func TestSingletonTaskDisabled(t *testing.T) {
+	db := getDB(t)
+	ctx := context.Background()
+
+	tk := newTestTask("SingletonDisabledTest", 5)
+	tk.iAmBored = harmonytask.SingletonTaskAdder(300*time.Millisecond, tk)
+	t.Cleanup(cleanupTasks(tk))
+
+	_, err := db.Exec(ctx, `INSERT INTO harmony_task_singleton_disabled (task_name, reason) VALUES ($1, $2)`,
+		tk.name, "test: repair complete")
+	require.NoError(t, err)
+
+	makeEngine(t, db, []harmonytask.TaskInterface{tk}, "singdis:1000")
+
+	select {
+	case id := <-tk.doneCh:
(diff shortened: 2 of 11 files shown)


beck-8:
`PDPv0_FixPiece` is not using `SingletonTaskAdder`, I'm not sure if it should be handled here as well?

harmony/harmonydb/downgrade/20260807-singleton-task-disabled.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE IF EXISTS harmony_task_singleton_disabled;
```
LexLuthr: We don't drop table or remove columns in downgrade.

itests/harmonytask/singleton_disabled_test.go
LexLuthr, line 1: All tests in itest/harmonytask are disabled in CI. They should be enabled by adding a proper entry. Only 1 entry for the directory.

LexLuthr (changes requested):
Basics are good but I am not expecting users to disable the task. Curio should disable the task automatically.

tasks/pdpv0/task_dataset_idx.go
```diff
@@ -87,6 +87,7 @@ func (t *DatasetIdxTask) TypeDetails() harmonytask.TaskTypeDetails {
 		},
 		MaxFailures: 3,
 		IAmBored:    harmonytask.SingletonTaskAdder(datasetIdxInterval, t),
+		CanDisable:  true,
```
beck-8, line 94: @LexLuthr  Shouldn't this task be disabled?

itests/harmonytask/singleton_disabled_test.go
beck-8: It looks like it keeps failing. Maybe it was intentional that he didn't add it at the beginning.

beck-8:
Should I remove this CI test first?

LexLuthr:
@beck-8 We need to fix the tests. Please use your AI to fix the harmony_task one.

web/static/pages/task/singleton-details.mjs
LexLuthr, line 1: While I do see the appeal to enable and disable tasks from GUI. We should not allow disabling Singletons from there. Some of them are very important and will break the SP because they decided to disable something without full understanding of consequences. GUI is for ease but it can easily turn against the user if we are not careful.

LexLuthr (changes requested):

tasks/pdpv0/task_dataset_idx.go
LexLuthr: @snadrus Hey! Is this task for one time repair or ongoing?

web/static/pages/task/singleton-details.mjs
beck-8: Yes, so I added a task to restrict shutdown in subsequent commit submissions. Only those that are explicitly candisabled are allowed to be closed. Currently there is a fix piece and a dataset idx

So is it acceptable at present?

Or should we remove it from the frontend but keep this API?

LexLuthr: Remove it from frontEnd. Keep in API for future use if we end up requiring it.

LexLuthr (approved):
Looks good now. Can you please add a screenshot of disabled task in GUI.

tasks/pdpv0/task_dataset_idx.go
LexLuthr: @beck-8 Can you disable the task correctly? It should be disabled once repair is complete.

beck-8:
image

beck-8:
image

PR #1411, opened 2026-08-10
fix: don't assign expired deals to sectors for seailng

### LINKED ISSUE #781: mk12: Don't assign expired deals to sectors for seailng
Repro:
* Make a deal that can't be assigned to a sector (e.g. no sectors available for snap)
  * (ideally curio would reject the deal here already but that's a separate thing)
* Wait to pass deal start epoch
* Extend sectors such that those deals can be assigned to sectors
* Curio now assigns the deal to be sealed, which will obviously fail

Probably applies to mk20 as well

LexLuthr:
fixes #781


tasks/storage-market/task_commp.go
```diff
@@ -379,29 +379,38 @@ var _ = harmonytask.Reg(&CommpTask{})
...
-	if updatePipeline {
-		n, err := db.Exec(ctx, `DELETE FROM market_mk12_deal_pipeline WHERE uuid = $1`, deal)
+	committed, err := db.BeginTransaction(ctx, func(tx *harmonydb.Tx) (bool, error) {
+		n, err := tx.Exec(`UPDATE market_mk12_deals SET error = $1 WHERE uuid = $2`, reason, deal)
 		if err != nil {
-			return xerrors.Errorf("store deal pipeline cleanup: updating deal pipeline: %w", err)
+			return false, xerrors.Errorf("store deal failure: updating MK1.2 deal: %w", err)
+		}
+		if n == 0 {
+			n, err = tx.Exec(`UPDATE market_direct_deals SET error = $1 WHERE uuid = $2`, reason, deal)
```
ZenGround0, line 387: Was this rearrangement into 2 statements in a TX just for improved readability?  It's easier for me to understand but I want to know if I'm missing a functional reason for the refactor.

ZenGround0: I guess the real change was glueing together these operations with the updatePipeline branch?  And this was a classic gluing together two seperate Execs that should commit together into one Tx.

Did you find this during testing?

tasks/storage-market/storage_market_test.go
```diff
@@ -0,0 +1,179 @@
...
+		})
+	}
+
+	t.Run("final assignment rechecks expiry", func(t *testing.T) {
+		id := "expiry-test-final-assignment"
+		insertExpiryTestDeal(t, ctx, db, id, 2000, head+sealDuration-1, false)
+		ingester := &expiryTestIngester{expectedSealDuration: sealDuration}
+		market := &CurioStorageDealMarket{db: db, api: api, pin: ingester}
+
+		err := market.processMk12Deal(ctx, MK12Pipeline{
```
ZenGround0, line 107: I don't understand why this test is not just in the market package.  Why put it here ? Was this test added on after bad error handling was found in task_commp?  If so I don't know if that justifies putting a test here.

ZenGround0 (approved):
Overall it makes sense.  I'm curious about the story behind the commp task change and if tests are in the right place.

ZenGround0:
This looks pretty close to merging so just bumping this one in your inbox @LexLuthr.

ZenGround0:
@LexLuthr ping on getting this landed

tasks/storage-market/storage_market_test.go
LexLuthr: This was because AI wrote the test. When code was initially written check expiry was in commP. So it just added test file there instead of moving the code to a generic file. Fixed it

tasks/storage-market/task_commp.go
LexLuthr: Yes. This should have always been a tx for atomicity.

PR #1414, opened 2026-08-10
send-durability

### LINKED ISSUE #1413: pdp: eth sender has no crash recovery - orphaned lock or pending-send rows permanently deadlock all sends
Two production incidents on v1.28.3 within four days, same root cause: the eth sender persists per-send state in `message_send_eth_locks` and `message_sends_eth`, but nothing reconciles that state after a crash. A process killed mid-send leaves rows that block the wallet's entire nonce lane, restarts do not recover, and the only remedy is manual SQL against the Curio DB.

**Case 1 - orphaned sender lock (calibnet, upgrade-day restart churn)**

A SendTransaction task was killed between claiming the `message_send_eth_locks` row and its deferred release (`SendTaskETH.Do`, claim loop at sender_eth.go:88-117). The lock row outlived the task. Every subsequent send task then spun in the 100ms retry loop waiting for the lock, and Prove/ProvPeriod tasks hung in the inline `SenderETH.send` poll loop (sender_eth.go:529), pinning every task slot. A restart produced a brief burst then re-wedged within minutes - the lock row persists in the DB, so the fresh process inherits the same deadlock. ~1,300 tasks backlogged before recovery.

Recovery: `DELETE FROM message_send_eth_locks WHERE task_id = <id>` after verifying the holding task no longer exists in harmony_task. Backlog drained within seconds.

**Case 2 - orphaned pending send row (mainnet, endorsed SP, network outage)**

A send was interrupted after nonce assignment but before an outcome was recorded, leaving a `message_sends_eth` row with `send_success` NULL at the wallet's head nonce. The unique index `message_sends_eth_success_index` on `(from_address, nonce) WHERE send_success IS NOT FALSE` then collided with every new send at the nonce-assignment update (sender_eth.go:184):

```
updating db record: ERROR: duplicate key value violates unique constraint "message_sends_eth_success_index" (SQLSTATE 23505)
```

Result: zero messages sent for 3 days. The state survived multiple restarts, a version upgrade and a wallet top-up, with a continuous `Do() returned error` storm in the log. eth_getTransactionCount confirmed the stuck row's tx never mined.

Recovery: `UPDATE message_sends_eth SET send_success = FALSE WHERE send_task_id = <id> AND send_success IS NULL`. Sends resumed within a minute; 20+ mined in the following hour.

**The gap**

The previously-signed resume branch (sender_eth.go:190+) only helps while the original task row survives. When the task is gone, neither the lock nor the pending row is checked again. A startup or periodic reconciliation could restore the invariant: drop lock rows whose task_id no longer exists in harmony_task, and for pending send rows compare against the chain nonce - below it means confirmed or replaced, at or above it with no live task means safe to mark failed.

Both cases on v1.28.3. Happy to share full logs and the recovery transcripts.

snadrus:
Fixes Case 1 of #1413

**Portion 1:  Task design**
- Introduces an "uninterruptible" harmony_task flag to prevent preemption for send_task & send_task_eth
(This is basically "TimeSensitive" without the pre-emption line-jumping. 

**Portion 2:  Unlock exceptions**
- Unlock via TTL+harmony_machine liveness gone
- Unlock irrespective of the CTX (so unlock on shutdown). 
- Unlock for self task_id (in a restart case). 
- Unlock when holder task_id is not running.

### DIFF
--- a/tasks/message/send_lock.go
+++ b/tasks/message/send_lock.go
@@ -1,0 +1,122 @@
+package message
+
+import (
+	"context"
+	"database/sql"
+	"time"
+
+	"github.com/filecoin-project/curio/harmony/harmonydb"
+	"github.com/filecoin-project/curio/harmony/harmonytask"
+	"github.com/filecoin-project/curio/harmony/resources"
+)
+
+const sendLockReleaseTimeout = 15 * time.Second
+
+// SendLockStaleTTL is how old a lock must be before it may be stolen from a
+// holder whose harmony_task row still exists but has no live owning machine.
+// Missing harmony_task rows may be stolen immediately. Overridable in tests.
+var SendLockStaleTTL = 2 * time.Minute
+
+func releaseLockCtx() (context.Context, context.CancelFunc) {
+	// Must not use the task ctx: preemption/cancel would make DELETE fail and
+	// leave an orphan lock that wedges all subsequent sends for the address.
+	return context.WithTimeout(context.Background(), sendLockReleaseTimeout)
+}
+
+func releaseMessageSendLock(db *harmonydb.DB, fromKey string, taskID harmonytask.TaskID) error {
+	ctx, cancel := releaseLockCtx()
+	defer cancel()
+	_, err := db.Exec(ctx, `
+		DELETE FROM message_send_locks WHERE from_key = $1 AND task_id = $2`, fromKey, taskID)
+	return err
+}
+
+func releaseEthSendLock(db *harmonydb.DB, fromAddress string, taskID harmonytask.TaskID) error {
+	ctx, cancel := releaseLockCtx()
+	defer cancel()
+	_, err := db.Exec(ctx, `
+		DELETE FROM message_send_eth_locks WHERE from_address = $1 AND task_id = $2`, fromAddress, taskID)
+	return err
+}
+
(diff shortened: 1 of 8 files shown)


tasks/message/send_lock.go
```diff
@@ -0,0 +1,122 @@
...
+)
+
+const sendLockReleaseTimeout = 15 * time.Second
+
+// SendLockStaleTTL is how old a lock must be before it may be stolen from a
+// holder whose harmony_task row still exists but has no live owning machine.
+// Missing harmony_task rows may be stolen immediately. Overridable in tests.
+var SendLockStaleTTL = 2 * time.Minute
+
+func releaseLockCtx() (context.Context, context.CancelFunc) {
```
ZenGround0, line 20: Question for my understanding:

Why not use the ctx from the callers?  Since they are made uninterruptible Preempt should not be reachable on their context and so their ctx's should satisfy the same properties listed here.

ZenGround0 (approved):
Thanks for jumping on the quick fix

ZenGround0 (commented):
@LexLuthr has some dissenting comments that you should hear before we merge this

tasks/message/send_lock.go
LexLuthr: This part is removed in https://github.com/filecoin-project/curio/pull/1417

Issue #1386, opened 2026-07-30
PDP: renable replace by fee once curio and synapse integration is better configured


ZenGround0:
This is a placeholder for removing the `replaceByFeeEnabled` const and check preventing MessageReplacer task from doing anything.

LexLuthr:
When enabling back the replacement subsystem, we must update the Fee calculation method to same as eth_sender

PR #1415, opened 2026-08-10
replace by fee

snadrus:
Here's what's needed to make ReplaceByFee friendly to Synapse:
- on-off switch still defaults to off
- communicates a DIFFERENT txHash that's confirmed:
 -- so as not to break old Synapse's retries or confirmation. (it will break old Synapse's follow-up chain verifies, if any). 
 -- to verify the system is NOT broken but that 2 hashes are associated. 

Associated Synapse client PR will link here in a bit.

This PR (when enabled by config setting) should enable E2E testing of the SynapseSDK feature: https://github.com/FilOzone/synapse-sdk/pull/912 which we want to be disbursed widely before we finally enable this PR by default (and remove the option).


tasks/message/replacer.go
```diff
@@ -41,13 +45,10 @@ type replaceTrigger struct {
...
 
-// Removal tracked in https://github.com/filecoin-project/curio/issues/1386 complete
-// Message replacement only safe to enable once synapse sdk caller can be made aware of
-// replacement flow with small curio api redesign.
-const replaceByFeeEnabled = false
-
+// Enabled when Fees.ReplaceByFee is true. Safe once Synapse (and other clients)
+// treat Location hashes as wait keys and read confirmedTxHash from PDP status.
 func NewMessageReplacer(ctx context.Context, cfg ReplacerConfig) error {
-	if !replaceByFeeEnabled {
```
ZenGround0, line 50: Nice this closes #1386

tasks/message/replace_eth.go
```diff
@@ -452,11 +457,18 @@ func (t *ethMessageReplacer) prepareFeeReplacementEthMessage(ctx context.Context
 		return nil, false, fmt.Errorf("base fee is not available for %s nonce %d", from.Hex(), candidate.Nonce)
 	}
 
+	// Match eth_sender fee construction: gasFeeCap = 2*baseFee + tip, with a
+	// recent FeeHistory max used as a base-fee floor.
```
ZenGround0, line 461: I'm pretty sure this is coming from / tracked by [this comment ](https://github.com/filecoin-project/curio/issues/1386#issuecomment-5141480057) ?

ZenGround0 (approved):

ZenGround0:
I might be misremembering but I think our plan was to merge after latest release so we should be good to go.  I'm ok keeping this unmerged for a few more days in case we need to do another quick follow on release.

PR #1416, opened 2026-08-11
pdp always return Allow-Origin *

beck-8:
1. Correct the original wrong gui cors position
2. pdp path always returns `"Access-Control-Allow-Origin", "*"`
3. Current limiting settings allow determination of LAN IP
>3 This is what I just noticed. Because the reverse proxy settings of the SP are not necessarily placed on the local machine, I modified them together here.

---
context:
https://filecoinproject.slack.com/archives/C095WFA0QK1/p1786423884375819?thread_ts=1786409312.417809&cid=C095WFA0QK1
image

### DIFF
--- a/cuhttp/clientip/clientip.go
+++ b/cuhttp/clientip/clientip.go
@@ -9,18 +9,18 @@ import (
 	"github.com/go-chi/httprate"
 )
 
-// Middleware resolves the client IP from the TCP peer by default. A loopback
-// peer is treated as one trusted reverse-proxy hop, matching Curio's documented
-// same-host proxy setup. The proxy must overwrite or append X-Forwarded-For.
+// Middleware resolves the client IP from the TCP peer. A loopback or
+// private-network peer is trusted for one hop of X-Forwarded-For; the
+// proxy must overwrite or append that header.
 func Middleware(next http.Handler) http.Handler {
 	direct := middleware.ClientIPFromRemoteAddr(next)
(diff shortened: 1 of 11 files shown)


LexLuthr (changes requested):
I don't think we should make the entire public HTTP API universally accessible from browsers. Curio is infrastructure software, not a public web API. The default security model should be least privilege, with explicit allowlists for trusted frontends.

Trusting every RFC1918 address as a reverse proxy is a security regression. Any host on the private network can now influence the derived client IP through X-Forwarded-For. Trusted proxies should be explicitly configured rather than inferred from IsPrivate(). This needs alignment within the team before deciding on the policy. 

I am not sold 100% on moving the config either. Moving config which is in use has some impact. None of that seems to be tested or covered in this PR.

LexLuthr:
@snadrus Hey! We need to fix the relax the reverse proxy IP but this PR seems to be doing much more than that. I don't agree with this approach due to security concerns. Can you please review the changes. Can you also put your thoughts on IsPrivate vs allowing specific IP via config to be trusted reverse proxy?

snadrus:
There are a few moving parts here, so I broke out the most urgent piece: 
https://github.com/filecoin-project/curio/pull/1419 is the Access Control Allow * fix which fixes the bug.

ZenGround0:
This may be good to close now that #1419 is merged.  But if there are separate pieces from this we still want feel free to keep it open @beck-8

PR #1417, opened 2026-08-11
fix: make sender lock finalization durable

### LINKED ISSUE #1413: pdp: eth sender has no crash recovery - orphaned lock or pending-send rows permanently deadlock all sends

LexLuthr:
## Problem

Message send results and sender locks were persisted separately.

For ETH sends, a process could record a terminal result and crash before the deferred lock deletion. On retry, the finalized-row fast path returned immediately, Harmony removed the task, and the sender lock remained permanently orphaned.

## Changes

- Atomically persist the send result and delete the sender lock in one database transaction.
- Apply the same finalization invariant to both Filecoin and ETH senders.
- Keep chain submission outside the database transaction.
- Allow lock reacquisition only by the same Harmony task ID.
- Remove the stale-lock takeover helper and its standalone tests.
- Add foreign keys from sender locks to `harmony_task` with `ON DELETE CASCADE`.
- Remove existing orphan lock rows before installing the constraints.
- Add supporting indexes and a downgrade migration.
- Add integration coverage for:
  - successful and failed send finalization;
  - pre-send error cleanup;
  - finalization rollback and same-task retry;
  - stored signed-message reuse;
  - contention and sender serialization;
  - rejection of locks without a Harmony task;
  - lock cleanup when a task is deleted;
  - lock survival when a dead machine is removed;
  - recovery of the legacy finalized ETH lock case.


snadrus:
This is really a brilliant fix to the problem

snadrus (approved):

PR #1424, opened 2026-08-12
add pdp balance check

beck-8:


beck-8:
I accidentally saw that there is a problem with ping health.
1. Almost no errors are displayed
2. No recovery after error

Solve this part together in the last commit

LexLuthr (dismissed):
Let's not create a whole new config and PDP specific params. You can just extend the wallet balance check to also check for PDP.

alertmanager/alerts.go
```diff
@@ -146,11 +146,49 @@ func balanceCheck(al *alerts) {
...
+
 	if ret.String() != "" {
 		al.alertMap[Name].alertString = ret.String()
 	}
 }
 
+// appendPDPWalletBalance checks the PDP wallet (used to pay gas for PDP on-chain
+// transactions) against the same MinimumWalletBalance threshold as the other
+// wallets.
+func appendPDPWalletBalance(al *alerts, ret *strings.Builder) {
```
ZenGround0: nit: I think your original `pdpBalanceCheck` name was better.  Just going off the basic principle that development or caller path dependent names like `append` are less good than callee descriptive names.

ZenGround0 (approved):

beck-8:
I think we should have no problem here. Who just needs to agree?

PR #1426, opened 2026-08-13
ipni: replace ad-fetch tracking with indexer sync-status confirmation

### LINKED ISSUE #1405: Integrate IPNI /sync/status/ad/{adCid} to track confirmed advertisement indexing
### Checklist

- [x] This is **not** a question or a support request. If you have any Curio related questions, please ask in the [discussion forum](https://github.com/filecoin-project/curio/discussions).
- [x] This is **not** a new feature request. If it is, please file a [feature request](https://github.com/filecoin-project/curio/issues/new?assignees=&labels=need%2Ftriage%2Ckind%2Ffeature&template=feature_request.yml) instead.
- [x] I **have** searched on the [issue tracker](https://github.com/filecoin-project/curio/issues) and the [discussion forum](https://github.com/filecoin-project/curio/discussions), and there is no existing related issue or discussion.
- [x] I am running the [`Latest release`](https://github.com/filecoin-project/curio/releases), or the most recent RC(release candidate) for the upcoming release or the dev branch(master), or have an issue updating to any of these.

### Curio component

- [ ] PoRep Pipeline
- [ ] Snap Deals
- [ ] Market
- [x] IPNI provider
- [ ] Curio UI
- [ ] Other

### Improvement Suggestion

Clients (dealbot, filecoin-pin, FOC upload flows) need to know when it is safe to query cid.contact/cid/{cid} for content uploaded to Curio. Querying too early returns a 404 that seeds IPNI's negative cache, which can delay discoverability by ~20+ minutes even after the content is fully indexed (see [filecoin-pin](https://github.com/filecoin-project/filecoin-pin/blob/master/documentation/content-routing-faq.md#how-long-does-an-ipni-indexer-cache-results)
 [content-routing FAQ](https://github.com/filecoin-project/filecoin-pin/blob/master/documentation/content-routing-faq.md#how-long-does-an-ipni-indexer-cache-results)

Curio currently cannot answer "has IPNI actually indexed this ad?":
 - GET /pdp/piece/{pieceCid}/status reports retrieved based on ipni_ad_fetches — rows inserted when the indexer fetches the ad from our provider endpoint (market/ipni/ipni-provider/ipni-provider.go, logPDPFetch). A fetch means the indexer pulled the ad, not that it processed/indexed it. It's also unreliable in  the best case (fetches can be missed, retried, or served from intermediate state). 
 - The Publishing head for provider announce log only means we told IPNI a new head exists; ingestion is entirely IPNI-driven after that.
 -  this means  tracking and benchmarking index latency also very challenging; 

 storetheindex now exposes  ([ipni/storetheindex#2859](https://github.com/ipni/storetheindex/issues/2859) [PR #2861](https://github.com/ipni/storetheindex/pull/2861) deployed on cid.contact:

 ```
   $ curl -s https://cid.contact/sync/status/ad/baguqeera... | jq .
   {
     "Ad": "baguqeera...",
     "Indexed": true
   }
 ```

Indexed: true means the indexer has fully processed that advertisement. With this, Curio can gate its "IPNI-ready" signal on confirmed ingestion, and clients polling the piece status endpoint never need to touch /cid/{cid} before it's positive; 

 Proposal (to be reviewed and corrected by @LexLuthr )
 1. a lightweight periodic task that, for each provider (peer ID) with unconfirmed announced ads, queries {ServiceURL}/sync/status/ad/{adCid} and records confirmation. 
 2.l record confirmation to allow us tracking ad and piece index status and latency; 
 3. Surface in GET /pdp/piece/{pieceCid}/status (pdp/handlers.go, handleGetPieceStatus): add synced / syncedAt fields and a new terminal status: pending → indexing → creating_ad → announced  → synced
 4. Metrics/logging: log ad-confirmed events with timestamps so we can measure announce→indexed latency end to end (announce time is already logged with `log.Infow("Publishing head for provider", "provider", provider, "cid", c.String())` ). 

Others
- I would love to have some benchmarking/metric  from a curio ipni publisher that what is the `time to availability` is like, meaning how long does a cid can be looked up from ipni since we have published the ad; 
- this will help us evaluate if IPNI is performant enough for us to drop customized  pin.contact

beck-8:
Closes https://github.com/filecoin-project/curio/issues/1405

## Summary

Previously, we judged whether a piece was reachable via IPNI by checking if anyone had *fetched* the ad — that signal isn't reliable. Now we ask storetheindex's new endpoint directly and get a real confirmation that it's actually been processed.

The core idea: an indexer processes a provider's ad chain strictly in order, so once it's confirmed the latest ad, everything before it is confirmed too. `order_number` is just the existing auto-increment column on that table, not something new — we're using it as a watermark. Each provider keeps two numbers in memory: how far we've announced, and how far we've confirmed synced. A piece counts as synced once its own ad's number is at or before the confirmed one. No need to track status per-ad, no new table.

Breaking change: the piece status endpoint replaces `retrieved`/`retrievedAt` with `synced`/`syncedAt`, no compat alias.

### DIFF
--- a/documentation/en/configuration/metrics-reference.md
+++ b/documentation/en/configuration/metrics-reference.md
@@ -183,6 +183,7 @@ This document lists Prometheus metrics exported by Curio using the exact metric 
 | `curio_http_response_status_count` | counter | `status_code`, `path`, `method` | Counter of HTTP response status codes |
 | `curio_ipni_announce_attempts_total` | counter | `provider`, `result` | Total number of IPNI direct announce attempts. |
 | `curio_ipni_announce_http_roundtrip_milliseconds` | histogram | `provider`, `status` | Duration of outbound IPNI announce HTTP round trips in milliseconds. |
+| `curio_ipni_announce_to_indexed_latency_milliseconds` | histogram | `provider`, `provider_type`, `service` | Duration from a successful head announce to confirmed indexing by an indexer service. |
 | `curio_ipni_entry_cache_hit_wait_milliseconds` | histogram | `request`, `origin`, `state` | Duration callers wait after hitting the IPNI entry cache. |
 | `curio_ipni_entry_cache_hits_total` | counter | `request`, `origin`, `state` | Total number of IPNI entry cache hits by request type, cache origin, and readiness state. |
 | `curio_ipni_entry_cache_lookups_total` | counter | `request`, `result` | Total number of IPNI entry cache lookups. |
--- a/documentation/en/configuration/metrics-reference.md
+++ b/documentation/en/configuration/metrics-reference.md
@@ -183,7 +183,7 @@ This document lists Prometheus metrics exported by Curio using the exact metric 
 | `curio_http_response_status_count` | counter | `status_code`, `path`, `method` | Counter of HTTP response status codes |
 | `curio_ipni_announce_attempts_total` | counter | `provider`, `result` | Total number of IPNI direct announce attempts. |
 | `curio_ipni_announce_http_roundtrip_milliseconds` | histogram | `provider`, `status` | Duration of outbound IPNI announce HTTP round trips in milliseconds. |
-| `curio_ipni_announce_to_indexed_latency_milliseconds` | histogram | `provider`, `provider_type`, `service` | Duration from a successful head announce to confirmed indexing by an indexer service. |
+| `curio_ipni_announce_to_indexed_latency_seconds` | histogram | `provider`, `provider_type`, `service` | Duration from a successful head announce to confirmed indexing by an indexer service. |
 | `curio_ipni_entry_cache_hit_wait_milliseconds` | histogram | `request`, `origin`, `state` | Duration callers wait after hitting the IPNI entry cache. |
 | `curio_ipni_entry_cache_hits_total` | counter | `request`, `origin`, `state` | Total number of IPNI entry cache hits by request type, cache origin, and readiness state. |
 | `curio_ipni_entry_cache_lookups_total` | counter | `request`, `result` | Total number of IPNI entry cache lookups. |
--- a/harmony/harmonydb/downgrade/20260813-drop-ipni-ad-fetches.sql
+++ b/harmony/harmonydb/downgrade/20260813-drop-ipni-ad-fetches.sql
@@ -1,0 +1,6 @@
+CREATE TABLE IF NOT EXISTS ipni_ad_fetches (
+    ad_cid TEXT NOT NULL,
+    fetched_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
+);
+
+CREATE INDEX IF NOT EXISTS ipni_ad_fetches_ad_cid_time ON ipni_ad_fetches(ad_cid, fetched_at DESC);
--- a/harmony/harmonydb/downgrade/20260814-ipni-provider-order-number.sql
+++ b/harmony/harmonydb/downgrade/20260814-ipni-provider-order-number.sql
@@ -1,0 +1,53 @@
+CREATE OR REPLACE FUNCTION insert_ad_and_update_head_checked(
+    _ad_cid TEXT,
+    _context_id BYTEA,
+    _metadata BYTEA,
+    _piece_cid_v2 TEXT,
+    _piece_cid TEXT,
+    _piece_size BIGINT,
+    _is_rm BOOLEAN,
+    _provider TEXT,
+    _addresses TEXT,
+    _signature BYTEA,
+    _entries TEXT,
+    _expected_previous TEXT
+) RETURNS BOOLEAN AS $$
+DECLARE
+    _current_head TEXT;
+    _rows BIGINT;
+BEGIN
+    SELECT head INTO _current_head
+    FROM ipni_head
+    WHERE provider = _provider;
+
+    IF _current_head IS DISTINCT FROM _expected_previous THEN
+        RETURN FALSE;
+    END IF;
+
+    INSERT INTO ipni (
+        ad_cid, context_id, metadata, is_rm, previous, provider, addresses,
+        signature, entries, piece_cid_v2, piece_cid, piece_size
+    )
+    VALUES (
+        _ad_cid, _context_id, _metadata, _is_rm, _expected_previous, _provider, _addresses,
+        _signature, _entries, _piece_cid_v2, _piece_cid, _piece_size
+    );
+
+    IF _expected_previous IS NULL THEN
+        INSERT INTO ipni_head (provider, head)
+        VALUES (_provider, _ad_cid)
+        ON CONFLICT (provider) DO NOTHING;
+    ELSE
+        UPDATE ipni_head
+        SET head = _ad_cid
+        WHERE provider = _provider
+          AND head IS NOT DISTINCT FROM _expected_previous;
+    END IF;
+
+    GET DIAGNOSTICS _rows = ROW_COUNT;
+    RETURN _rows = 1;
+END;
+$$ LANGUAGE plpgsql;
+
+ALTER TABLE ipni_head DROP COLUMN IF EXISTS head_order_number;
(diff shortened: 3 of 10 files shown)


documentation/en/configuration/metrics-reference.md
```diff
@@ -183,6 +183,7 @@ This document lists Prometheus metrics exported by Curio using the exact metric
 | `curio_http_response_status_count` | counter | `status_code`, `path`, `method` | Counter of HTTP response status codes |
 | `curio_ipni_announce_attempts_total` | counter | `provider`, `result` | Total number of IPNI direct announce attempts. |
 | `curio_ipni_announce_http_roundtrip_milliseconds` | histogram | `provider`, `status` | Duration of outbound IPNI announce HTTP round trips in milliseconds. |
+| `curio_ipni_announce_to_indexed_latency_milliseconds` | histogram | `provider`, `provider_type`, `service` | Duration from a successful head announce to confirmed indexing by an indexer service. |
```
LexLuthr: millisecond is probably a bad counter. Seconds would be better because this can go for hours in worst cases.

LexLuthr (changes requested):
I am not sure I understand the approach here. Can you please add summary to the PR. I cannot figure out a good reason behind the whole ordered thing.

beck-8:
updated
Use `order_number` to identify this ad, for easy comparison of statuses.

LexLuthr:
@beck-8 Order_number is very much useless due to how Postgres auto increment caching works. I would not recommend it to judge the real order of ad creation.

LexLuthr:
More context: The new sync endpoint is only deployed on cid.contact so trying to query it on pin indexer is useless and will generate unnecessary errors. The aim is also to fix the delays and issue around cid.contact so we can eventually remove the pin indexer.

The whole order system it to simply allow you to find a metric number. While you are thinking in right direction, you are making it too complicated by depending on order number. In doing so you are doing an rounding of actual time taken from publish to sync. 

Think about this - We get to publish, before publishing we query new head, we know old published head. We get the ads between them with a query (we can add indexes as required to speed it up). Now we have a list, keep it in memory, publish, save published time from that list. Then simply check for oldest ad, if it is synced, we know how long it took to sync the first one. Moment we see 1 synced, we switch to next one. This basically allows us to get exact sync timing or closest possible to it without asking indexer the exact time stamp. This approach requires a lot more memory. We can easily move it into DB if memory becomes an issue later. For now we can do in memory with some sensible limits.

beck-8:
https://github.com/filecoin-project/curio/blob/550f2ee0aadd3491da2bc71df13673075a803ccb/harmony/harmonydb/sql/20250505-market-mk20.sql#L156

I'm taking another look. I originally thought these would be safe autoincrements.

LexLuthr:
> https://github.com/filecoin-project/curio/blob/550f2ee0aadd3491da2bc71df13673075a803ccb/harmony/harmonydb/sql/20250505-market-mk20.sql#L156
> 
> I'm taking another look. I originally thought these would be safe autoincrements.

There is caching (100 number per client node) in there. While your guess would be correct for single node but it will fail for multi node clusters. Since IPNI is a common subsystem, it will be impacted on PoRep nodes. So, you need to get creative.

beck-8:
@LexLuthr  I've thought about the method you mentioned. There are a few questions.
1. It is impossible to know the time when the previously backlogged data will be pushed after restarting.
2. The status of piece status itself is not easy to handle.
3. It is meaningless to check the ADs one by one in this batch of lists. Assuming that there are 20 ADs between this head and the old data, then these 20 ADs must send http requests to query. Even if the other party has processed all the data in 1 second, it may be waiting due to network reasons. Whether to retry here or whether to perform concurrency is a question. (There may be one point to be clarified here. The other party's implementation only returns this cid, whether it is indexd or not. There is no timestamp)

**So I am more inclined to my current solution. Regarding the question you raised, I might want AI to help me see how to build a really safe self-increasing ID (not sure if there are any other concerns here)**

> More context: The new sync endpoint is only deployed on cid.contact so trying to query it on pin indexer is useless and will

Yes. I understand. The logic here will only process the first service that returns the result correctly. And cid.contact is ranked first.

LexLuthr:
> @LexLuthr I've thought about the method you mentioned. There are a few questions.
> 
> 1. It is impossible to know the time when the previously backlogged data will be pushed after restarting.

Restart is one case not handled by memory. For that we can switch to durable storage like DB but I don't agree that it is urgent. I don't want to complicate things for a metric. 

> 2. The status of piece status itself is not easy to handle.

It is. We just change the json and let one application using it know about the change. 

> 3. It is meaningless to check the ADs one by one in this batch of lists. Assuming that there are 20 ADs between this head and the old data, then these 20 ADs must send http requests to query. Even if the other party has processed all the data in 1 second, it may be waiting due to network reasons. Whether to retry here or whether to perform concurrency is a question. (There may be one point to be clarified here. The other party's implementation only returns this cid, whether it is indexd or not. There is no timestamp)

This is incorrect assumption. Ads the not processed that fast. Practically, you will see some diff between each ad sync. I would suggest reading my original explanation again. 

> 
> **So I am more inclined to my current solution. Regarding the question you raised, I might want AI to help me see how to build a really safe self-increasing ID (not sure if there are any other concerns here)**

There is no fixing the IDs unless you somehow want to walk full ad chain and correct it. Even then you will run into inconsistencies which are impossible to resolve due to a bug we had in the implementation. That is a huge undertaking for the sake of a metric. I don't find that worth the effort.

> 
> > More context: The new sync endpoint is only deployed on cid.contact so trying to query it on pin indexer is useless and will
> 
> Yes. I understand. The logic here will only process the first service that returns the result correctly. And cid.contact is ranked first.

beck-8:
Just fixed a previously missed issue: if there are many ADs or the opponent's processing speed drops, they will keep checking status with the latest head, causing status checks to fail or be delayed.

BigLep:
Jotting down some of my notes from the 2026-08-20 standup breakout:
- We should track when we announce an advertisement to IPNI for a given pieceCId. This enables clients like dealbot to know when Curio "did it's job" of announcing the advertisement.
  - It's fine for this to live in memory. 
- We don't need to track the time when IPNI reaches back to Curio. This is fuzzy, and we can't guarantee that that pull from IPNI succeeded or that it pertains to the advertisement we advertisement.
  - This adds complexity and isn't tight.
  - We absolutely should debug log all interactions with IPNI so a motivated party can do analysis but we  won't publish a metric.
- Curio won't proactive poll IPNI's `/sync/status/ad/{adCid}` endpoint for all pieces. This IPNI endpoint will be queried in realtime when a Curio client calls Curio's `GET /pdp/piece/{pieceCid}/status`. 
- We are fine not to track when "IPNI indexed" the content of a given pieceCid. We would only be guessing, and our guess is based on a user asking us to go do a realtime check. 

Bottomline:
- Keep it simple. Make sure Curio's `GET /pdp/piece/{pieceCid}/status` returns whether IPNI has synced/indexed the relevant pieceCid.
- Debug log IPNI interactions so a motivated party can inspect more of what's going on.

Hopefully the above simplifies the PR.

beck-8:
> We don't need to track the time when IPNI reaches back to Curio.

It is no longer there and this part has been deleted.

> will be queried in realtime

We have reached an agreement on this, right? The time-consuming polling of this interface will increase (increase the round-trip delay between curio and IPNI).

> We should track when we announce an advertisement to IPNI for a given pieceCId

This is not simplified. If you rely on memory alone, the judgment itself will be complicated. If the status is lost after restarting, the status will become never push.

I will reconsider how to proceed based on your discussion later.

BigLep:
@beck-8 : I'll @LexLuthr answer authoritatively here.  I'll comment in https://github.com/filecoin-project/curio/pull/1450

LexLuthr:
@beck-8 Let's drop the metrics completely. We will redo the metrics based on https://github.com/ipni/storetheindex/pull/2898. Once clean up is done I will do the final review and we can merge.

BigLep:
Replaced by https://github.com/filecoin-project/curio/pull/1450

PR #1427, opened 2026-08-13
Analyze where necessary

### LINKED ISSUE #1315: pdp: large tables run with no statistics (autoanalyze disabled), causing query plan regressions at scale
On a mainnet PDP node, several large tables have no planner statistics, and `last_autoanalyze` is NULL on every table in the `curio` schema. YugabyteDB is not auto-analyzing, so the planner mis-plans queries on these tables (full scans/sorts instead of index probes).

## Evidence (live mainnet node)

- `pdp_piece_pull_items` and `ipni_ad_fetches` had no statistics at all: `reltuples = -1`, zero analyzed columns, `last_analyze` NULL.
- `last_autoanalyze` is NULL across all tables.

## Measured impact

- Piece-status query (`pdp/handlers.go`): with no stats the `ipni_ad_fetches` join ran a full seq scan of 4.65M rows plus a 383 MB external disk sort (25.6 s). `ANALYZE` alone removed the sort (down to 16.7 s). (The query is also rewritten in #1314.)
- `completeAlreadyParkedItems` (`tasks/pdpv0/task_pull_piece.go`): the planner hash-builds all ~1.8M `parked_pieces` rows because it does not know `pdp_piece_pull_items` holds only ~240 rows.

Running `ANALYZE` manually fixes the plans, but with autoanalyze off the statistics go stale again as the tables are written.

## Suggested direction (needs design)

- Enable/repair YugabyteDB autoanalyze for the `curio` schema, or
- have curio run a periodic / post-bulk-load `ANALYZE` on its large tables.

Relates to #1314 (query rewrites that work around the worst cases).

snadrus:
24h max frequency: 
    if the stats table has values > 10% then lets run analyze on the table. 

Fixes #1315

### DIFF
--- a/cmd/curio/tasks/tasks.go
+++ b/cmd/curio/tasks/tasks.go
@@ -44,6 +44,7 @@ import (
 	"github.com/filecoin-project/curio/market/libp2p"
 	"github.com/filecoin-project/curio/pdpnode"
 	"github.com/filecoin-project/curio/tasks/balancemgr"
+	"github.com/filecoin-project/curio/tasks/dbmaint"
 	"github.com/filecoin-project/curio/tasks/expmgr"
 	"github.com/filecoin-project/curio/tasks/f3"
 	"github.com/filecoin-project/curio/tasks/gc"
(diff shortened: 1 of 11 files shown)


ZenGround0 (approved):
way cool

PR #1430, opened 2026-08-14
pdp: serve long-term parked pieces with no deal or pdp ref from the cached reader

TippyFlitsUK:
**Setup:** mainnet PDP SPs holding Storacha-migration first copies as long-term parked aggregates (~226k+ pieces), now sourcing second-copy replication pulls. Running v1.28.4.

Retrieving a parked piece that has neither a market deal nor a `pdp_piecerefs` row returns `ErrNoDeal`, so `/piece` requests 404 while the bytes sit complete in the piece park. Migration aggregates are parked `long_term = TRUE` by design with no pdp ref until claimed - stock Curio cannot serve them at all. Hit this on main2.ezpdpz.net on 2026-07-29 (every retrieval of a parked complete aggregate 404'd), and these pieces are exactly what the second-copy repair campaign pulls from a first-copy SP, so on stock the campaign has no source.

The fix adds a last-resort fallback in `getPieceReaderFromMarketPieceDeal`: when there is no market deal and no pdp ref, look the piece up in `parked_pieces` with `long_term = TRUE AND complete = TRUE` and serve it via the piece park reader. `ErrNoDeal` is unchanged when the fallback finds nothing.

Running in production on both my mainnet SPs since late July, carried locally across v1.28.2-rc3, v1.28.3 and v1.28.4 - currently serving the live second-copy pulls to two other SPs. Happy to adjust or add coverage if wanted.

### DIFF
--- a/lib/cachedreader/cachedreader.go
+++ b/lib/cachedreader/cachedreader.go
@@ -234,7 +234,11 @@ func (cpr *CachedPieceReader) getPieceReaderFromMarketPieceDeal(ctx context.Cont
 				return nil, 0, fmt.Errorf("failed to query pdp_piecerefs for piece cid %s: %w", pieceCid, err)
 			}
 			if !isPDP {
-				return nil, 0, fmt.Errorf("piece cid %s: %w", pieceCid, ErrNoDeal)
+				reader, rawSize, err := cpr.getPieceReaderFromLongTermPark(ctx, pieceCid)
+				if err != nil {
+					return nil, 0, fmt.Errorf("piece cid %s: %w", pieceCid, ErrNoDeal)
+				}
+				return reader, rawSize, nil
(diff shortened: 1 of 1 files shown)


LexLuthr:
We are basically opening another security hole here.

ZenGround0:
Following onto what @LexLuthr is saying I think this is not the best fix.  I want to first understand what pieces are not being tracked by a pdp_piecref and why.  Our first recourse should be to track these pieces correctly.  If there is a reason the code is explicitly not tracking them we need to ammend this wholistically.  If this is a matter of troubled database state resulting from the understandbly difficult storacha migration then a better option is probably to repair your db.  Is there an issue describing the exact problem and if we know why these pieces are not tracked by piecrefs?

PR #1439, opened 2026-08-17
Make enqueued removals still more conservative

ZenGround0:
Due to reports that [our existing conservative clamping of removals is not conservative enough](https://github.com/FilOzone/pdp/issues/283) we need to reduce this constant to prevent blocking datasets.


snadrus (approved):

snadrus:
@ZenGround0 should we stand behind this change?

PR #1444, opened 2026-08-19
fix: upload-to-first-download latency

### LINKED ISSUE #1399: PDP: Optimize upload-to-first-download latency
The latency is coming from some old design elements.
1. Notify task - This task is used to mark the piece ready for PDP and thus enable the retrievals for it. 
2. Double upload in classic upload path - we stream the bytes to a temporary upload space and then park it later. This basically means double write. Ideally on a single machine this is copy on same disk so it should be fast but this should still be removed.

## Current status
1. Next Curio release will have fast harmony scheduler. It should reduce the latency from scheduling notify task. We can test again after release. If the improvement in not there, we get rid of the notify task altogether.
2. Once Notify task has been removed, if latency is still significant, we must rework the classic upload API to directly write into the piece store.

## Retrieval
Hot latency is 100-200 ms. This is acceptable and cold latency after first upload should ideally come closer to this. Cold latency when retrievable is around 670.2 ms. We can test if there is perf improvment by turning off the mime sniffer. If there is then perhaps, we need to rethink mime sniffer side of things.

LexLuthr:
fixes #1399


LexLuthr:
## Latest test summary — Provider 23
The post-upload availability problem appears fixed.

- 168/168 uploads succeeded.
- Every immediate retrieval succeeded on attempt 1 with HTTP 200.
- 24/24 full-download verification checks passed.
- Upload complete → actual first byte:
- p50: 0.442 s
- p90: 0.516 s
- p95: 0.552 s
- Worst sample: 0.976 s
- Normal shared-connection TTFB p50: 0.436 s
- Cold-connection TTFB p50: 0.878 s

The immediate first-byte p50 is only about 5 ms slower than ordinary shared-connection TTFB. There is no longer a separate five-second piece-readiness wait after upload completion.

documentation/en/experimental-features/PDPCURIOSPEC.md
```diff
@@ -180,11 +178,17 @@ The sections below describe how tasks and watchers connect to form the PDP lifec
...
+3. After the write succeeds, one database transaction marks the parked piece complete, creates `pdp_piecerefs`, and deletes the upload row. The legacy `notify` request field is accepted for compatibility but this path does not call the URL or schedule a notify task.
+
+**Streaming upload path:**
+1. The client creates an upload session and PUTs bytes without declaring a PieceCID. The handler claims a provisional `parked_pieces` identity for that session, then streams the request once directly into final piece storage while calculating CommP and the raw size.
+2. After the PieceCID is known, one transaction promotes the provisional row to the calculated identity. If a matching complete piece already exists, the session is pointed at that piece and the unreferenced provisional copy is left for normal parked-piece cleanup.
+3. The client finalizes with the calculated PieceCID. The handler validates it, creates `pdp_piecerefs`, and deletes the streaming session in one transaction. It does not use scratch space, create an intermediate `pdp_piece_uploads` row, or schedule a notify task.
```
snadrus, line 194: On disk, is this a file move?

pdp/README.md
```diff
@@ -197,7 +195,7 @@ The streaming upload API provides a way to upload large pieces in a streaming fa
 2. Stream the data via `PUT`.
 3. Finalize the upload with the pieceCid to link and validate.
 
-> **Note:** Each streaming upload chunk is limited to **1 GiB** (unpadded). The server computes the CommP on-the-fly.
+> **Note:** Each streaming upload is limited to **1,065,353,216 raw bytes** (1 GiB padded). The server writes it once directly to piece storage while computing CommP on-the-fly.
```
snadrus, line 198: Ensure this doesn't collide with the 64gb work

snadrus (approved):
So what about finalize? Should we allow that in the same call?

documentation/en/experimental-features/PDPCURIOSPEC.md
LexLuthr: No. File is already in storage. We just fix the piece park tables to use correct piece_cid.

pdp/README.md
LexLuthr: Yes. 64 GiB will be built on top of this. Even then we will need to put a limit on streaming upload. Maybe 4 GiB. Otherwise, we find out after 64 GiB that data is garbage.

PR #1450, opened 2026-08-21
ipni: confirm ad sync status via indexer, publish latency metric

### LINKED ISSUE #1405: Integrate IPNI /sync/status/ad/{adCid} to track confirmed advertisement indexing

beck-8:
https://github.com/filecoin-project/curio/pull/1426#issuecomment-5362366873
His follow-up update. Replace #1426 @BigLep @LexLuthr @jennijuju 

~~I chose to add the return adCid, so that the caller can choose his own check logic.
Users should verify the advertised information, but should not rely on the advertisedAt date. After confirming the advertised details, users should continue to check the ad's status and the cid.~~

~~## Summary of Changes~~

~~1. **`ipni-provider.go`**: Dropped `ipni_ad_fetches` writes; fixed a bug where `LastPublishTime` never got set after a restart; `PublishInterval` 5s → 1s.
2. **`pdp/handlers.go`**: Removed `retrieved`/`synced` fields, added `adCid` so callers can check sync status against the indexer themselves; `advertised`/`advertisedAt` logic unchanged.
3. **`pdp/README.md`**: Updated docs for the new response fields.
4. Added SQL migration to drop the `ipni_ad_fetches` table, with matching downgrade.~~

I chose to add and return `adCid`, so that the caller can choose their own check logic. Curio also checks the indexer's real sync-status endpoint on demand and exposes the result as `synced`/`syncedAt`, so callers who don't want to query the indexer themselves still get a confirmed answer.

Users should verify the `advertised` information, but should not rely on `advertisedAt` as a precise timestamp — it's a fixed estimate (`adCreatedAt` + publish interval), not an observed time, so it won't drift but also isn't exact. For an authoritative, indexer-confirmed time, use `synced`/`syncedAt` instead, or query the indexer directly with `adCid`.

`synced`/`syncedAt` are intentionally **not** persisted to the database. An indexer's confirmation isn't a one-time fact — it can resync an ad later — so we treat it as a live, in-memory-cached answer instead of a permanent DB record. This also means the cache clears on a Curio restart: a piece that showed `synced: true` before a restart can briefly show `false` again until it's re-checked. `true` is always a confirmed signal callers can act on immediately; `false` isn't proof the ad isn't indexed, just that Curio hasn't (re-)confirmed it yet.

## Summary of Changes

1. **`market/ipni/ipni-provider/ipni-provider.go`**: Dropped `ipni_ad_fetches` writes; fixed a bug where `LastPublishTime` never got set after a restart; `PublishInterval` 5s → 1s; inter-provider announce delay 1s → 200ms. Added on-demand ad sync-status checking against the configured indexer services' `/sync/status/ad/{adCid}` endpoint (ipni/storetheindex#2898): a small in-memory LRU cache (10k ads) answers `synced`/`syncedAt` immediately, and a cache miss triggers a non-blocking background check, deduped per `ad_cid` and capped at 20 concurrent checks against the indexer (skipped, not queued, past that cap). Added an `ipni_announce_to_indexed_latency_seconds` metric, computed from the ad's exact `ipni.created_at` timestamp.
2. **`market/ipni/ipni-provider/metrics.go`**: Added the `ipni_announce_to_indexed_latency_seconds` histogram, tagged only by `provider` — no `provider_type`/`service`, since pin.contact is being deprecated and cid.contact is the only indexer service left.
3. **`pdp/handlers.go`**: Removed the old `retrieved` field and the unreliable `MIN(ad_cid)` ad-selection logic, replaced with picking the most recently created ad per piece via `ipni.created_at` (with `order_number` only as a tiebreaker — `order_number` alone isn't reliably ordered under Yugabyte's cached sequences). Added `adCid` so callers can check sync status against the indexer themselves, and `synced`/`syncedAt` sourced live from the in-memory cache above; `status` gets a new `"synced"` terminal value. `advertised` is only true on a real confirmed announce; `advertisedAt` is a fixed, non-drifting estimate.
4. **`pdp/README.md`**: Updated docs for the new response fields, including that `synced` is checked on demand (not persisted) and can revert to `false` after a restart without the ad losing its indexed state.
5. Added a single SQL migration (with matching downgrade): drop the `ipni_ad_fetches` table, and add `ipni.created_at` (used for the ad-ordering fix above and for exact per-ad latency computation). No `indexed_at`/sync-status column — that's intentionally never persisted, per the note above.


beck-8:
If you think it is inappropriate to throw out the Ad information and let the user inquire by himself. And accepting this interface will be stuck, so you can put this part of the check in curio (I can't accept this kind of thing very well, and this is a remote call. You can call the interface of curio frequently to make curio frequently want to initiate cid requests. To handle this part, you need to add more Add cache status).

pdp/README.md
```diff
@@ -174,12 +177,14 @@ All endpoints are rooted at `/pdp`.
...
     - `indexed`: Whether the piece has been indexed and is ready for IPNI.
-    - `advertised`: Whether an IPNI advertisement has been published.
-    - `retrieved`: Whether the piece has been retrieved by a client.
-    - `retrievedAt`: Timestamp of last retrieval (omitted if never retrieved).
+    - `indexedAt`: Timestamp CAR indexing completed (omitted if not yet indexed).
+    - `adCreated`: Whether an IPNI advertisement has been created for this piece.
+    - `adCreatedAt`: Timestamp the advertisement was created (omitted if not yet created).
+    - `adCid`: This piece's advertisement CID, once created. Curio doesn't check whether an indexer finished processing it - callers can, e.g. `GET https://cid.contact/sync/status/ad/{adCid}`.
+    - `advertised`: Whether the provider has sent an HTTP announce covering this ad.
+    - `advertisedAt`: Approximate, not a fixed record: it's the last known successful announce time for the *provider*, not this ad specifically, so it can drift forward on later calls once the provider announces newer ads, and is lost on a Curio restart (the provider re-announces its current head once on startup to recover it).
```
BigLep: > it's the last known successful announce time for the *provider*, not this ad specifically,

Is that because we're no longer storing state in the database (memory only), or has this always been the case. I would have thought we could be more precise with ourselves on the first advertising chain we announced that includes this adCid. If we don't know that, then I think we should drop this field.

BigLep (commented):
Thanks @beck-8 . This is interesting. I like the thinking of getting Curio out of being the "middle-man" between a client (e.g., Dealbot) and an IPNI instance (e.g., cid.contact). dealbot can just query cid.contact directly to determine whether the advertisment has been ingested.

I'm trying to think of any downsides of taking Curio out of the loop. I don't understand failure cases enough to know, but is there a world where it would be beneficial for Curio to know that a client's cids haven't been indexed yet by cid.contact?

Other things I would imagine adding in the PR:
- In the return struct, lets include the ipniInstance that the adCid was advertised to. I don't want client's hardcoding cid.contact. 
- I think it would be good if the return structure also accounted for potentially multiple ipniInstances. 
- (I'm suggesting the above so that we don't need to worry about client updates. Updating SPs is easier I think that getting clients to update if we ever added a different/additional IPNI instance besides cid.contact.)

I defer to @LexLuthr on final decisions here.

pdp/README.md
beck-8: It turns out that there are three layers of judgment:

1. The first layer is determined by whether ipni comes to fetch, and we delete this logic here. This is fixed
2. The second layer is inferred by time. This will be offset.
3. The third floor is a pocket bottom, fixed time


So here we have to make a choice, either to retain the original logic or possibly delete everything? Only keep the state?

beck-8:
> Curio to know that a client's cids haven't been indexed yet by cid.contact?

I saw that there was another alarm in the previous design. It is to judge the gap between head and cid. Curio will send an alarm 1 hour behind.


> I don't want client's hardcoding cid.contact.

~~I've thought about this, but I don't have a good idea at present. Because there are two cid and filecoinpin now, two can be returned here, because both of them push head, but one of them does not have an ad status interface. Or the configuration we add to curio after this time must be a completely equal ipni node. That is to say, delete filecoinpin this time, otherwise, he will return two ipni.
What I prefer is that the two ipni become completely equal instances. Then we follow the migration process, release it first, and then delete one. But how should I use it when I get two IPNI results downstream?~~

I reconsidered it. This part is not elegant. And it is our consensus which IPNI to use now. It does not allow mistakes or problems. This website has been hard-coded in many downstream places. If you want to replace it, you need everyone to replace it


I will make changes when I get more input from you.

BigLep:
HI @beck-8 - thanks for engaging. I'll let @LexLuthr make the calls here as he knows more of the specifics.

In general, I want to make sure we have some flexibility in our system to not always assume cid.contact. If cid.contact has problems again we'll likely spin up our own IPNI instance (not the current filecoinpin.contact with bespoke logic but a storetheindex instance that has the same endpoints as cid.contact). It would be great if that could "just work" by having SPs update config to use it and not require server API code and client code for that to work... But if that is too much extra we can bite the bullet then. My general position is I want to be able to adapt quickly on the content routing front, and API flexibility is one area where I think we can.

LexLuthr (changes requested):
Let's drop the publish time between providers to 200ms. We should show ad created, ad published and ad synced. Remove all time stamps. Make sure to inform Russel about the changes. Let him know we will redo the metrics soon once IPNI side changes are in production.

beck-8:
> 200ms

Just like the information obtained in slack communication, although it does not set the flow limit now. I'm not sure what will happen in the future. Do we really have to set it to exceed its upper limit? 60 RPM

> synced

I didn't add this status here. I left it to the downstream to check by myself. So I don't understand what we want to change here?

LexLuthr:
> > 200ms
> 
> Just like the information obtained in slack communication, although it does not set the flow limit now. I'm not sure what will happen in the future. Do we really have to set it to exceed its upper limit? 60 RPM
> 
> > synced
> 
> I didn't add this status here. I left it to the downstream to check by myself. So I don't understand what we want to change here?

Let's put is 200ms for now. If things start to falter we will backoff. 
We have created_at (you can add the column as required)
We can easily do announced (created+1 second)
We now have sync time from IPNI. (synced - (created +1)) == sync duration. 

I hope this makes things clear. Let me know if there is still confusion. I will let you figure out the JSON object we send.

beck-8:
@LexLuthr  I think I'm ready for another review. Although some places use order_number for comparison this time, I think it is completely acceptable, just to get the items that have not been checked.

harmony/harmonydb/sql/20260827-ipni-indexed-at.sql
LexLuthr, line 1: Indexing is not a one time event. We should not be recording it. Also, please don't use more than 1 SQL in the same PR. We must keep SQL file number low. Each file ads more complexity to the upgrade or init.

market/ipni/ipni-provider/metrics.go
```diff
@@ -39,13 +39,16 @@ func (w *metricResponseWriter) Status() int {
...
+	ipniProviderHTTPRequestBuckets      = []float64{1, 5, 10, 25, 50, 100, 250, 500, 1000, 2500, 5000, 10000, 30000}
+	ipniAnnounceRoundTripBuckets        = []float64{10, 25, 50, 100, 250, 500, 1000, 2500, 5000, 10000, 30000, 60000}
+	ipniAnnounceToIndexedLatencyBuckets = []float64{1, 3, 5, 10, 15, 30, 45, 60, 120, 300, 420, 600, 900, 1800}
 
-	providerTag, _ = tag.NewKey("provider")
-	contentTag, _  = tag.NewKey("content")
-	statusTag, _   = tag.NewKey("status")
-	resultTag, _   = tag.NewKey("result")
+	providerTag, _     = tag.NewKey("provider")
+	providerTypeTag, _ = tag.NewKey("provider_type")
```
LexLuthr: Why do we have types now?

market/ipni/ipni-provider/metrics.go
```diff
@@ -39,13 +39,16 @@ func (w *metricResponseWriter) Status() int {
...
+	serviceTag, _      = tag.NewKey("service")
```
LexLuthr: What service?

pdp/handlers.go
```diff
@@ -270,18 +270,17 @@ func (p *PDPService) handleGetPieceStatus(w http.ResponseWriter, r *http.Request
...
-		Provider                 sql.NullString `db:"provider"`
+		PieceCID               string         `db:"piece_cid"`
+		PieceRawSize           uint64         `db:"piece_raw_size"`
+		CreatedAt              time.Time      `db:"created_at"`
+		Indexed                bool           `db:"indexed"`
+		IndexedAt              sql.NullTime   `db:"indexed_at"`
+		AdvertisementCreated   bool           `db:"advertisement_created"`
+		AdvertisementCreatedAt sql.NullTime   `db:"advertisement_created_at"`
+		AdCID                  sql.NullString `db:"ad_cid"`
+		SyncedAt               sql.NullTime   `db:"synced_at"`
```
LexLuthr:
```suggestion
		AdvertisementSyncedAt               sql.NullTime   `db:"synced_at"`
```

pdp/handlers.go
```diff
@@ -270,18 +270,17 @@ func (p *PDPService) handleGetPieceStatus(w http.ResponseWriter, r *http.Request
...
-		Status                   string         `db:"status"`
-		Provider                 sql.NullString `db:"provider"`
+		PieceCID               string         `db:"piece_cid"`
+		PieceRawSize           uint64         `db:"piece_raw_size"`
+		CreatedAt              time.Time      `db:"created_at"`
+		Indexed                bool           `db:"indexed"`
+		IndexedAt              sql.NullTime   `db:"indexed_at"`
+		AdvertisementCreated   bool           `db:"advertisement_created"`
+		AdvertisementCreatedAt sql.NullTime   `db:"advertisement_created_at"`
+		AdCID                  sql.NullString `db:"ad_cid"`
```
LexLuthr, line 281: Add AdvertisementSynced as well

LexLuthr (changes requested):
We should show sync status in status json.

market/ipni/ipni-provider/metrics.go
beck-8: I just added it when I saw this field. For example, the node of james runs together with RoRep and PDP.

market/ipni/ipni-provider/metrics.go
beck-8: service: `cid.contact `

Will it be clearer if we change it to another name?

market/ipni/ipni-provider/metrics.go
LexLuthr: IPNI is a common system. Let's not add unnecessary complexity if we don't require it.

market/ipni/ipni-provider/metrics.go
LexLuthr: As we are removing pin.contact, we don't need this. cid.contact is only service left.

BigLep:
@beck-8 : I saw you have followup commits. I assume you'll re-request review when it's ready for review again.

pdp/README.md
```diff
@@ -174,12 +179,17 @@ All endpoints are rooted at `/pdp`.
         - `"pending"` – Not yet indexed.
         - `"indexing"` – CAR indexing task is in progress.
         - `"creating_ad"` – IPNI advertisement is being created.
-        - `"announced"` – Advertisement published to IPNI network.
-        - `"retrieved"` – Piece has been retrieved by a client.
+        - `"announced"` – The advertisement row exists. Not a guarantee it's been broadcast (see `advertised`) or indexed (see `synced`).
```
BigLep:
```suggestion
        - `"announced"` – The advertisement row exists. Not a guarantee it's been broadcast (see `advertised`) or indexed externally (see `synced`).
```

pdp/README.md
```diff
@@ -174,12 +179,17 @@ All endpoints are rooted at `/pdp`.
...
+        - `"synced"` – A configured indexer service (e.g. cid.contact) has confirmed it fully processed the advertisement.
     - `indexed`: Whether the piece has been indexed and is ready for IPNI.
```
BigLep:
```suggestion
    - `indexed`: Whether the piece has been indexed locally and is ready for IPNI.
```

pdp/README.md
```diff
@@ -174,12 +179,17 @@ All endpoints are rooted at `/pdp`.
...
-    - `advertised`: Whether an IPNI advertisement has been published.
-    - `retrieved`: Whether the piece has been retrieved by a client.
-    - `retrievedAt`: Timestamp of last retrieval (omitted if never retrieved).
+    - `indexedAt`: Timestamp CAR indexing completed (omitted if not yet indexed).
```
BigLep:
```suggestion
    - `indexedAt`: Timestamp when local CAR indexing completed (omitted if not yet indexed).
```

pdp/README.md
```diff
@@ -174,12 +179,17 @@ All endpoints are rooted at `/pdp`.
         - `"pending"` – Not yet indexed.
         - `"indexing"` – CAR indexing task is in progress.
```
BigLep:
```suggestion
- `"pending"` – Not yet locally indexed.
- `"indexing"` – Local CAR indexing task is in progress.
```

pdp/README.md
```diff
@@ -174,12 +179,17 @@ All endpoints are rooted at `/pdp`.
...
         - `"creating_ad"` – IPNI advertisement is being created.
-        - `"announced"` – Advertisement published to IPNI network.
-        - `"retrieved"` – Piece has been retrieved by a client.
+        - `"announced"` – The advertisement row exists. Not a guarantee it's been broadcast (see `advertised`) or indexed (see `synced`).
+        - `"synced"` – A configured indexer service (e.g. cid.contact) has confirmed it fully processed the advertisement.
```
BigLep:
```suggestion
        - `"synced"` – A configured IPNI instance (e.g., cid.contact) has confirmed it fully processed the advertisement.
```

pdp/README.md
```diff
@@ -174,12 +179,17 @@ All endpoints are rooted at `/pdp`.
...
     - `indexed`: Whether the piece has been indexed and is ready for IPNI.
-    - `advertised`: Whether an IPNI advertisement has been published.
-    - `retrieved`: Whether the piece has been retrieved by a client.
-    - `retrievedAt`: Timestamp of last retrieval (omitted if never retrieved).
+    - `indexedAt`: Timestamp CAR indexing completed (omitted if not yet indexed).
+    - `adCreated`: Whether an IPNI advertisement has been created for this piece.
+    - `adCreatedAt`: Timestamp the advertisement was created (omitted if not yet created).
+    - `adCid`: This piece's advertisement CID, once created. Callers who want to double-check independently can query an indexer directly, e.g. `GET https://cid.contact/sync/status/ad/{adCid}`.
```
BigLep:
```suggestion
    - `adCid`: This piece's advertisement CID, once created. Callers who want to double-check independently can query an IPNI instance directly (e.g., `GET https://cid.contact/sync/status/ad/{adCid}`).
```

pdp/README.md
```diff
@@ -174,12 +179,17 @@ All endpoints are rooted at `/pdp`.
...
+    - `advertised`: Whether the provider has sent an HTTP announce covering this ad.
+    - `advertisedAt`: `adCreatedAt` plus the announce publish interval - a fixed estimate, not the provider's raw last-announce time, so it doesn't drift on later calls as the provider announces newer ads.
+    - `synced`: Whether an indexer has confirmed it fully processed this ad, per an in-memory (not persisted) cache checked on demand. A cache miss triggers a background check against the configured indexer services and returns `false` for this call - a later call for the same piece will see the result once that check completes. `true` is a confirmed signal - callers can act on it right away. `false` is inconclusive rather than proof the ad isn't indexed: it can also show up on the first call after a Curio restart (cache cleared) or if a later call catches an indexer resync, without the ad actually losing its indexed state - poll again if that matters for your workflow.
```
BigLep:
```suggestion
    - `synced`: Whether an IPNI instance has confirmed it fully processed this ad, per an in-memory (not persisted) cache checked on demand. A cache miss triggers a background check against the configured IPNI instance and returns `false` for this call - a later call for the same piece will see the result once that check completes. `true` is a confirmed signal - callers can act on it right away. `false` is inconclusive rather than proof the ad isn't indexed: it can also show up on the first call after a Curio restart (cache cleared) or if a later call catches an IPNI instance resync, without the ad actually losing its indexed state - poll again if that matters for your workflow.
```

pdp/README.md
```diff
@@ -174,12 +179,17 @@ All endpoints are rooted at `/pdp`.
...
+    - `syncedAt`: The indexer-reported processing time, when `synced` is true.
```
BigLep:
```suggestion
    - `syncedAt`: When `synced` becomes true, this stores the IPNI instance's reported processing time. 
```

BigLep (commented):
Nits on the documentation, just being extra clear on local vs. external. Used the term "IPNI instance" to refer to cid.contact.

pdp/handlers.go
```diff
@@ -381,72 +375,31 @@ func (p *PDPService) handleGetPieceStatus(w http.ResponseWriter, r *http.Request
...
+	// formula below, so it doesn't drift as the provider announces later ads.
+	if result.AdvertisementCreatedAt.Valid {
+		var publishedAt *time.Time
+		if p.ipp != nil && result.Provider.Valid {
+			publishedAt = p.ipp.LastPublishTime(result.Provider.String)
+		}
+		if publishedAt != nil && publishedAt.After(result.AdvertisementCreatedAt.Time) {
+			response.Advertised = true
+			response.AdvertisedAt = new(result.AdvertisementCreatedAt.Time.Add(ipni_provider.PublishInterval))
 		}
```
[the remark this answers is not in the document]
beck-8: go 1.26

beck-8:
> @beck-8 : I saw you have followup commits. I assume you'll re-request review when it's ready for review again.

Yes, we have gone through several rounds of private communication and applied the suggestions Lex gave me. Currently waiting for confirmation from @LexLuthr.

harmony/harmonydb/downgrade/20260821-ipni-piece-status.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE ipni DROP COLUMN IF EXISTS created_at;
```
LexLuthr: Remove this from downgrade.

harmony/harmonydb/sql/20260821-ipni-piece-status.sql
```diff
@@ -0,0 +1,10 @@
+-- ipni_ad_fetches recorded raw ad-body fetches, which only proves an indexer
+-- pulled the ad, not that it finished processing it. The piece-status
+-- endpoint now returns the ad's CID directly so callers can check sync status
+-- against the indexer themselves, and "advertised" is tracked in memory by
+-- the ipni-provider process (see market/ipni/ipni-provider), so nothing here
+-- needs replacing.
+DROP TABLE IF EXISTS ipni_ad_fetches;
+
+-- Used for ordering (pdp/handlers.go) and latency (market/ipni/ipni-provider).
+ALTER TABLE ipni ADD COLUMN IF NOT EXISTS created_at TIMESTAMPTZ NOT NULL DEFAULT TIMEZONE('UTC', NOW());
```
LexLuthr:
```suggestion
ALTER TABLE ipni ADD COLUMN IF NOT EXISTS created_at TIMESTAMPTZ NOT NULL DEFAULT NOW();
```

pdp/handlers.go
```diff
@@ -381,72 +375,31 @@ func (p *PDPService) handleGetPieceStatus(w http.ResponseWriter, r *http.Request
...
-				if publishedAt.After(time.Now().Add(ipni_provider.PublishInterval)) {
-					response.AdvertisedAt = new(result.AdvertisementCreatedAt.Time.Add(ipni_provider.PublishInterval))
-				}
-			} else {
-				response.Advertised = false
-				response.AdvertisedAt = nil
+	if result.AdCID.Valid && p.ipp != nil {
+		if adCid, err := cid.Parse(result.AdCID.String); err == nil {
+			if syncedAt := p.ipp.SyncedAt(adCid, result.Provider.String); syncedAt != nil {
+				response.Status = "synced"
```
LexLuthr: Let's not change Status here. Correct Status end i.e. pipeline end is announced/ad created

market/ipni/ipni-provider/ipni-provider.go
```diff
@@ -771,3 +776,131 @@ func (p *Provider) LastPublishTime(providerPeerID string) *time.Time {
...
+	default:
+		p.syncChecksInFlight.Delete(adCid)
+		return
+	}
+
+	go func() {
+		defer p.syncChecksInFlight.Delete(adCid)
+		defer func() { <-p.syncCheckSem }()
+
+		ctx := context.Background()
```
LexLuthr, line 815: ~Lets use a timeout 5 second max.~

market/ipni/ipni-provider/ipni-provider.go
```diff
@@ -771,3 +776,131 @@ func (p *Provider) LastPublishTime(providerPeerID string) *time.Time {
...
+	}()
+}
+
+// queryAdSyncStatus checks each configured service in turn for a definitive answer.
+func (p *Provider) queryAdSyncStatus(ctx context.Context, adCid cid.Cid) (service string, status adSyncStatus, ok bool) {
+	for _, svc := range p.serviceURLs {
+		u := *svc
+		u.Path = path.Join(u.Path, "sync", "status", "ad", adCid.String())
+
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, u.String(), nil)
```
LexLuthr, line 848: Let's use a timeout here with max 5 second

LexLuthr (changes requested):
Few more minor changes required. ~Apart from this please remove pin.contact from the default config.~ Let's remove later in a separate PR.

market/ipni/ipni-provider/ipni-provider.go
```diff
@@ -92,6 +104,17 @@ type Provider struct {
 	announceURLs   []*url.URL
 	httpServerBase *url.URL
 	startLock      sync.Once
+
+	// serviceURLs are indexer services checked for ad sync status, e.g.
+	// https://cid.contact. Not all of them support that endpoint.
+	serviceURLs []*url.URL
+	syncClient  *http.Client
+	// syncCache holds confirmed ad sync results, populated lazily on request
+	syncCache *lru.Cache[cid.Cid, syncResult]
```
LexLuthr: Why not?
```suggestion
	syncCache *lru.Cache[cid.Cid, *time.Time]
```

market/ipni/ipni-provider/ipni-provider.go
beck-8: `syncClient:     &http.Client{Timeout: 5 * time.Second},`

It was set here at the time of initialization.

LexLuthr (approved):
Looks good. Can we test this in calibnet before we merge? Maybe run it for few hours and then we merge if output is consistent.

beck-8:
The test situation looks good. Additional processing of inaccurate merics (the`latency":4969.648333334`)

```
{"level":"info","ts":"2026-08-28T16:53:16.474+0800","logger":"ipni-provider","caller":"ipni-provider/ipni-provider.go:832","msg":"IPNI advertisement confirmed indexed","provider":"12D3KooWPRuubsz8p49yAFBZGwsKVEF9MqxN1CQ1QRMeaDE4VCZu","ad_cid":"baguqeeraal3bcyv3p5vjbdukwa7jyyudjwhlmh72swyskn3uec5wvo5n6pya","service":"https://cid.contact","indexed_at":"2026-08-27T21:38:38.552Z","check_took":0.981434971,"latency":null}
{"level":"info","ts":"2026-08-28T17:04:50.188+0800","logger":"ipni-provider","caller":"ipni-provider/ipni-provider.go:832","msg":"IPNI advertisement confirmed indexed","provider":"12D3KooWPRuubsz8p49yAFBZGwsKVEF9MqxN1CQ1QRMeaDE4VCZu","ad_cid":"baguqeeraco6wpef7r6nax5uriefiqaismdgmu6mbbx4aynwrotwvvcolfrlq","service":"https://cid.contact","indexed_at":"2026-08-26T16:22:57.826Z","check_took":0.863642209,"latency":null}
{"level":"info","ts":"2026-08-28T17:09:20.761+0800","logger":"ipni-provider","caller":"ipni-provider/ipni-provider.go:832","msg":"IPNI advertisement confirmed indexed","provider":"12D3KooWPRuubsz8p49yAFBZGwsKVEF9MqxN1CQ1QRMeaDE4VCZu","ad_cid":"baguqeerapygdx5cmwi2xlvvlkpehga3ieuqsimksxohq63ggmsngmaxr7dgq","service":"https://cid.contact","indexed_at":"2026-08-28T09:09:20.601Z","check_took":0.312414905,"latency":1.542224}
{"level":"info","ts":"2026-08-28T17:10:24.091+0800","logger":"ipni-provider","caller":"ipni-provider/ipni-provider.go:832","msg":"IPNI advertisement confirmed indexed","provider":"12D3KooWPRuubsz8p49yAFBZGwsKVEF9MqxN1CQ1QRMeaDE4VCZu","ad_cid":"baguqeeragfv2dwz56wsjlltsuqeidj63kcme2brp4rh4p6nsrgevnkpr4cbq","service":"https://cid.contact","indexed_at":"2026-08-28T09:06:15.202Z","check_took":0.424712273,"latency":0.180321}
{"level":"info","ts":"2026-08-28T17:15:34.789+0800","logger":"ipni-provider","caller":"ipni-provider/ipni-provider.go:832","msg":"IPNI advertisement confirmed indexed","provider":"12D3KooWPRuubsz8p49yAFBZGwsKVEF9MqxN1CQ1QRMeaDE4VCZu","ad_cid":"baguqeerargw5fin46phq7h3y4k62pdvqvbk2455bcon4p7wufooogdevg2ya","service":"https://cid.contact","indexed_at":"2026-08-28T09:15:30.956Z","check_took":0.282425723,"latency":3.533442}

{"level":"info","ts":"2026-08-28T17:22:45.539+0800","logger":"ipni-provider","caller":"ipni-provider/ipni-provider.go:832","msg":"IPNI advertisement confirmed indexed","provider":"12D3KooWPRuubsz8p49yAFBZGwsKVEF9MqxN1CQ1QRMeaDE4VCZu","ad_cid":"baguqeera6mjadskymsd67rrl5txphsyfwlejyalveka2sqgczreq5o7u6nja","service":"https://cid.contact","indexed_at":"2026-08-28T09:22:45.507Z","check_took":0.74257171,"latency":4969.648333334} <-- fixed

{"level":"info","ts":"2026-08-28T17:24:41.301+0800","logger":"ipni-provider","caller":"ipni-provider/ipni-provider.go:832","msg":"IPNI advertisement confirmed indexed","provider":"12D3KooWPRuubsz8p49yAFBZGwsKVEF9MqxN1CQ1QRMeaDE4VCZu","ad_cid":"baguqeera43sebh7fc65hy4mjvyrev5rlsrdoaev3vyphtrcan5wlpyoakmka","service":"https://cid.contact","indexed_at":"2026-08-28T09:24:38.653Z","check_took":0.407780708,"latency":0.5904}
{"level":"info","ts":"2026-08-28T17:27:14.131+0800","logger":"ipni-provider","caller":"ipni-provider/ipni-provider.go:832","msg":"IPNI advertisement confirmed indexed","provider":"12D3KooWPRuubsz8p49yAFBZGwsKVEF9MqxN1CQ1QRMeaDE4VCZu","ad_cid":"baguqeera7nod2wl2yrgcrxfsft77gdqzgzwuyjli32szgp3t7lqldhyeqpxa","service":"https://cid.contact","indexed_at":"2026-08-28T09:27:12.772Z","check_took":0.28451591,"latency":1.562837}
{"level":"info","ts":"2026-08-28T17:28:19.418+0800","logger":"ipni-provider","caller":"ipni-provider/ipni-provider.go:832","msg":"IPNI advertisement confirmed indexed","provider":"12D3KooWPRuubsz8p49yAFBZGwsKVEF9MqxN1CQ1QRMeaDE4VCZu","ad_cid":"baguqeeracqncflgcxf3v4nipfiwvdsdcaylqugs343vdkc7n3oxh26ljoiaq","service":"https://cid.contact","indexed_at":"2026-08-28T09:28:14.779Z","check_took":0.393532267,"latency":1.513534}
{"level":"info","ts":"2026-08-28T17:39:01.140+0800","logger":"ipni-provider","caller":"ipni-provider/ipni-provider.go:832","msg":"IPNI advertisement confirmed indexed","provider":"12D3KooWPRuubsz8p49yAFBZGwsKVEF9MqxN1CQ1QRMeaDE4VCZu","ad_cid":"baguqeerarexjjbq46tafq2mtj3rtvbpd3ongwaga5vnv2qmgxuj4f22x33vq","service":"https://cid.contact","indexed_at":"2026-08-28T09:38:58.878Z","check_took":0.306918079,"latency":1.475841} 
```

---
request preview

```
#https://pdp.660688.xyz:8443/pdp/piece/bafkzcibet2nryd7fuwxwu2u3ljz7ov4qwlobdvlb77zdblibekanpqjkpycaeftseq/status

{
  "pieceCid": "bafkzcibet2nryd7fuwxwu2u3ljz7ov4qwlobdvlb77zdblibekanpqjkpycaeftseq",
  "status": "announced",
  "indexed": true,
  "indexedAt": "2026-08-28T08:21:44.78419Z",
  "adCreated": true,
  "adCreatedAt": "2026-08-28T08:22:11.606784Z",
  "adCid": "baguqeeraavfjndbvjanon6czonh72tl6h7v52rcho3eyvyvifssrefdsngda",
  "advertised": true,
  "advertisedAt": "2026-08-28T08:22:12.606784Z",
  "synced": true,
  "syncedAt": "2026-08-28T08:22:15.260522Z"
}

```

PR #1454, opened 2026-08-24
Feat: UI of PDP datasets in grace period

### LINKED ISSUE #1435: Easily Tracking Faulty Datasets
Copying over from #1352:

1. Detect when a dataset is affected by a terminal or sustained retention/proving condition that can lead to termination.
2. Move the dataset into an explicit review required state.
3. Record the reason, first detection time, current state, and review deadline.
4. Activate a persistent operator alert through the alert manager.
5. ~Resolve the alert and cancel pending termination if the underlying condition recovers.~
6. Preserve enough failure and service context for the SP to investigate the condition, including through external systems such as the MSP.
7. Maintain an audit trail of state transitions and operator actions.

After recent work our pending termination marking should only be happening if errors are truly unrecoverable so I've crossed out point 5.

PR #1332 probably handles point 4 as well, though I'll let @LexLuthr comment on this since he opened the originating issue after filing that PR.

Overall I am imagining a UI page which immediately presents operator with the biggest dataset problems and information on the nature of the problems.  Since I haven't been running the latest UI its possible that this is mostly covered by the latest UI improvements cc @snadrus.

snadrus:
The UI should show datasets at-risk so SPs can contact them, and expect their removal (at the time listed). 

Resolves #1435 minus the "delete now" button which I recommend de-scoping until it becomes an SP-requested feature.

Note: 
- Glif was used to populate this list faster (with chain fallback). 
- It ignores datasets < 100kb in total size for this UI view as they probably won't add up to much, and to save chain calls (this UI is excessive already). 

Load time of this page was about 30 seconds, populated progressively and largest-first.
image


LexLuthr (approved):
Remove glif

PR #1477, opened 2026-08-28
fix: keep IPNI announce dedupe when one indexer is down

SgtPooki:
## What changed

`publishhttp` now announces to each configured indexer through its own sender and returns how many accepted it, instead of handing every URL to a single `httpsender`. `announceProviderHead`, extracted from `publishHead`, records a head as published when at least one indexer accepted it.

Before this, one failing indexer made `publishHead` treat the whole publish as failed, so `p.latest` never recorded the head and the duplicate-announce guard never fired again. An unchanged head was then re-announced to every healthy indexer on every publish tick. With `PublishInterval` at 1s (v1.28.6) that is one announce per second per provider to cid.contact instead of one per new advertisement, plus a matching ERROR log line.

This is worth landing now because filecoinpin.contact is being shut down on 2026-09-01 (filecoin-project/filecoin-pin#664) while the default `DirectAnnounceURLs` in `deps/config/types.go` still lists it. Removing it from the defaults is planned for 2026-09-02, so PDP providers will be announcing to a dead indexer until that ships and SPs pick it up.

## How to verify

Commit 1 adds the tests and fails CI on purpose. Commit 2 makes them pass.

```
go test ./market/ipni/ipni-provider/ -race -count=1
```

At commit 1 CI recorded `test-all` failing: run [33197251504](https://github.com/filecoin-project/curio/actions/runs/33197251504).

## Notes / risks

- A dead indexer never blocked a healthy one: go-libipni already sends to multiple URLs concurrently. Only the dedupe bookkeeping was wrong.
- `recordAnnounceAttempt` gains a `partial` result alongside `success` and `error`. An operator alert matching `result="error"` stops firing when some indexers accepted the announce. The metric description now spells the values out and `metrics-reference.md` is regenerated.
- `sendAnnounce` treats anything that is not 200 or 204 identically, so 410, 5xx and 404 are one path. Covered as subtests.
- An indexer that accepts the connection and never answers is covered by bounding the announce context, so the test runs in 0.2s rather than waiting out the 1 minute client timeout.
- Dropped a duplicate `Infow` in `publishhttp` that logged the unfiltered URL list.
- The web UI has the same trigger and is fixed separately in #1478: `web/api/webrpc/ipni.go` called `http.Get` with no timeout and failed the entire IPNI page when one service URL was bad. That one reaches every Curio node on default config, not just PDP providers.

### DIFF
--- a/market/ipni/ipni-provider/ipni-provider.go
+++ b/market/ipni/ipni-provider/ipni-provider.go
@@ -674,40 +674,52 @@ func (p *Provider) publishHead(ctx context.Context) {
 			continue
 		}
 
-		if _, ok := p.latest[provider]; ok && p.latest[provider] == c {
-			log.Debugw("Skipping duplicate announce for provider", "provider", provider, "cid", c.String())
-			continue
-		}
-
-		log.Infow("Publishing head for provider", "provider", provider, "cid", c.String())
-		err = p.publishhttp(ctx, c, provider)
-		if err != nil {
-			recordAnnounceAttempt(provider, "error")
-			log.Errorw("failed to publish head for provide", "provider", provider, "error", err)
-		} else {
-			recordAnnounceAttempt(provider, "success")
-			p.latest[provider] = c
-			p.mu.Lock()
-			info, ok := p.providerInfos[provider]
-			if !ok {
-				log.Warnw("cannot update provider announce times", "provider", provider, "error", "provider not found")
-			} else {
-				info.lastPublishTime = new(time.Now())
-			}
-			p.mu.Unlock()
-		}
+		p.announceProviderHead(ctx, provider, c)
 
 		i++
 	}
 }
 
-// publishhttp sends an HTTP announce message for the given advertisement CID and peer ID.
-// It creates an HTTP announce sender using the provided announce URLs and the private key of the peer.
-// It obtains the HTTP addresses for the peer and sends the announce message to those addresses.
-func (p *Provider) publishhttp(ctx context.Context, adCid cid.Cid, peer string) error {
-	// Create the http announce sender.
-	log.Infow("Creating http announce sender", "urls", p.announceURLs)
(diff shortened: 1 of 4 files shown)


ZenGround0:
@LexLuthr could you take a look at this one?

market/ipni/ipni-provider/ipni-provider.go
```diff
@@ -727,22 +738,48 @@ func (p *Provider) publishhttp(ctx context.Context, adCid cid.Cid, peer string)
...
 
 	addrs := []multiaddr.Multiaddr{info.httpServerAddresses}
 
-	log.Infow("Announcing advertisements over HTTP", "urls", p.announceURLs)
-	return announce.Send(ctx, adCid, addrs, httpSender)
+	log.Infow("Announcing advertisements over HTTP", "urls", a)
+
+	// One sender per URL: a single sender joins every outcome into one error, leaving a
+	// partial failure indistinguishable from a total one.
+	var wg sync.WaitGroup
```
LexLuthr, line 758: Can we use errGroup instead of waitGroup? I think it will make for a much cleaner code given our intent. Also, another PR has already removed pin URL. You should rebase your PR on main

LexLuthr (changes requested):
PR is good. Please rebase on main and then tackle the review.

LexLuthr:
I was looking at this PR to abstract away the change into a new PR and combine with https://github.com/filecoin-project/curio/pull/1478 and I found a serious issue. This PR seems to just move the parallel announce ability of go-libipni into Curio which doesn't seem like a good choice. Worse part is that with these changes, Curio will never announce to second indexer if it fails once till a new head comes in. That seems like a bad idea no matter how you look at it. Head change is not governed by what's announced. If a new head comes in, Curio will simply switch to announcing that. The correct solution was always removing the dead indexer from config. 

I am closing this PR based on the current state on main and findings.

PR #1478, opened 2026-08-28
fix: keep IPNI summary alive when an indexer is unreachable

SgtPooki:
## What changed

`IPNISummary` probed each configured indexer service with `http.Get` and returned on any transport error, so a single unreachable service took down the entire IPNI page for every provider. `http.Get` also ignores the request context, and `DefaultClient` sets no timeout, so a service that accepts the connection and never answers hung the call with nothing to cancel it.

The probe moves into `fetchIPNIProvider`, which builds the request with the caller's context under a 10s bound and returns the failure instead of aborting. `IPNISummary` records it against that service and carries on with the rest.

This matters on the same date as #1477: filecoinpin.contact is being shut down on 2026-09-01 (filecoin-project/filecoin-pin#664) while the default `ServiceURL` in `deps/config/types.go` still lists it. Unlike `DirectAnnounceURLs`, that default is not filtered by `SPID`, so this reaches every Curio node on default config, PoRep-only included, not just PDP providers.

Which shutdown shape you get decides the damage. A deleted DNS record or a refused connection breaks the page outright; a host that connects and never answers hung it indefinitely; 410 or 5xx was already handled gracefully.

## How to verify

Commit 1 adds the itest and fails. Commit 2 makes it pass. The test needs a database, so it runs as its own CI matrix entry, `test-itest-ipni-summary`.

```
go test ./itests/ -run TestIPNISummaryUnreachableIndexerService -count=1
```

At commit 1 it fails at `web/api/webrpc/ipni.go:297` with `connection refused`, against unmodified source.

## Notes / risks

- Also fixes response bodies being closed by a `defer` inside a nested loop, which held every body open until the whole summary returned.
- A service returning 410 or a 5xx behaved correctly before and still does; only transport failures changed.
- The 10s bound is a constant, so the timeout path itself has no test. Testing it means either a 10s test or making the bound injectable.
- Config still beats code for the deadline: dropping filecoinpin.contact from `ServiceURL` needs no release and fixes this for any SP who does it.

Related: #1477 fixes the announce-side bug with the same trigger.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -219,6 +219,8 @@ jobs:
             target: "./itests/dyncfg_test.go"
           - name: test-itest-harmonyDB
             target: "./itests/harmonydb_test.go"
+          - name: test-itest-ipni-summary
+            target: "./itests/ipni_summary_test.go"
           - name: test-itest-park-piece
             target: "./itests/park_piece_test.go"
           - name: test-itest-pdp-prove
--- a/itests/ipni_summary_test.go
+++ b/itests/ipni_summary_test.go
@@ -1,0 +1,107 @@
+package itests
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/filecoin-project/curio/deps"
+	"github.com/filecoin-project/curio/harmony/harmonydb"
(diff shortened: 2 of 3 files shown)


LexLuthr (changes requested):
Thanks for the PR. Looks good. I would request change to the test. `itest` are integration test which require running full Curio or a big subsystem. This test can easily be converted into a regular unit test. Current test is actually testing fetchIPNIProvider and not IPNI summary really. Unit test is more suitable for it. We don't even need DB for it. 2 simple web server with hardcoded reply to any request.

Issue #1425, opened 2026-08-12
pdp: completed DropPiece tasks re-scheduled indefinitely under scheduler congestion; one slow CanAccept starves all task types


TippyFlitsUK:
**Setup:** curio v1.28.3 (mainnet, PDP-only, single machine), `parked_pieces` at 3.98M rows.

A batch of 11 DropPiece tasks posted at 10:45, was claimed at 10:56, and all 11 completed successfully by 11:01 (confirmed in `harmony_task_history`; rows correctly deleted from `harmony_task`). The scheduler then kept re-scheduling those same 11 completed IDs every pass for the next 90+ minutes:

```
{"ts":"2026-08-12T12:28:56.660Z","msg":"Slow SQL run","query":"SELECT COALESCE(array_agg(cleanup_task_id)...","milliseconds":154993}
{"ts":"2026-08-12T12:28:56.661Z","msg":"did not accept task","task_ids":[56681928,...,56681938],"reason":"CanAccept() refused","name":"DropPiece"}
```

Two defects compound:

1. **CanAccept runs inline in the single scheduler goroutine** (`pollerTryAllWork` -> `considerWork`). DropPiece's accept query (`task_cleanup_piece.go:186`) was running at ~150s on this node after planner statistics went stale (#1315). While it runs, no other task type gets scheduled and no scheduler events are processed - Prove, ProvPeriod, PullPiece, StorePiece and SendTransaction backlogs built up, the oldest sitting unclaimed for 2.5+ hours.
2. **Completed tasks are not evicted while the scheduler is congested.** The DB-poll path replaces the available-set map with a fresh snapshot (`scheduler.go`, `schedulerSourceDBPoll`), which should garbage-collect them, but at 150s per pass the completed IDs keep returning - the loop self-sustains. A restart cleared it for exactly 15 minutes before the next DropPiece batch (8 tasks, claimed in three sub-batches, each completing in under a second) re-established it with fresh IDs. Fixing the query cost instead (ANALYZE restored the 2ms plan) ended the loop on its own within a few minutes - the tail was pooled connections holding pre-ANALYZE cached plans - with no restart, so the retention is congestion-driven rather than permanent state.

Reproduced twice; timelines and `harmony_task_history` extracts available for both.

**Suggestion:** a deadline on CanAccept calls so one type's DB query cannot block the scheduling pass indefinitely, and reconciling the available set against completion so finished tasks cannot be re-offered.

PR #1479, opened 2026-08-30
pdp: accept DropPiece tasks for pieces that have no storage location

TippyFlitsUK:
## Priority

Not urgent relative to the contract upgrade work. Nothing is starved; the stuck tasks accumulate quietly and cost only log noise and permanent queue rows. It closes a stranded-work leak, so worth landing when there is room.

## The problem

`CleanupPieceTask.CanAccept` only accepts a task if the piece has a `sector_location` row (`miner_id = 0, sector_filetype = 32`) on the node's own storage. A parked piece that never got a location row can therefore never be accepted by any node, and its DropPiece task sits in `harmony_task` forever. `MaxFailures` never engages because the task never runs.

Such pieces exist in normal operation. `releaseStreamingUploadClaim` drops an abandoned streaming upload's ref and leaves the zero-ref provisional piece "for normal piece cleanup" — but that piece never completed, so no file was written and no location was ever declared. The cleanup poller schedules a DropPiece for it, and the task is born unschedulable.

On my calibnet SP: 5 stuck DropPiece tasks accrued over 32 hours, the oldest unschedulable for 38 hours, each pinned to a `skip = TRUE, complete = FALSE` piece with zero `sector_location` rows at any filetype, created in the same transaction as a streaming upload session that was later released. The scheduler logged the matching refusal continuously:

```
"did not accept task","task_ids":[69634852,69700952,69767126,70197805,70204539],
"reason":"CanAccept() refused","name":"DropPiece"
```

DropPiece itself was healthy throughout — 52 completions in the same window, zero failures. Only the locationless tasks strand.

## Notes on the change

The fix accepts a task when the piece has no location row anywhere: there is nothing on disk to be local to, so any node can run the deletion. Locally-stored pieces accept exactly as before, remotely-stored pieces are still refused, and `Do()` needs no change — it already deletes the DB row ref-guarded first and treats storage removal as best-effort. Race safety is unchanged: cleanup is only scheduled at `ref_count = 0`, every writer path holds a ref while writing, and the delete re-checks refs.

`LEFT JOIN LATERAL` rather than `NOT EXISTS` is deliberate: the anti-join form plans as a hash anti-join reading every piece-location row (75 ms over 21,283 rows on calibnet; a mainnet node here holds ~4M), and `CanAccept` runs on the scheduler goroutine where a slow query starves all task types (#1425). The lateral form probes `sectorlocation_pk` once per candidate: 1.5 ms measured.

## Verification

Running on the calibnet SP. On the first scheduler poll after restart, all 5 stuck tasks were claimed within 3 seconds of startup and completed successfully in under 120 ms each; the 5 orphaned `parked_pieces` rows are deleted and the refusal log line is gone. Pieces with real files continue to be accepted (checked against completed pieces on the same node) and the mainnet nodes show no regression.

### DIFF
--- a/tasks/piece/task_cleanup_piece.go
+++ b/tasks/piece/task_cleanup_piece.go
@@ -186,9 +186,13 @@ func (c *CleanupPieceTask) CanAccept(ids []harmonytask.TaskID, _ *harmonytask.Ta
 	err = c.db.QueryRow(ctx, `SELECT COALESCE(array_agg(cleanup_task_id), '{}')::bigint[] AS cleanup_task_ids FROM 
 										(
 										    SELECT pp.cleanup_task_id FROM parked_pieces pp
-											INNER JOIN sector_location l ON l.miner_id = 0 AND l.sector_num = pp.id AND l.sector_filetype = 32
-											WHERE cleanup_task_id = ANY ($1) 
-											  AND l.storage_id = ANY ($2)
+											LEFT JOIN LATERAL (
+												SELECT bool_or(l.storage_id = ANY ($2)) AS here, count(*) AS n
+												FROM sector_location l
+												WHERE l.miner_id = 0 AND l.sector_num = pp.id AND l.sector_filetype = 32
+											) loc ON TRUE
+											WHERE pp.cleanup_task_id = ANY ($1)
+											  AND (COALESCE(loc.here, FALSE) OR loc.n = 0)
 											  LIMIT 100
 										) s`, indIDs, storageIDs).Scan(&acceptedIDs)
 	if err != nil {
--- a/tasks/piece/task_cleanup_piece.go
+++ b/tasks/piece/task_cleanup_piece.go
@@ -185,14 +185,16 @@ func (c *CleanupPieceTask) CanAccept(ids []harmonytask.TaskID, _ *harmonytask.Ta
 	var acceptedIDs []harmonytask.TaskID
 	err = c.db.QueryRow(ctx, `SELECT COALESCE(array_agg(cleanup_task_id), '{}')::bigint[] AS cleanup_task_ids FROM 
 										(
-										    SELECT pp.cleanup_task_id FROM parked_pieces pp
+										    SELECT DISTINCT pp.cleanup_task_id FROM parked_pieces pp
 											LEFT JOIN LATERAL (
-												SELECT bool_or(l.storage_id = ANY ($2)) AS here, count(*) AS n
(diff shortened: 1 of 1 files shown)


LexLuthr (changes requested):
Just do a LEFT JOIN and match on $2 or NULL. There is no need to have counters to SELECT in a SELECT.

LexLuthr:
> Just do a LEFT JOIN and match on $2 or NULL. There is no need to have counters to SELECT in a SELECT.

```sql
SELECT COALESCE(array_agg(cleanup_task_id), '{}')::bigint[] AS cleanup_task_ids
FROM (
    SELECT DISTINCT pp.cleanup_task_id
    FROM parked_pieces AS pp
    LEFT JOIN sector_location AS sl
        ON sl.miner_id = 0
       AND sl.sector_num = pp.id
       AND sl.sector_filetype = 32
    WHERE pp.cleanup_task_id = ANY ($1)
      AND (
          sl.sector_num IS NULL
          OR sl.storage_id = ANY ($2)
      )
    LIMIT 100
) AS s;
sql

**Comment by TippyFlitsUK** [comment]
Adopted - your `WHERE (sl.storage_id IS NULL OR sl.storage_id = ANY ($2))` and the `DISTINCT`, counters gone. One deliberate deviation: the location lookup stays a `LATERAL` rather than a bare `LEFT JOIN`.

The reason is the plan, not taste. On my mainnet node's YugabyteDB the plain join form loses `sector_num` from the index condition and plans the join as a hash over every piece-location row:

```
->  Hash Right Join
      Hash Cond: (sl.sector_num = pp.id)
      ->  Index Scan using sectorlocation_pk on sector_location sl
            Index Cond: ((miner_id = 0) AND (sector_filetype = 32))
```

That's 3,996,960 rows on this node - a bare count of the same scan takes 3.4 s - inside the scheduler goroutine, the #1425 class, and this node ran 860 DropPiece tasks in the last 7 days so it would fire constantly. With the lateral, both my calibnet and mainnet nodes keep the per-piece probe:

```
->  Index Scan using sectorlocation_pk on sector_location sl
      Index Cond: ((miner_id = 0) AND (sector_num = pp.id) AND (sector_filetype = 32))
```

This looks like YB's heuristic cost model rather than anything wrong with the SQL - it ignores per-column stats, so the plain join never gets the parameterised path. Happy to change it if you'd rather take that trade or see a different shape!

LexLuthr (approved):

PR #1483, opened 2026-08-31
chore: remove filecoinpin

beck-8:
This is the subsequent step to complete the other parts.（Do not merge before other steps are completed.）
https://github.com/filecoin-project/filecoin-pin/issues/664

### DIFF
--- a/deps/config/types.go
+++ b/deps/config/types.go
@@ -127,8 +127,8 @@ func DefaultCurioConfig() *CurioConfig {
 					MaximumChunkSize:          256 * 1024 * 1024, // 256 MiB
 				},
 				IPNI: IPNIConfig{
-					ServiceURL:         []string{"https://cid.contact", "https://filecoinpin.contact"},
-					DirectAnnounceURLs: []string{"https://cid.contact/ingest/announce", "https://filecoinpin.contact/announce"},
+					ServiceURL:         []string{"https://cid.contact"},
+					DirectAnnounceURLs: []string{"https://cid.contact/ingest/announce"},
 				},
 			},
 		},
--- a/documentation/en/configuration/default-curio-configuration.md
+++ b/documentation/en/configuration/default-curio-configuration.md
@@ -835,13 +835,13 @@ description: The default curio configuration
       # The network indexer web UI URL for viewing published announcements
       #
       # type: []string
-      #ServiceURL = ["https://cid.contact", "https://filecoinpin.contact"]
+      #ServiceURL = ["https://cid.contact"]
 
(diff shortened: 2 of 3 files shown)


LexLuthr (changes requested):
We have a few special case handling for Filecoin pin in the index-provider. Let's remove them as well.

LexLuthr (approved):
Looks good. We can test is real world for any missed pin specific code paths.

beck-8:
Except for this link in one test, nothing else was found. This link is only used to test toml analysis, so I didn't touch it.

TippyFlitsUK:
Quick one on this PR. Should it have been merged yet? I have just spent the last 8 hours failing every dealbot deal on both of my mainnet nodes after moving them onto main to resolve #1499 . With filecoinpin.contact gone from the default DirectAnnounceURLs, new pieces stop reaching that indexer, and the dealbot still verifies there until dealbot#690 lands. The order in filecoin-pin#664 was dealbot first, then the switch.

I have resolved it by setting both URLs explicitly in the base layer, so I am back up. Just want to make sure nobody else on default config hits this in the middle of the night! 😉

beck-8:
I don’t think the merger here means we’ll be pushing it out to users.  Of course, the dealbot merger has also been delayed—we were originally scheduled to wrap it up last week, but it’s still not finished as of now.

Additionally, please note that you’ve explicitly configured two URLs. Originally, PoRep wouldn’t be propagated to Filecoin Pin, but now it will be as well; for the time being, this shouldn’t cause any issues.

TippyFlitsUK:
@beck-8 Your second paragraph turned out to be the problem. With both URLs set and the guard gone, main2's PoRep chain announced to filecoinpin.contact at 02:57 UTC; it spent the night ingesting ~9k PoRep ads and 4.5 GB of entries, and the dealbot's IPNI checks timed out for every SP from ~03:00 until I cut that chain off with a 404 at 11:46 UTC.

I frequently run main between releases to test the merged patches, and the fixes I am asked to test (#1504 for example) are PRs on main, so anything merged there is running on my nodes whether it has been released or not. Would it be worth holding this one until dealbot#690 lands?

LexLuthr:
@TippyFlitsUK How did this PR break the SP? Dealbot still using pin.contact?

TippyFlitsUK:
Yes, the dealbot still checks filecoinpin.contact ([dealbot#690](https://github.com/FilOzone/dealbot/pull/690) is open). With the URL out of the defaults, new pieces never reach it and every deal fails IPNI verification. Putting the URL back in config then hits the second change: the `info.SPID > 0` filter is gone, so main2's PoRep chain announced there too and its ~9k ads took the indexer down for everyone.

PR #1485, opened 2026-08-31
Fix: crash in tree-r

snadrus:
Leftover tmp files were confusing scrub, then supraseal's bad accesses was crashing curio!

This change:
- ignores tmp files now
- switches to using the crash-protection system

It was customer-verified and is running in production right now.


ZenGround0:
Needs some lint and gen-check

lib/ffi/scrub_commr.go
```diff
@@ -200,3 +206,31 @@ func (sb *SealCalls) getCachePath(ctx context.Context, s storiface.SectorRef, ft
...
+func validateCacheDir(path string) error {
+	st, err := os.Stat(path)
+	if err != nil {
+		return err
+	}
+	if !st.IsDir() {
+		return xerrors.Errorf("%s is not a directory", path)
+	}
+	if _, err := os.Stat(filepath.Join(path, "p_aux")); err != nil {
+		return xerrors.Errorf("p_aux in %s: %w", path, err)
```
ZenGround0: nit: "failed to stat p_aux in %x"

lib/paths/local.go
```diff
@@ -629,13 +629,15 @@ func (st *Local) declareSectors(ctx context.Context, p string, id storiface.ID,
...
+			info, err := ent.Info()
+			if err != nil {
+				log.Warnw("skipping unreadable storage entry", "path", p, "type", t, "name", ent.Name(), "err", err)
 				continue
 			}
 
-			sid, err := storiface.ParseSectorID(ent.Name())
-			if err != nil {
-				return xerrors.Errorf("parse sector id %s: %w", ent.Name(), err)
+			sid, ok := declareableSectorEntry(ent.Name(), t, info)
```
ZenGround0, line 638: I do not know this subsystem well but flagging that this is a behavior change.  I would expect us to log a warning here if we are in the !ok case

ZenGround0 (approved):
A couple of small suggestions and a reminder to run CI fixup chores

lib/paths/local.go
snadrus: The "or" case previously would crash.

Issue #1489, opened 2026-09-03
No IPNI removal advertisements are ever published for expired PoRep (MK12) deals


TippyFlitsUK:
Curio 1.28.6 (da6d1063), mainnet, f08403, YugabyteDB 2.25.1.0.

## What I am seeing

My PoRep deals on this node expired naturally between 19 May and 13 August 2026. Nothing was deleted by hand. The node's market IPNI publisher still advertises every one of them and, until today, had never published a single removal advertisement.

Numbers from the node's database on 3 September (chain epoch 6338228):

- 8,955 rows in `market_mk12_deals`, 4,296 with `end_epoch` still ahead, 4,659 expired
- 4,594 of the expired deals still hold `market_piece_deal` rows for this SP
- 4,591 of those pieces still carry a live (`is_rm = false`) advertisement on the market publisher chain
- 9,024 advertisements on the market publisher in total: 8,959 originals and 65 removals, and all 65 removals were published today by the manual process described below

## Why it matters

The chain was announced to cid.contact on 14 August. Within a day a fleet of IPFS gateways running the `rainbow` user agent started hitting this node's `/ipfs` endpoint for CIDs from those expired deals, and it has not stopped. The content is long gone, so almost every answer is a 404.

- 14 August: about 15,000 requests a minute. All four of my FOC second-copy jobs, which pull piece data from this node, went from completing normally on 13 August to timing out on every batch overnight, with the site uplink measured saturated before any repair traffic was added. One job's daily completions fell from 207 on 13 August to 36 on the 14th and 41 on the 15th.
- 26 August: 176,454,138 requests in the nginx access log for this host by 14:57 UTC, 31 GB of plaintext, which took the 199 GB pool holding it to 81% before I caught it. Of the last million lines, 96.8% were 404, 99.2% were `HEAD /ipfs`, 99.0% carried the `rainbow` user agent. Curio's own retrieval limiter had logged `Request limit reached` 1,357,294 times in the log tail I checked.
- 3 September, 03:22 UTC: nginx completed 296,977 requests in 60 seconds, 291,317 of them 404, from the OVH `rainbow` ranges, with 9,252 established connections on the proxy host. The one second-copy pull stream running at the time was getting 21.4 Mbit/s of a 92 Mbit/s uplink.
- 3 September, 16:15 UTC, one minute of the same log: 156,518 requests, 151,360 of them 404, 156,472 from `rainbow` or `kubo` user agents.

## Where the gap is

[...]

PR #1490, opened 2026-09-03
fix(indexstore, ipni): batch PieceBlockOffsetSize deletes in RemoveIndexes and keep piece_cleanup rows on their own IPNI scheduler

TippyFlitsUK:
Two fixes found while publishing IPNI removal advertisements for expired PoRep deals on a mainnet node (Curio 1.28.6, YugabyteDB 2.25.1.0). Background in #1489.

## RemoveIndexes

The last step was a single partition-wide `DELETE FROM PieceBlockOffsetSize WHERE PieceCid = ?`. On YugabyteDB that is one replicate operation, and once the partition is large enough it exceeds the leader's single op limit:

```
Operation replicate msg size (545878234) exceeds limit of leader side single op size (254017536)
```

The function already holds every `PayloadMultihash` for the piece from the select at the top, so the delete now goes through the same batched, idempotent path as the `PayloadToPieces` delete, using `InsertBatchSize` per batch and `executeBatchWithRetry`. The `len(batch.Entries) >= 0` check after the loop is always true and executed an empty batch; it is now `> 0`.

## IPNI schedulers

`task_ipni.go` and `task_pdp_ipni.go` both select from `piece_cleanup` where `after_cleanup = TRUE AND complete = FALSE` with no check on `pdp`. When a PoRep row (`pdp = false`) reaches that state the PDP scheduler can claim it and fail with `not original ad found for removal ad`, because the original advertisement lives on the market publisher chain. Each scheduler now takes only its own rows.

## Tested

Mainnet, f08403. On the unpatched binary three 8 GiB pieces failed nine times with the error above. All three completed after the change. Since the change went in, 63 PoRep pieces have been cleaned and 63 removal advertisements published through this path (38 of 32 GiB, 24 of 8 GiB, 1 of 4 GiB). No PDP rows were touched, no PDP removal advertisements were published, and the PDP IPNI scheduler has not claimed a row since the filter went in.

One note for anyone removing many large pieces at once. Per-row deletes leave one tombstone per row. After about 60 pieces, scans on the affected `PieceBlockOffsetSize` tablets hit the YCQL 60 second deadline (`Deadline for query passed`) until the table was compacted. Compaction cleared it.

## Files

- `market/indexstore/indexstore.go` (+15, -5)
- `tasks/indexing/task_ipni.go` (+1)
- `tasks/indexing/task_pdp_ipni.go` (+1)

### DIFF
--- a/market/indexstore/indexstore.go
+++ b/market/indexstore/indexstore.go
@@ -325,17 +325,27 @@ func (i *IndexStore) RemoveIndexes(ctx context.Context, pieceCidv2 cid.Cid) erro
 		}
 	}
 
-	if len(batch.Entries) >= 0 {
+	if len(batch.Entries) > 0 {
 		if err := i.executeBatchWithRetry(ctx, batch, pieceCidv2); err != nil {
 			return xerrors.Errorf("executing batch delete for PayloadToPieces for piece %s: %w", pieceCidv2, err)
 		}
 	}
 
 	// Delete from PieceBlockOffsetSize
-	delPieceBlockOffsetSizeQry := `DELETE FROM PieceBlockOffsetSize WHERE PieceCid = ?`
-	err := i.session.Query(delPieceBlockOffsetSizeQry, pieceCidBytes).WithContext(ctx).Exec()
-	if err != nil {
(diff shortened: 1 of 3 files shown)


LexLuthr (approved):
I am glad someone tested this. I was written as part of MK20 but was never tested in real world.

PR #1501, opened 2026-09-07
Remove small curio deletion limits

ZenGround0:


snadrus:
I presume we are waiting for MAX_ENQUEUED_REMOVALS to be removed from the contract first? https://github.com/FilOzone/pdp/issues/310

ZenGround0:
We could but not necessarily.  If we merge it as is the smart contract will hit the error when estimating gas.  

35 < 2000 so there is value in unleashing the limit up to the higher threshold.

If you think UX is too bad here we can reinstate the limit and make it match the on chain constant too.

ZenGround0:
But yeah if we just roll the constant to 2k we should keep this PR stale/in draft waiting for the PDPVerifier change.

jennijuju:
do we have a curio release before the next mega contract upgrade? if so, id prefer we ship this first

LexLuthr (approved):

jennijuju:
@ZenGround0 actually, before I click the merge button; wanna confirm that removing this would be fine for old (non compact) dataset as well?

ZenGround0:
All old datasets are now upgraded to handle unlimited deletions in the queue via processRemovals call so this should be totally safe for old and new datasets alike.

PR #1502, opened 2026-09-07
Supported flag check should not demand sm_80

snadrus:
The supported flag check should not demand sm_80. SPPark is broken here and not accepting my fix, so I've forked it and switched over filecoin-ffi's PR is not referenced here, so as a follow-up to this PR, the next filecoin-ffi release will need to be referenced (likely will happen automatically).


ZenGround0 (approved):
Very nice thank you.  Since CI is failing hard on install FFI step perhaps you need to actually get the latest ffi commit into a release.  Let me know if you need help trouble shooting.  Once CI is fixed we can merge.

PR #1505, opened 2026-09-08
market: honor unlimited MaxQueueSDR

hyunmoon:
## Problem

`MaxQueueSDR` documents zero as unlimited, but the SDR queue check currently treats any nonempty queue as backpressure when the limit is zero.

## Change

Apply the SDR threshold only when it is nonzero. Extract the existing queue decisions into a private helper used by the production path so the regression can run without a database.

The default remains `8`, positive limits still use `buffered > limit`, and DealSector, Trees, and PoRep pressure remain effective. SQL accounting, other queue policies, and cache behavior are unchanged.

## Validation

Database-free checks passed on upstream base `0bbdaf9ec3203e62a4d0a1e0e754864047736c7d`:

```sh
go test -tags=cgo,fvm,nosupraseal -count=1 ./market/backpressure ./deps/config
go test -race -tags=cgo,fvm,nosupraseal -count=1 ./market/backpressure ./deps/config
go vet -tags=cgo,fvm,nosupraseal ./market/backpressure ./deps/config
golangci-lint run -v --timeout 15m --concurrency 4 --build-tags cgo,fvm,nosupraseal ./market/backpressure ./deps/config
git diff --check
```

The regression covers zero with empty/nonempty queues, positive-limit boundaries, neighboring queue pressure, and unchanged negative-limit behavior. Lint used v2.11.4. PostgreSQL/Yugabyte execution was not performed; these tests exercise the production queue-decision helper, not SQL row effects.

### DIFF
--- a/market/backpressure/backpressure.go
+++ b/market/backpressure/backpressure.go
@@ -109,24 +109,28 @@ func (c *CachedBackPressure) checkSectorBackpressure(ctx context.Context, cfg *c
 		return false, xerrors.Errorf("counting buffered sectors: %w", err)
 	}
 
+	return sdrQueueBackpressure(cfg, bufferedSDR, bufferedTrees, bufferedPoRep, waitDealSectors), nil
+}
+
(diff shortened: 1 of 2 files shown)


hyunmoon:
Sorry. Codex went haywire.

PR #1507, opened 2026-09-09
enable 64 GiB piece for PDP

LexLuthr:
- [ ] Upload benchmarking
- [ ] Proving benchmarking

### DIFF
--- a/cuhttp/server_common.go
+++ b/cuhttp/server_common.go
@@ -59,7 +59,7 @@ func StartServer(ctx context.Context, cfg *config.HTTPConfig, db *harmonydb.DB, 
 		Addr:              cfg.ListenAddress,
 		Handler:           handler,
 		ReadTimeout:       cfg.ReadTimeout,
-		WriteTimeout:      time.Hour * 2,
+		WriteTimeout:      2*time.Hour + time.Minute,
 		IdleTimeout:       cfg.IdleTimeout,
 		ReadHeaderTimeout: cfg.ReadHeaderTimeout,
 	}
--- a/pdp/handlers_pull_test.go
+++ b/pdp/handlers_pull_test.go
@@ -17,6 +17,8 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/snadrus/must"
 	"github.com/stretchr/testify/require"
+
+	commcid "github.com/filecoin-project/go-fil-commcid"
 )
 
 // mockPullStore implements PullStore for testing
@@ -539,6 +541,43 @@ func TestHandlePull_Backpressure(t *testing.T) {
 	require.False(t, store.getPullPiecesCalled)
 }
 
+func TestHandlePullPieceSizeBoundary(t *testing.T) {
+	for _, test := range []struct {
+		name   string
+		raw    uint64
+		status int
+	}{
+		{name: "64 GiB padded", raw: 68182605824, status: http.StatusOK},
+		{name: "one byte above maximum", raw: 68182605825, status: http.StatusBadRequest},
+		{name: "64 GiB raw", raw: 64 << 30, status: http.StatusBadRequest},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			pieceCID, err := commcid.DataCommitmentToPieceCidv2(make([]byte, 32), test.raw)
+			require.NoError(t, err)
+			store := &mockPullStore{}
+			handler := NewPullHandler(&NullAuth{}, store, &mockValidator{shouldPass: true}, nil)
+			body := PullRequest{
+				ExtraData: testAddPiecesOnlyExtraData(t),
+				DataSetId: &testDataSetId,
+				Pieces: []PullPieceRequest{
+					{PieceCid: pieceCID.String(), SourceURL: "https://example.com/piece/" + pieceCID.String()},
+				},
+			}
+			req := httptest.NewRequest(http.MethodPost, "/pdp/piece/pull", bytes.NewReader(must.One(json.Marshal(body))))
+			rec := httptest.NewRecorder()
+			handler.HandlePull(rec, req)
+			require.Equal(t, test.status, rec.Code, rec.Body.String())
+			if test.status == http.StatusOK {
+				require.Len(t, store.createdPieces, 1)
+				require.Equal(t, test.raw, store.createdPieces[0].RawSize)
+			} else {
+				require.False(t, store.createPullCalled)
+				require.Contains(t, rec.Body.String(), "exceeds maximum")
+			}
+		})
+	}
+}
+
 func TestPadPieceSize(t *testing.T) {
 	// Test vectors from go-fil-commp-hashhash/testdata/zero.txt
 	// These are authoritative values: PayloadSize -> PieceSize
@@ -566,6 +605,8 @@ func TestPadPieceSize(t *testing.T) {
 		{name: "192 -> 256", rawSize: 192, expected: 256},
 		{name: "384 -> 512", rawSize: 384, expected: 512},
 		{name: "768 -> 1024", rawSize: 768, expected: 1024},
+		{name: "64 GiB padded maximum", rawSize: 68182605824, expected: 64 << 30},
+		{name: "one byte above 64 GiB padded", rawSize: 68182605825, expected: 128 << 30},
 	}
 
 	for _, tt := range tests {
--- a/tasks/pdpv0/task_prove.go
+++ b/tasks/pdpv0/task_prove.go
@@ -9,7 +9,6 @@ import (
 	"io"
 	"math/big"
 	"math/bits"
-	"sort"
 	"sync/atomic"
 	"time"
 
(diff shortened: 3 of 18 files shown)


snadrus (approved):

cuhttp/server_common.go
```diff
@@ -59,7 +59,7 @@ func StartServer(ctx context.Context, cfg *config.HTTPConfig, db *harmonydb.DB,
 		Addr:              cfg.ListenAddress,
 		Handler:           handler,
 		ReadTimeout:       cfg.ReadTimeout,
-		WriteTimeout:      time.Hour * 2,
+		WriteTimeout:      2*time.Hour + time.Minute,
```
ZenGround0, line 62: what's the deal with this?

pdp/handlers_pull_test.go
```diff
@@ -539,6 +541,43 @@ func TestHandlePull_Backpressure(t *testing.T) {
...
+			req := httptest.NewRequest(http.MethodPost, "/pdp/piece/pull", bytes.NewReader(must.One(json.Marshal(body))))
+			rec := httptest.NewRecorder()
+			handler.HandlePull(rec, req)
+			require.Equal(t, test.status, rec.Code, rec.Body.String())
+			if test.status == http.StatusOK {
+				require.Len(t, store.createdPieces, 1)
+				require.Equal(t, test.raw, store.createdPieces[0].RawSize)
+			} else {
+				require.False(t, store.createPullCalled)
+				require.Contains(t, rec.Body.String(), "exceeds maximum")
```
ZenGround0, line 575: nit: if this test gets expanded to other cases this string contain likely wont match.  This could be programmed into test cases in a string match field.

tasks/pdpv0/task_prove.go
```diff
@@ -317,27 +316,6 @@ func (p *ProveTask) Do(ctx context.Context, taskID harmonytask.TaskID, stillOwne
 		return false, xerrors.Errorf("failed to pack data: %w", err)
 	}
```
ZenGround0, line 318: This was quite helpful in its day but I guess we can get rid of this now yeah.

tasks/pdpv0/task_prove.go
```diff
@@ -522,17 +477,12 @@ func padTo32Bytes(b []byte) []byte {
...
 
-func (p *ProveTask) genSubPieceMemtree(ctx context.Context, subPieceCid string, subPieceSize abi.PaddedPieceSize) ([]byte, error) {
-	subPieceCidObj, err := cid.Parse(subPieceCid)
-	if err != nil {
-		return nil, xerrors.Errorf("failed to parse subPiece CID: %w", err)
-	}
-
-	if subPieceSize > proof.MaxMemtreeSize {
+func (p *ProveTask) genSubPieceMemtree(ctx context.Context, subPieceCid cid.Cid, subPieceSize abi.PaddedPieceSize) ([]byte, error) {
+	if subPieceSize > proof.MIN_PADDED_PIECE_SIZE_FOR_CACHE {
```
ZenGround0, line 481: Is this the right constant?  We are failing when we exceed MIN padded piece size for cache saying that we are exceeding a maximum, yet the constant is named minimum.  This is either incorrect or the name is too confusing.

tasks/pdpv0/task_prove.go
```diff
@@ -558,20 +508,8 @@ func (p *ProveTask) genSubPieceMemtree(ctx context.Context, subPieceCid string,
 	return proof.BuildSha254Memtree(r, subPieceSize.Unpadded())
 }
 
-// cprPieceReader adapts CachedPieceReader to the proof.PieceReader interface.
-// Converts v2 CIDs to v1 for pdpv0's v1-native CachedPieceReader.
-// NOTE: On main branch, GetSharedPieceReader accepts v2 natively so this
-// conversion can be dropped when pdpv0 merges.
```
ZenGround0, line 564: Glad someone got around to it eventually!

tasks/pdpv0/task_prove.go
LexLuthr: It is confusing. I should probably call it something else. Current names means minimum size of the piece at which cache is required but it is confusing. Especially if you have old context in head. I will come up with something better.

ZenGround0: That name actually makes sense then I am working my way back through this code and hadn't realized that this function will fail if the cache is needed.

tasks/pdpv0/task_prove.go
```diff
@@ -603,27 +541,26 @@ func (c *idxProofCache) GetLayer(ctx context.Context, pieceCidV2 cid.Cid, layerI
...
+		SubPieceOffset int64  `db:"sub_piece_offset"` // padded offset
+		SubPieceSize   int64  `db:"sub_piece_size"`   // padded piece size
+		Removed        bool   `db:"removed"`
+		PieceRawSize   uint64 `db:"piece_raw_size"` // raw size from parked_pieces, for constructing v2 CID
 	}
 
 	var subPieces []subPieceMeta
 
-	err := p.db.Select(context.Background(), &subPieces, `
-			SELECT ppr.id as pieceref_id, ppr.cached_proofgen_failure_count as cached_proofgen_failure_count,
```
ZenGround0, line 626: wow we were just reading this for no reason

tasks/pdpv0/task_prove.go
```diff
@@ -939,7 +849,7 @@ func (p *ProveTask) TypeDetails() harmonytask.TaskTypeDetails {
 		Cost: resources.Resources{
 			Cpu: 1,
 			Gpu: 0,
-			Ram: proveTaskRAM,
+			Ram: 100 << 20,
```
ZenGround0, line 852: Can you explain why we are reducing to 100 MiB?  The comment you've removed is pretty useful so please update your reasoning on why RAM is 10x smaller in a comment

tasks/pdpv0/task_prove.go
```diff
@@ -653,63 +590,60 @@ func (p *ProveTask) provePiece(ctx context.Context, dataSetId int64, pieceId int
 
 	var subPieceProof *proof.RawMerkleProof
 
-	isIdxNotNull := p.idx != nil
-	isLargeSubPiece := uint64(challSubPiece.SubPieceSize) > MinSizeForCache && challSubPiece.PieceRawSize > 0
-	isUsingCachedProof := isIdxNotNull && isLargeSubPiece
+	isLargeSubPiece := abi.PaddedPieceSize(challSubPiece.SubPieceSize) > proof.MIN_PADDED_PIECE_SIZE_FOR_CACHE
+
+	pcid1, err := cid.Parse(challSubPiece.SubPiece)
```
ZenGround0, line 595: I would not object to a subfunction parseCommitment

tasks/pdpv0/task_prove.go
```diff
@@ -653,63 +590,60 @@ func (p *ProveTask) provePiece(ctx context.Context, dataSetId int64, pieceId int
...
-		// First, let's extract the CIDs
-		if spCidV1, parseErr := cid.Parse(challSubPiece.SubPiece); parseErr == nil {
-			if spCidV2, v2Err := commcid.PieceCidV2FromV1(spCidV1, challSubPiece.PieceRawSize); v2Err == nil {
-				subpieceCidV1 = spCidV1
-				subpieceCidV2 = spCidV2
-				isPieceCidsExtracted = true
+	if isLargeSubPiece {
+		// Validate once after any cached-proof failure, including final verification.
+		defer func() {
+			if proofErr == nil {
```
ZenGround0, line 614: I would not object to a subfunction validateCacheUponError

tasks/pdpv0/task_prove.go
```diff
@@ -653,63 +590,60 @@ func (p *ProveTask) provePiece(ctx context.Context, dataSetId int64, pieceId int
...
-	}
 
-	// Fallback conditions for doing full memtree proof generation:
-	// 1. sub-piece is too small for caching (isLargeSubPiece == false), so we never even attempt cached proofgen
-	// 2. sub-piece is large, we attempted cached proofgen, but it fails for some reason
-	if subPieceProof == nil {
-		log.Debugw("using full memtree proof generation for sub-piece", "dataSetId", dataSetId, "pieceId", pieceId, "subPieceCid", challSubPiece.SubPiece, "subPieceSize", challSubPiece.SubPieceSize, "isLargeSubPiece", isLargeSubPiece, "isIdxNotNull", isIdxNotNull)
-		memtree, memErr := p.genSubPieceMemtree(ctx, challSubPiece.SubPiece, abi.PaddedPieceSize(challSubPiece.SubPieceSize))
+		subPieceProof = cachedProof
+
```
ZenGround0, line 643: nit remove new line

tasks/pdpv0/task_prove.go
ZenGround0: Yeah on rereading this now makes sense, feel free to ignore.  If you do anything here the error can be updated to something like "uncached code path taken to prove piece that must use cache.

tasks/pdpv0/task_save_cache.go
```diff
@@ -69,14 +63,6 @@ func (t *TaskPDPSaveCache) Do(ctx context.Context, taskID harmonytask.TaskID, st
 
 	task := tasks[0]
 
-	n, err := t.db.Exec(ctx, `UPDATE pdp_piecerefs SET caching_task_started = NOW() WHERE id = $1 AND needs_save_cache = TRUE`, task.ID)
```
ZenGround0, line 72: Can we now remove caching_task_started and caching_task_completed fields from the sql schema for pdp_piecrefs in a migration?

ZenGround0 (approved):

LexLuthr:
CC: @hugomrdias

cuhttp/server_common.go
LexLuthr: Ahh! This is basically 2 hours + 1 minute to allow larger uploads. 2 hour limit is enforced in piece upload and download. This adds 1 minute to allow system to wrap up things before terminating on a much lower network level. This is an edge case handling where upload took exactly 2 hours. Maybe I should not handled it?

ZenGround0 (approved):
Thanks!

PR #1509, opened 2026-09-09
market: schedule global MK20 stages through the deal poller

hyunmoon:
## Summary

Keep MK20 completion callbacks deal-local and schedule global aggregation/ingestion through the existing deal poller. Concurrent callbacks currently invoke those global stages directly and can overlap their passes.

`signalNextMK20` now processes only the requested deal's loaded pieces, then requests a coalesced poller wake. The full piece pass marks completed downloads before loading fresh rows; individual piece callbacks no longer run that global marking query. The existing full-poll order, idle/wake timing and cancellation behavior remain unchanged.

This is process-local poller ownership, not cross-process allocation safety or exactly-once processing. It adds no scheduler, configuration or schema and has no dependency on another proposed patch.

## Validation

Validated on upstream base `86e95967cd602dbf8db5ea538abe986c3968e2e3`, topic `065f0d0973f3df02873ebbce573506f015caf251`, with Go 1.26.1:

```sh
go test -p=2 -tags=cgo,fvm,nosupraseal -count=1 -timeout=3m \
  -skip '^(TestIntegration_CheckExpiry|TestIntegration_FailMK20DealBeforeSector)$' ./tasks/storage-market
go test -race -p=2 -tags=cgo,fvm,nosupraseal -count=1 -timeout=3m \
  -skip '^(TestIntegration_CheckExpiry|TestIntegration_FailMK20DealBeforeSector)$' ./tasks/storage-market
go vet -p=2 -tags=cgo,fvm,nosupraseal ./tasks/storage-market
```

All passed in a sanitized, database-free environment. The two explicitly excluded pre-existing integration tests can connect through fallback database settings and are unrelated to this scheduling change.

Executable tests cover mark-before-load, requested-deal processing, piece/load errors, cancellation, wake coalescing and shutdown. AST call-site tests additionally protect full-poll ordering and the absence of direct global stages in completion callbacks; these are source-shape checks, not SQL concurrency evidence. No database, chain API or actual sealing execution is claimed for this topic's new tests.

### DIFF
--- a/market/mk20/DEVELOPER_FLOW_MAP.md
+++ b/market/mk20/DEVELOPER_FLOW_MAP.md
@@ -119,7 +119,7 @@ If deal has `Data` and matching piece already exists in `parked_pieces`, it inse
    - Offline: create pipeline row with `offline=TRUE, started=FALSE`.
    - Aggregate: one pipeline row per subpiece; HTTP subpieces also get download refs.
 3. `processMK20Deals` then progresses data readiness:
-   - `downloadMk20Deal` calls SQL `mk20_ddo_mark_downloaded` to bind completed refs to `url=pieceref:*` and set `downloaded=TRUE`.
+   - The full piece pass calls SQL `mk20_ddo_mark_downloaded` once before loading its pipeline snapshot. The function binds completed refs to `url=pieceref:*` and sets `downloaded=TRUE`.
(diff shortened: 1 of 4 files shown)


hyunmoon:
Sorry. Codex went haywire.

PR #1510, opened 2026-09-09
seal: exclude failed sectors from precommit batching

hyunmoon:
## Summary

Exclude failed sectors before PreCommit batch numbering and recheck the failed
state when assigning a batch. If an already-created task contains failed
sectors, detach only their task/provider/sector assignments and preserve the
failure evidence. Send and associate a CID only with the remaining sectors;
an all-failed task completes without sending a message.

Batch size, ordering, timeout/slack policy, sector schedules and expiration
calculation are unchanged. This does not provide exactly-once chain submission
or repair historical tasks.

## Validation

- `go test -p=2 -tags=cgo,fvm,nosupraseal ./tasks/seal -count=1 -timeout=3m`
- Same package with `-race`: PASS.
- Exact-tag `go vet`, golangci-lint 2.11.4, repository fiximports and diff check: PASS.
- Production-SQL tests on disposable PostgreSQL 16.15, reported Read Committed:
  `TestPrecommitSQLCandidatesAndAssignment` and
  `TestPrecommitSQLDetachAndCIDMembership`: PASS.
- Removing the failed candidate predicate produced the expected row-count and
  ordering regression; the restored baseline passed.
- `SubmitPrecommitTask.Do` tests inspect the actual outgoing CBOR using a fake
  chain API/sender. SQL tests execute production constants; they are row-effect
  and rollback tests, not deliberate cross-connection contention tests.

Yugabyte execution of the new SQL tests and live actor/message execution were
not performed. Opt-in target safeguards and exact test invocation are documented
in `documentation/en/precommit-sql-tests.md`.

### DIFF
--- a/documentation/en/precommit-sql-tests.md
+++ b/documentation/en/precommit-sql-tests.md
@@ -1,0 +1,43 @@
+# PreCommit SQL row-effect tests
+
+The opt-in `integration` tests execute the same candidate, assignment, failed
+detach, and message-CID SQL constants used by the PreCommit poller/task. They
+complement the database-free `SubmitPrecommitTask.Do` tests, whose fake API and
+sender inspect actual outgoing CBOR membership. They do not submit messages,
+prove chain acceptance, or establish exactly-once send/persistence.
+
+Use only an explicitly disposable local PostgreSQL or Yugabyte database. Remove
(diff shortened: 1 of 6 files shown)


hyunmoon:
Sorry. Codex went haywire.

PR #1512, opened 2026-09-09
storageingest: bound and serialize ready-sector transfers

hyunmoon:
# storageingest: bound and serialize ready-sector transfers

Ready-sector transfer currently scans and moves an unbounded provider set in one transaction without sharing a serialization point with open-sector allocation.

This change:

- transfers at most 64 eligible sectors per transaction, preserving full-sector, maximum-wait and deadline eligibility for both SDR and Snap;
- continues in fair provider rounds and keeps the ten-second fallback poll plus coalesced wakeups;
- uses a preserving write to `sectors_allocated_numbers` for first-use allocation, open-sector mutation, manual seal-now and transfer;
- acquires multiple SupraSeal provider locks in ascending provider order, without changing weighted counts, remainder behavior or proof selection;
- keeps pipeline creation, initial-piece transfer and open-piece deletion atomic, retains inconsistent rows as evidence, and isolates only deterministic candidate errors rather than retrying every sector during a database outage.

The 64-sector quantum is not a sealing WIP cap. Participating writers must use the same locking contract; older binaries or direct SQL bypasses are outside this guarantee. Existing chain reads inside some single-provider allocation transactions can extend lock duration. CC quota/remainder correction is a separate change and is not included here.

## Validation

- Database-free normal/race tests: `market/storageingest`, `tasks/seal`, `tasks/sealsupra`, `tasks/storage-market`; exact unsafe upstream fallback-DB tests excluded by name.
- Exact `cgo,fvm,nosupraseal,integration` vet/lint and test-binary compilation: PASS.
- Disposable PostgreSQL 16.15, reported Read Committed: actual first-use allocator, two-handle 129-sector drain, inconsistent-state preservation/continuation, Snap transfer and late-transfer rollback: PASS.
- Deterministically held provider transaction: linked `pg_blocking_pids` observation established that the actual allocator waited before holder release; both committed distinct sector numbers. This added test also passed with the race detector.
- The late Snap failure test verifies the entire failed batch rolled back, then only the valid candidate moves during bounded fallback; race PASS.
- Historical operator-reported YugabyteDB 2025.2.2.2-b11 RC/wait-queue allocation/transfer execution is separate evidence. Current-source Yugabyte execution: NOT RUN. PostgreSQL-specific observation is not presented as a Yugabyte observer.

No schema, public configuration, scheduling threshold, deal schedule or chain-message behavior change is introduced. Current normal production sealing is compatibility evidence, not deliberate contention/failure-path proof.

### DIFF
--- a/market/storageingest/deal_ingest_batch.go
+++ b/market/storageingest/deal_ingest_batch.go
@@ -1,0 +1,510 @@
+package storageingest
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/yugabyte/pgx/v5/pgconn"
+	"golang.org/x/xerrors"
+
+	"github.com/filecoin-project/go-state-types/abi"
+
+	"github.com/filecoin-project/curio/harmony/harmonydb"
+	"github.com/filecoin-project/curio/tasks/seal"
+)
+
+// sealBatchSize bounds a ready-sector transfer transaction. It is an
+// internal work quantum, not a limit on sealing work in progress.
+const sealBatchSize = 64
+
+const selectSealCandidatesSQL = `SELECT sector_number
+	FROM open_sector_pieces
+	WHERE sp_id = $1
+	  AND NOT (sector_number = ANY($3::bigint[]))
+	GROUP BY sector_number
+	HAVING BOOL_AND(is_snap = $2)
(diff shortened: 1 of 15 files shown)


hyunmoon:
Sorry. Codex went haywire.

PR #1513, opened 2026-09-09
indexing: wait for sector offsets before assigning tasks

hyunmoon:
## Summary

Require a non-NULL sector offset before assigning Indexing work for both MK12
and MK20. The producer can leave the offset unavailable while `IndexingTask.Do`
loads it into a non-nullable integer, so assignment must wait for this prerequisite.

Guard both the candidate CTE before ORDER BY/LIMIT and the conditional UPDATE.
This prevents an older unready row from hiding a later ready row and avoids
relying solely on advisory selection. Zero is a valid offset; it is not excluded
or substituted for NULL. Existing task/provider/deal/aggregate identities,
sealed/indexed/task conditions, ordering and cadence are unchanged.

Metadata-only/no-physical-indexing rows still receive their required metadata
processing. No offsets or completion markers are fabricated. Assigned/exhausted
tasks, retry limits and historical state are unchanged. Post-GC discovery and
recovery are separate work, not included here.

## Validation

On upstream `86e95967cd602dbf8db5ea538abe986c3968e2e3`, executable extracted
source `ac6bf2dc1d7d0c43cbbcd2009f40b00ce010ac0f` (later changes are docs only):

- `go test` and `go test -race -tags=cgo,fvm,nosupraseal -count=1
  -timeout=3m ./tasks/indexing`: PASS in a database-free environment.
- Exact integration-tag file selection/compilation, vet, pinned golangci-lint
  2.11.4, canonical fiximports and diff check: PASS.
- Disposable PostgreSQL 16.15, reported Read Committed, exact production SQL:
  `TestIndexingOffsetSQLReadiness`, `TestIndexingOffsetSQLAllNullThenReady`,
  `TestIndexingOffsetSQLAssignmentRollback`: all MK12/MK20 cases PASS.
  Tests cover NULL vs zero/positive, metadata-only rows, existing tasks,
  completed/not-ready rows, unrelated composite identities and rollback.
- In a throwaway checkout, removing only the candidate-side non-NULL predicate
  caused the later-ready-row assertion to fail for each market: expected one
  affected row, got zero. Both restored baselines passed. These are executable
  SQL row effects, not regex/model evidence or an invented interleaving inside
  one atomic statement.

The fixture is opt-in (`integration` tag plus dedicated target variables), uses
literal loopback, disabled load balancing, owned random schemas and finite
timeouts. Current-head Yugabyte execution is NOT RUN; another engine run is not
required before code review for this narrow predicate change given actual
PostgreSQL row effects, negative controls and unchanged query identity semantics.
Neither ongoing production sealing nor the earlier cordoned-worker canary
reproduced this NULL-offset bug.

### DIFF
--- a/documentation/en/market-2.0/indexing-offset-sql-tests.md
+++ b/documentation/en/market-2.0/indexing-offset-sql-tests.md
@@ -1,8 +1,11 @@
 # Indexing Offset Readiness: Operator SQL Verification
 
 This follow-up adds executable SQL assertions to the existing NULL-offset
-readiness fix. PostgreSQL and YugabyteDB execution remains **NOT RUN** until
-an operator runs these tests in a separately authorized disposable database.
+readiness fix. PostgreSQL 16.15 execution passed at follow-up `6dace4e4`
+and current-upstream extraction `ac6bf2dc`, including all six MK12/MK20
+baseline subtests and the candidate-side negative controls described below.
+YugabyteDB execution remains **NOT RUN**. Future execution requires a
+separately authorized disposable database.
 No production target, API, task engine, payload processing, or recovery is used.
 
 ## Statements and Coverage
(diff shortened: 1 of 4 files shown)


hyunmoon:
Sorry. Codex went haywire.

PR #1514, opened 2026-09-09
market: account for preceding pieces in MK20 sector offsets

hyunmoon:
## Summary

Correct MK20 sector-offset accumulation for a target preceded by other pieces. The old loop advances past a piece only inside the matching branch, immediately before returning, so preceding nonmatching pieces do not contribute to the offset. A first/single-piece fixture can mask this bug.

The production `addDealOffset` path now uses a small private calculation helper that applies existing padding, returns the matching piece's aligned start, and advances past every nonmatching predecessor. It preserves authoritative `piece_index` order, padded/unpadded units, CID-and-size matching and first-match behavior, the existing transaction/NULL-only update, and empty/not-found handling. The MK12 path is unchanged.

No retained-metadata fallback, stored-offset repair, assignment, scheduling or completion-policy change is included. This patch has no dependency on another proposed patch.

## Validation

Validated on upstream base `86e95967cd602dbf8db5ea538abe986c3968e2e3`, topic `7aa54787d6042d20d76e2b9c4100c8ef1e5c1b64`, with Go 1.26.1:

```sh
go test -p=2 -tags=cgo,fvm,nosupraseal -count=1 -timeout=3m \
  -skip '^(TestIntegration_CheckExpiry|TestIntegration_FailMK20DealBeforeSector)$' ./tasks/storage-market
go test -race -p=2 -tags=cgo,fvm,nosupraseal -count=1 -timeout=3m \
  -skip '^(TestIntegration_CheckExpiry|TestIntegration_FailMK20DealBeforeSector)$' ./tasks/storage-market
go vet -p=2 -tags=cgo,fvm,nosupraseal ./tasks/storage-market
```

All passed in a sanitized, database-free environment; the two excluded pre-existing integration tests can use fallback database settings and are unrelated to the arithmetic.

The executable production-helper tests cover first/single-piece zero; `[512,256,128]` offsets `[0,512,768]`; equal-size distinct pieces; an arithmetic alignment fixture `[128,256,128]`; CID-and-size matching; missing/empty lists; and input preservation. A source-level call-site regression protects the helper's use by `addDealOffset`.

In a throwaway negative control, removing predecessor accumulation made the later-piece assertion fail (`expected 512, actual 0`); restoring the implementation passed. This is executable arithmetic evidence, not a compilation failure or a database model. PostgreSQL/Yugabyte execution is not claimed or required to establish this pure calculation. Existing non-NULL offsets and post-GC discovery remain outside scope.

### DIFF
--- a/tasks/storage-market/mk20.go
+++ b/tasks/storage-market/mk20.go
@@ -842,17 +842,32 @@ func (d *CurioStorageDealMarket) createCommPMk20Piece(ctx context.Context, piece
 	return nil
 }
 
+type mk20SectorPiece struct {
+	CID   string              `db:"piece_cid"`
+	Size  abi.PaddedPieceSize `db:"piece_size"`
+	Index int64               `db:"piece_index"`
+}
+
+func findMK20PieceOffset(pieces []mk20SectorPiece, pieceCID string, pieceSize abi.PaddedPieceSize) (abi.PaddedPieceSize, bool) {
+	var offset abi.UnpaddedPieceSize
+
+	for _, piece := range pieces {
+		_, padLength := proofs.GetRequiredPadding(offset.Padded(), piece.Size)
+		offset += padLength.Unpadded()
+		if piece.CID == pieceCID && piece.Size == pieceSize {
+			return offset.Padded(), true
+		}
+		offset += piece.Size.Unpadded()
+	}
+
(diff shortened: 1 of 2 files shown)


hyunmoon:
Sorry. Codex went haywire.

PR #1516, opened 2026-09-09
market: revalidate MK20 deals before sector allocation

hyunmoon:
## Cumulative draft dependency

Depends on [storageingest: bound and serialize ready-sector transfers](https://github.com/filecoin-project/curio/pull/1512), pinned parent head `87871897619cb7ebe481c22bda65f3176cc3a22a`.
This draft targets upstream `main` and intentionally includes the unmerged parent's commits. The incremental review is `87871897619cb7ebe481c22bda65f3176cc3a22a..a2792d04f33168d6f6a5d548e793fb436daf5f5e`. Both drafts are submitted together; merge the parent first. No unrelated policy or feature stack is included.

# market: revalidate MK20 deals before sector allocation

Two callers can discover the same ready MK20 row before either assigns it. The second caller must not allocate another sector from that stale discovery and overwrite the first assignment.

Candidate discovery becomes advisory. The production retryable transaction takes the shared provider lock, locks the deal's pipeline rows, requires exactly one eligible aggregate identity, allocates and conditionally persists the exact `(id, aggr_index, sp_id)` mapping. Allocation/open-piece writes roll back if the mapping does not affect exactly one still-unassigned row. Only a committed assignment wakes ingestion.

This preserves nonzero aggregate indices, upstream expiry-before-assignment behavior, nil/explicit StartEpoch, requested Duration, allocation validation, piece identity and notifications. It is not cross-deal CID/allocation deduplication, a new admission cap or chain-message exactly-once policy.

## Validation

- Production attempt/orchestration unit tests: stale discovery, invalid/missing claim, allocation/persistence errors, retry-local result isolation, commit-only wake, provider/composite identity and unexpected aggregate count: normal/race PASS.
- Upstream nil/explicit start-boundary test remains PASS. Actual expiry DB tests with unsafe inherited/default targets were excluded by exact name; their production checks remain unchanged.
- Exact `cgo,fvm,nosupraseal,integration` vet/lint and test compilation: PASS.
- Disposable PostgreSQL 16.15, reported Read Committed: concurrent production transaction assignment, allocation rollback after failed conditional persistence, and deterministically linked provider-lock waiting: PASS. The allocator callback writes synthetic open pieces; it is not the full chain-aware storage allocator.
- The held production attempt committed sector 42; its waiting contender then skipped without allocating or waking. This added test passed under the race detector.
- Executable SQL negative control: temporarily removing the assigned-sector/proof readiness check caused the stale contender tests to fail; the exact source was restored and all assignment SQL baselines passed again.
- Historical operator-reported Yugabyte assignment/rollback evidence is not a current-source DB PASS. Current-source Yugabyte execution: NOT RUN. The new linked observer is PostgreSQL-specific.

The provider write is the shared conflict point. Row locks do not upgrade Read Committed to serializable, and snapshot/conflict behavior remains engine-specific. Current normal production sealing is compatibility evidence, not deliberate assignment-race proof.

### DIFF
--- a/market/storageingest/deal_ingest_batch.go
+++ b/market/storageingest/deal_ingest_batch.go
@@ -1,0 +1,510 @@
+package storageingest
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/yugabyte/pgx/v5/pgconn"
+	"golang.org/x/xerrors"
+
+	"github.com/filecoin-project/go-state-types/abi"
+
+	"github.com/filecoin-project/curio/harmony/harmonydb"
+	"github.com/filecoin-project/curio/tasks/seal"
+)
+
+// sealBatchSize bounds a ready-sector transfer transaction. It is an
+// internal work quantum, not a limit on sealing work in progress.
+const sealBatchSize = 64
+
+const selectSealCandidatesSQL = `SELECT sector_number
+	FROM open_sector_pieces
+	WHERE sp_id = $1
+	  AND NOT (sector_number = ANY($3::bigint[]))
+	GROUP BY sector_number
+	HAVING BOOL_AND(is_snap = $2)
+	   AND (SUM(piece_size) = $4
+	        OR MIN(created_at) < $5
+	        OR MIN(COALESCE(direct_start_epoch, f05_deal_start_epoch, 0)) < $6)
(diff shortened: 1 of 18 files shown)


hyunmoon:
Sorry. Codex went haywire.

PR #1517, opened 2026-09-09
web: preserve configuration schema and explicit layer overrides

hyunmoon:
## Problem

The Configuration editor can emit root-relative references to definitions that
exist only inside a reflected `Dynamic[T]` subschema. Separately, saving a sparse
layer through a default-filled typed configuration can remove explicit zero,
false and default-valued overrides. Those values may intentionally override an
earlier layer.

## Changes

- Reflect Dynamic inner values as self-contained schema fragments and supply
  help from the actual model/documentation tree.
- Validate Curio layers with the existing runtime decoder, then preserve their
  sparse explicit values and canonical key names. Unknown existing/submitted
  keys fail closed instead of disappearing.
- Correct FIL/duration editor constraints; preserve empty array strings and
  disable unsafe saves when the editor loses or changes a loaded value.
- Recursively compare the actual model and GET schema, without a per-field
  allowlist. Independent tests cover nested structures and load/edit/save/load.

No config option, public default, runtime loader policy or database schema changes.
Sparse save changes target the Curio editor; Skiff's existing projection contract
is retained. This does not add optimistic locking between simultaneous editors.

## Validation

- Go normal/race tests for `web/api/config` and `deps/config`; Skiff config
  package tests; exact-tag vet, pinned golangci-lint and import/diff checks: PASS.
- Actual JSON Editor browser fixture, using the emitted schema and synthetic
  layers/defaults with all requests intercepted locally: load/no-edit and
  one-field edit/save representation PASS. No database POST was executed.
- Executed negative controls: old Dynamic reflection fails at `Addresses[]`;
  deleting a schema property fails recursive completeness; dropping an explicit
  boolean on save fails semantic round-trip. All mutations restored and tests PASS.
- PostgreSQL/Yugabyte config-layer SQL execution: NOT RUN. No generator input
  changes; generated default documentation equality is checked by the tests.

Independent of sealing, waiting-release and optional pacing changes.

### DIFF
--- a/documentation/en/SUMMARY.md
+++ b/documentation/en/SUMMARY.md
@@ -15,6 +15,7 @@
 * [Curio Service](curio-service.md)
 * [Storage Configuration](storage-configuration.md)
 * [Configuration](configuration/README.md)
+  * [Configuration editor fidelity](configuration/configuration-editor.md)
   * [Listen Address](configuration/listen-address.md)
   * [Prometheus Metrics](configuration/prometheus-metrics.md)
   * [Metrics Reference](configuration/metrics-reference.md)
--- a/documentation/en/configuration/configuration-editor.md
+++ b/documentation/en/configuration/configuration-editor.md
@@ -1,0 +1,58 @@
+# Configuration editor fidelity
+
+The Curio Configuration editor reflects the current Curio configuration model.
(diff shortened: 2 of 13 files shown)


hyunmoon:
Sorry. Codex went haywire.

PR #1518, opened 2026-09-09
web: expose bounded Cluster Tasks snapshots and current-attempt runtime

hyunmoon:
## Summary

Cluster Tasks currently loads an unbounded list and mixes posting/ownership age
with work duration. This adds a bounded monitoring endpoint and a view that
separates current-attempt **Took** from pending **Waiting**.

- At most 500 selected rows, filtering before selection, truthful totals and
  batched multi-provider enrichment. Existing RPC fields keep their meanings.
- Explicit running, awaiting-start and unknown timing states. Confirmed Took
  shares the exact Do-entry timestamp used by new History records; no backfill.
- Task/owner/attempt CAS rejects delayed writes from prior attempts. Ownership
  transitions clear telemetry; same-owner recovery prepares a fresh token.
- Refresh/pause/filter/coalescing controls, bounded polling, stale response
  rejection and monotonic interpolation without per-second RPCs. Took and Waiting
  use start alignment, including unknown/dash values.

The many small task-file changes implement one batched provider-lookup interface;
they do not change task scheduling policy. This topic has no dependency on SDR
pacing, admission caps, offset recovery, or site-specific profiles.

## Execution and rollout boundary

This is more than a GUI-only change. Preparation adds one conditional UPDATE per
accepted task before storage acquisition. Failed preparation prevents dispatch
and attempts a bounded ownership release; preparation and release each have a
five-second batch budget, including for time-sensitive tasks. The Do-entry writer
adds one asynchronous conditional UPDATE and a bounded goroutine, canceled and
joined even when Do panics. Its own panic is contained. Failed telemetry does not
invent task success or a live timestamp.

The additive migrations preserve existing fields/history and never backfill
execution timestamps. Already instrumented schemas use the same migration names
and SQL, without duplicate triggers. Older workers cannot report Do entry, and
old same-owner recovery without an ownership write cannot be recognized as a new
attempt. Worker/database clocks must be synchronized. This is task-body time,
not an FFI benchmark. Database counts can scan beyond the selected-row limit.

## Validation

- Go normal/race tests for `harmony/harmonytask` and `web/api/webrpc`: PASS with
  `cgo,fvm,nosupraseal`; includes real scheduler preparation-failure and bounded
  writer cancellation/panic tests.
- Exact `integration`-tag vet/lint, import formatting and diff checks: PASS.
- 41 Node tests: PASS for request ordering, hidden/pause/failure/resume, stale
  clocks, retry timing and response shaping.
- Actual Chromium component/Shadow DOM at 642x500 and 1280x760, coalescing on/off:
  PASS. Synthetic data only; computed start alignment, duration/unknown/dash
  values and adjacent column bounds verified. Original right alignment fails
  the same assertion.
- Disposable PostgreSQL 16.15: `TestTaskAttemptSQLMigrationAndIdentity`,
  `TestTaskAttemptSQLDoEntryAndRollback` and `TestClusterTaskSnapshotSQL`: PASS.
  These execute actual migrations, HarmonyDB adapter and snapshot SQL. Replacing
  the token-CAS predicate makes the delayed-writer assertion fail; restored PASS.
- Yugabyte execution: NOT RUN. Independent-handle stale-writer checks establish
  conditional row effects, not simultaneous lock contention or serializability.

The SQL fixtures require explicit `CURIO_TASK_ATTEMPT_ITEST=1`, dedicated target
fields, literal loopback, disabled load balancing and isolated schema ownership.
No production database, service or worker was accessed for these tests.

### DIFF
--- a/documentation/en/developer/cluster-task-monitoring.md
+++ b/documentation/en/developer/cluster-task-monitoring.md
@@ -7,7 +7,10 @@ older migrations remain unknown. No historical task rows are backfilled.
 
 `ClusterTaskSummaryLimited` provides a bounded snapshot (at most 500 rows),
 filtering before selection, running/pending counts, a database observation time,
-and nullable ages. The generic view orders each section by ownership/posting age
+and nullable ages. Owned rows display current-attempt Took, with an explicit
+awaiting-start/unknown state; Pending rows display posted-based Waiting. See
+[current-attempt monitoring](task-attempt-monitoring.md) for worker instrumentation,
+mixed-version compatibility and failure behavior. The generic view orders each section by ownership/posting age
 then task ID, without a task-type priority policy. Pending rows can use whatever
 capacity remains after running rows. Controls can reduce that preview to zero.
 The row bound is not a bound on database scan cost.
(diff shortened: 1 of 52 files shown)


hyunmoon:
Sorry. Codex went haywire.

PR #1519, opened 2026-09-09
seal: add optional per-instance SDR start pacing

hyunmoon:
## Summary

Add optional per-instance SDR start pacing without changing the default unpaced scheduler. Operators can set a minimum interval between SDR task `Do` entries and optionally spread first/long-idle entry using a stable phase.

`SealSDRMinStartInterval` defaults to zero and `SealSDRStartJitter` to false. Positive-interval jitter requires a stable `CURIO_NODE_NAME` and instance listen identity; configuration is applied at task construction, not hot-reloaded.

## Scheduling contract

- Reserve only after ordinary resource/eligibility checks and before ownership/storage claims; speculative or cached `CanAccept` does not consume an interval.
- Cancel token-owned reservations after claim/storage failures or pre-entry cancellation. Commit immediately before `Do`; later task failure does not refund the interval.
- Preserve unpaced batch behavior. Pacing serializes provisional starts on one process, not task execution or cluster-wide admission.
- Use monotonic elapsed time for spacing and a latched wall-clock-derived phase. Restart forgets in-memory spacing; phases can collide. No scheduler sleeps, new DB writes, migrations, or background services.
- Emit startup, rate-limited blocked-reason/remaining/next-eligible estimates, and committed-entry diagnostics only after releasing admission and diagnostic locks.

This is not a native SDR computation-start guarantee, a global rate limiter, a sector quota, or a change to CPU accounting, storage capacity, FFI backends, UnsealSDR, or SupraSeal batches. There is no dependency on task-timing telemetry or a UI patch.

## Validation

On Go 1.26.1 with `cgo,fvm,nosupraseal`:

- Normal and race tests: `./deps/config ./harmony/harmonytask ./tasks/seal`.
- Matching-tag vet and scoped golangci-lint 2.11.4, including `./cmd/curio/tasks`; task wiring build.
- Config-document generation, generated default-field verification, canonical import formatting, and diff checks.
- Deterministic reservation, cancellation, failure, recovery, idle/restart, identity, phase, default, and real scheduler-admission regressions.
- Executed negative controls: mutating pacing state during diagnostics failed every read-only stage assertion; logging under the pacer mutex failed the lock assertion without hanging. Both mutations were discarded and the restored baseline passed.

Database-free tests exercise production admission/orchestration with injected ownership/storage outcomes; they are not PostgreSQL/Yugabyte concurrency or native CUDA execution evidence. Linux role builds and operational interval selection remain separate validation.

### DIFF
--- a/cmd/curio/tasks/tasks.go
+++ b/cmd/curio/tasks/tasks.go
@@ -613,6 +613,9 @@ func addSealingTasks(
 		sdrMax := taskhelp.Max(cfg.Subsystems.SealSDRMaxTasks)
 
 		sdrTask := seal.NewSDRTask(full, db, sp, slr, sdrMax, cfg.Subsystems.SealSDRMinTasks)
+		if err := sdrTask.ConfigureStartPacing(cfg.Subsystems.SealSDRMinStartInterval, cfg.Subsystems.SealSDRStartJitter, os.Getenv("CURIO_NODE_NAME"), machineHostPort); err != nil {
+			return nil, nil, sp, nil, nil, nil, nil, xerrors.Errorf("configuring SDR start pacing: %w", err)
+		}
 		keyTask := unseal.NewTaskUnsealSDR(slr, db, sdrMax, full)
 
 		activeTasks = append(activeTasks, sdrTask, keyTask)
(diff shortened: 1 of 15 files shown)


hyunmoon:
Sorry. Codex went haywire.

PR #1520, opened 2026-09-09
sealsupra: bound CC allocation by provider quota

hyunmoon:
Depends on [provider-state locking and bounded transfers](https://github.com/filecoin-project/curio/pull/1512)
at `87871897619cb7ebe481c22bda65f3176cc3a22a`.
This draft targets upstream `main`, so its diff currently includes that parent.
The incremental review is `87871897619cb7ebe481c22bda65f3176cc3a22a..3ec0ca908b74fdb81abe4a58c95ffa0fd3152477`: CC quota planning/debit correctness and its tests/docs;
it does not include the authoritative MK20-assignment child.

## Problem and change

The old weighted planner assigns its remaining slots to the last provider without
respecting that provider's remaining quota. The quota update also accepts stale
or disabled schedules unconditionally.

Cap every weighted share at remaining provider quota, then distribute rounding
and capped-share remainders in deterministic schedule order. Recheck enabled
state and sufficient quota for the exact provider when debiting; an unexpected
row count aborts the existing allocation transaction.

This deliberately changes non-exact rounding: weights `3,2,1` and ten available
slots become `6,3,1`, not `5,3,2`, when all quotas suffice. Exact weighted shares
are unchanged. It preserves the parent's ascending provider-lock order, chain
reads, proof selection and stored duration; no new lock domain or configuration.

## Validation

- Production planner/debit-result unit and race tests; exact integration-tag
  vet/lint, selected SQL test compilation and formatting/diff checks: PASS.
- PostgreSQL 16.15, Read Committed, disposable schemas: actual
  `claimsFromCCScheduler` allocation and later-provider failure rollback PASS;
  exact conditional-debit row effects PASS; linked `pg_blocking_pids` evidence
  before holder release and final nonnegative quota/provider isolation PASS.
- Removing only the sufficient-quota SQL predicate makes the actual SQL
  regression fail. Mutation discarded; full SQL baseline restored PASS.
- Yugabyte execution: NOT RUN. The PostgreSQL linked-wait observer explicitly
  skips Yugabyte rather than asserting an unsupported observation.

SQL fixtures use repository schema definitions, dedicated explicit loopback
targets and owned cleanup. Chain APIs are test doubles; no native sealing or
chain-acceptance claim is made. This is not cluster-wide proportional fairness.

### DIFF
--- a/documentation/en/cc-scheduler-quota.md
+++ b/documentation/en/cc-scheduler-quota.md
@@ -1,0 +1,31 @@
+# SupraSeal CC quota allocation
+
+Each SupraSeal batch may fill unused slots from enabled CC schedules. The
+planner first floors each weighted share and caps it at the provider's
+remaining `to_seal` quota. It redistributes the remainder one sector per
+eligible provider in descending weight order, breaking weight ties by
+ascending provider ID. Exhausted providers receive nothing. A batch cannot
+use a partial plan if the combined quota is insufficient.
+
+This intentionally replaces the uncapped last-provider remainder. For
+weights `3, 2, 1` and ten requested sectors with sufficient quota, the new
+counts are `6, 3, 1`, rather than `5, 3, 2`. Exact proportional shares remain
+unchanged. The planner does not alter the supplied schedule or duration.
+
+Schedule discovery precedes the existing ascending provider-lock acquisition.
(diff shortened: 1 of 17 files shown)


hyunmoon:
Sorry. Codex went haywire.

PR #1521, opened 2026-09-09
web: add an optional unclaimed SDR filter to the PoRep view

hyunmoon:
## Summary

Add an explicit, initially disabled PoRep view filter for unfinished SDR sectors
without a current owner. Failed sectors, sectors past SDR, owned tasks and older
responses with unknown ownership remain visible. The default still shows every
returned row; the checkbox is local to this page and does not change scheduling.

The additive `SDROwned` field comes from the same sector query snapshot. It is
ownership telemetry, not proof of execution start. Existing API rows and fields
remain intact. The page reports shown/hidden/returned counts and calculates its
existing waiting counters over the full returned result.

## Validation

- Executable filter and production call-site tests: `node --test
  web/test/porep-visibility.test.mjs`, three tests PASS.
- An intentional filter bypass caused both membership assertions to fail;
  the restored committed source passed.
- Actual Chromium/Lit page with intercepted synthetic responses: default all
  five rows, enabled four rows, disabled five rows; source rows and truthful
  counts preserved. No live Curio API was used.
- API package compilation, matching-tag vet, golangci-lint 2.11.4 and canonical
  fiximports PASS with `cgo,fvm,nosupraseal`; diff check PASS. The API package
  has no Go tests, so compilation is not claimed as executed SQL coverage.

No database row-effect, query-cost or live deployment result is claimed.
This is independent of Cluster Tasks timing, pacing and other proposed fixes.

### DIFF
--- a/documentation/en/developer/porep-visibility.md
+++ b/documentation/en/developer/porep-visibility.md
@@ -4,6 +4,8 @@ The PoRep sector table initially shows every returned sector, preserving the
 existing default. The checkbox can hide sectors that are neither failed nor
 past SDR and have no current SDR owner. It affects this view only and is reset
 when the page is reconstructed; it does not change scheduling or shared API data.
+An embedding page can opt in initially with the `hide-pending-sdr` boolean
+attribute. The standard page does not set it; toggling remains reversible.
 
 The API still returns every sector. Page counts distinguish shown, hidden and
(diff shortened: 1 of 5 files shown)


hyunmoon:
Sorry. Codex went haywire.

PR #1523, opened 2026-09-09
Feat: Enable removal of 1-epoch datasets

### LINKED ISSUE #1353: Remove temporary FWSS activation-boundary settlement skip after contract upgrade
Curio currently has a temporary workaround in shouldSkipFWSSActivationTerminationBoundary to skip terminated FWSS PDP rails where endEpoch == activationEpoch.
This exists because the current FWSS contract cannot settle that boundary case: validatePayment allows toEpoch == activationEpoch, but the proving-period helper treats that epoch as invalid, leaving Curio with no valid settlement target.
Once the upgraded FWSS contracts are deployed and active, remove this workaround and its tests. Curio should then rely on the contract to handle the activation-boundary case correctly instead of skipping those rails client-side.

This issue should be resolved with a PR once https://github.com/FilOzone/filecoin-services/issues/544 is deployed

snadrus:
Fixes https://github.com/filecoin-project/curio/issues/1353

### DIFF
--- a/pdp/contract/FWSS/settle_resolver_test.go
+++ b/pdp/contract/FWSS/settle_resolver_test.go
@@ -154,13 +154,14 @@ func TestFWSSPDPSettleTarget(t *testing.T) {
 			wantResolvable: true,
 		},
 		{
-			name:           "terminated rail ending at activation is skipped for FWSS contract bug",
+			name:           "terminated rail ending at activation settles to activation boundary",
 			settledUpTo:    997,
 			endEpoch:       1000,
 			currentEpoch:   7000,
 			activation:     1000,
 			provingPeriod:  2880,
-			wantResolvable: false,
+			wantTarget:     1000,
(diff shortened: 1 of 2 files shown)


ZenGround0 (approved):

pdp/contract/FWSS/settle_resolver_test.go
```diff
@@ -154,13 +154,14 @@ func TestFWSSPDPSettleTarget(t *testing.T) {
 			wantResolvable: true,
 		},
 		{
-			name:           "terminated rail ending at activation is skipped for FWSS contract bug",
```
ZenGround0, line 157: Shouldn't we just be removing this testcase ?

snadrus: It now tests: settledUpTo < endEpoch == activation

PR #1524, opened 2026-09-10
Feat: Allow higher message count max for AddPieces

### LINKED ISSUE #1421: PDP: Smart batch size limits
Tracking the curio work behind this: https://github.com/FilOzone/filecoin-services/issues/575

Essentially we should be able to inspect the metadata requirements of an add pieces message and allow batch size to grow when metadata is empty.  There will be associated synapse integration work.

snadrus:
Fixes https://github.com/filecoin-project/curio/issues/1421
It won't matter to anyone until Synapse sends bigger sets that fit the window

### DIFF
--- a/itests/pdp_message_size_test.go
+++ b/itests/pdp_message_size_test.go
@@ -1,0 +1,18 @@
+package itests
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/filecoin-project/curio/pdp"
+
+	"github.com/filecoin-project/lotus/chain/messagepool"
+)
+
+// TestPDPFilecoinMessageSizeMatchesLotus pins pdp.MaxFilecoinMessageSize to Lotus.
+// pdp copies MaxMessageSize instead of importing messagepool, which pulls
+// filecoin-ffi and breaks the PDP no-CGO build.
+func TestPDPFilecoinMessageSizeMatchesLotus(t *testing.T) {
+	require.Equal(t, messagepool.MaxMessageSize, pdp.MaxFilecoinMessageSize)
+}
--- a/pdp/pull_types_test.go
+++ b/pdp/pull_types_test.go
@@ -212,13 +212,13 @@ func TestPullRequest_Validate(t *testing.T) {
 	}
 }
 
-func TestPullRequest_ValidateBatchLimit(t *testing.T) {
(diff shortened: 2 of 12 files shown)


itests/pdp_message_size_test.go
```diff
@@ -0,0 +1,18 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/filecoin-project/curio/pdp"
+
+	"github.com/filecoin-project/lotus/chain/messagepool"
+)
+
+// TestPDPFilecoinMessageSizeMatchesLotus pins pdp.MaxFilecoinMessageSize to Lotus.
+// pdp copies MaxMessageSize instead of importing messagepool, which pulls
+// filecoin-ffi and breaks the PDP no-CGO build.
```
ZenGround0, line 15: 👍

pdp/pull_types_test.go
```diff
@@ -212,13 +212,13 @@ func TestPullRequest_Validate(t *testing.T) {
...
 
-func TestPullRequest_ValidateBatchLimit(t *testing.T) {
+func TestPullRequest_ValidateAllowsLargeBatches(t *testing.T) {
 	const validCid = "bafkzcibf6x7poaqtr2pqm6qki6sgetps74xutpclzrwbux5ow6rw4nsfu6tbf2zfnmnq"
 	dataSetId := uint64(1)
 
-	// Distinct source URLs so the batch-size check is isolated from the
-	// duplicate and per-piece checks that run afterwards.
-	pieces := make([]PullPieceRequest, MaxAddPiecesBatchSize+1)
+	// Piece count is no longer capped; Filecoin message size is enforced later
```
ZenGround0, line 219: Why not just delete this test?  Making the limit 41 only makes sense in the context of the previous state.  That state is no longer there so right now this change makes for a bad test.

ZenGround0: If you really want to test large batches as the name change suggests why not try adding 2000 pieces?

ZenGround0 (approved):
FYI @Kubuxu

pdp/pull_types_test.go
snadrus: good idea

Kubuxu:
> It won't matter to anyone until Synapse sends bigger sets that fit the window

Synapse couldn't send larger until Curio supported it 😄

PR #1525, opened 2026-09-10
Publish Curio image along side curio-pdp image

LexLuthr:
This PR also enables arm64 on both images

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,6 +3,16 @@
 # - skiff: sparse-clones filecoin-ffi default branch (Go module stubs only, no rust/)
 
 extern/supraseal
+extern/filecoin-ffi/target
+extern/filecoin-ffi/rust/target
+extern/filecoin-ffi/.install-filcrypto
+extern/filecoin-ffi/filcrypto.h
+extern/filecoin-ffi/filcrypto.pc
+extern/filecoin-ffi/libfilcrypto.a
+build/.update-modules
+build/.filecoin-install
+build/.blst-install
+build/.supraseal-install
 
 docker/data
 docker/skiff/data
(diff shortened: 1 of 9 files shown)


Dockerfile
```diff
@@ -1,144 +1,83 @@
...
+      org.opencontainers.image.source="https://github.com/filecoin-project/curio" \
+      org.opencontainers.image.description="Curio Filecoin storage provider"
 
-#####################################
-FROM ubuntu:24.04 AS curio-all-in-one
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    libhwloc15 libnuma1 libssl3t64 libstdc++6 ocl-icd-libopencl1 \
+    && rm -rf /var/lib/apt/lists/* \
+    && useradd -r -u 532 -U fc \
+    && mkdir -p /etc/OpenCL/vendors /var/tmp/filecoin-proof-parameters /var/lib/curio \
```
snadrus, line 67: Do static stuff first as a layer and later runs complete faster

snadrus (approved):

.dockerignore
```diff
@@ -3,6 +3,16 @@
 # - skiff: sparse-clones filecoin-ffi default branch (Go module stubs only, no rust/)
 
 extern/supraseal
+extern/filecoin-ffi/target
```
ZenGround0, line 6: This probably saves quite a bit of time and space?

ZenGround0 (approved):
Not an in depth review here from me but nothing blocking

.dockerignore
LexLuthr: Yes. We are not building FFI even accidentally and of course saves 9GiB+ in the build context. This was discovered by Andy.

PR #1526, opened 2026-09-10
fix: IPNI GUI

LexLuthr:
Replaces https://github.com/filecoin-project/curio/pull/1478


ZenGround0 (approved):
I am not tracing through the logic closely but this looks reasonable and safe.  And I trust your judgement with IPNI UX.

snadrus (approved):

PR #1531, opened 2026-09-11
NV 29 compatibility

### LINKED ISSUE #1368: [Solstice] CLI changes for verifreg and datacap deprecation
This task will track the CLI changes required for verifreg and datacap deprecation, slated for nv29 in the Solstice FIP (draft: https://github.com/filecoin-project/FIPs/pull/1270)

Affected methods:
* `curio market ddo`
* `sptool toolbox mk12-client allocate`
* `sptool toolbox mk12-client list-allocations`
* `sptool toolbox mk12-client deal`
* `sptool toolbox mk12-client offline-deal` 
* `sptool toolbox mk20-client deal`
* `sptool sectors status`
* `sptool sectors extend`

As part of this, we want to provide a CLI method to help them batch-increase their QAP using the recommended method (https://github.com/filecoin-project/builtin-actors/issues/1763). According to the timing of the command with respect to the upgrade, deprecated CLI methods should fail and recommend new methods.

After the upgrade, we will want to eventually remove the deprecated methods (perhaps after 3 months). This removal should probably be a separate follow-on issue.

If the scope of this task is too large, please break it into subtasks.

### LINKED ISSUE #1544: Upgrade Sector Quality functionality
### Checklist

- [x] This is **not** a new feature or an enhancement to the Filecoin protocol. If it is,  please open an [FIP issue](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0001.md).
- [x] This is **not** brainstorming ideas. If you have an idea you'd like to discuss, please open a new discussion on [the Curio forum](https://github.com/filecoin-project/curio/discussions/categories/ideas) and select the category as `Ideas`.
- [x] I **have** a specific, actionable, and well motivated feature request to propose.

### Curio component

- [x] PoRep Pipeline
- [ ] Snap Deals
- [ ] Market
- [ ] IPNI provider
- [ ] Curio UI
- [ ] Other

### What is the motivation behind this feature request? Is your feature request related to a problem? Please describe.

nv29 is introducing a new sector extension method which allows for upgrading any sector to full 10x QAP.  A builtin miner actor method must be called for each CC / incompletely filled sector that the SP operator wishes to upgrade quality for.  An important aspect of this method is that it requires 10x the initial pledge requirement of that needed for a CC sector just as a fully verified deal sector used to command before nv29.

### Describe the solution you'd like

Either a configuration option, a UI interface or a cli command that triggers this for a user controlled set of sectors.  It can't be fully automatic because operators may not have or want to post the requisite IP requirement.

### Describe alternatives you've considered

_No response_

### Additional context

@LexLuthr has been working closely on design and review for same feature in lotus-miner so he will have good ideas on delivering this.

LexLuthr:


snadrus (approved):

cmd/curio/market.go
```diff
@@ -352,153 +345,3 @@ var marketMoveToEscrowCmd = &cli.Command{
 		return nil
 	},
 }
-
-var ddoCmd = &cli.Command{
-	Name:      "ddo",
```
ZenGround0, line 357: I would prefer we keep the capability to do ddo somewhere and just remove the verified deal pieces of it.

cmd/sptool/sector.go
```diff
@@ -637,7 +639,7 @@ var sectorsListCmd = &cli.Command{
 			tablewriter.Col("Events"),
 			tablewriter.Col("Deals"),
 			tablewriter.Col("DealWeight"),
```
ZenGround0, line 641: Now that deal weight is getting closer to fully deprecated and will be 0 even if a sector is full of user data it makes sense to leave this off entirely.

cmd/sptool/sector.go
```diff
@@ -649,12 +651,19 @@ var sectorsListCmd = &cli.Command{
...
 			_, inASet := activeIDs[s]
 
-			const verifiedPowerGainMul = 9
-			dw, vp := .0, .0
-			{
+			sectorSize, err := st.SealProof.SectorSize()
+			if err != nil {
+				return xerrors.Errorf("getting sector %d size: %w", s, err)
+			}
+			qaPower, err := curiochain.SectorQAPower(st)
```
ZenGround0, line 658: Probably just cleaner and more useful to print out QA power directly ?

cmd/sptool/sector.go
```diff
@@ -649,12 +651,19 @@ var sectorsListCmd = &cli.Command{
...
+			if err != nil {
+				return xerrors.Errorf("getting sector %d QA power: %w", s, err)
+			}
+			vp := float64(big.Sub(qaPower, big.NewIntUnsigned(uint64(sectorSize))).Uint64())
+			dw := .0
```
ZenGround0, line 663: I would prefer us to rethink using deal weight and verified deal weight enough that we don't need to compute this.  I suspect removing reporting of weight and moving over entirely to QAP reporting better suits SP user needs.

lib/curiochain/sector_power.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/filecoin-project/go-state-types/big"
+	miner19 "github.com/filecoin-project/go-state-types/builtin/v19/miner"
+
+	"github.com/filecoin-project/lotus/chain/actors/builtin/miner"
+)
+
+// TODO: Use the upstream constant when the Solstice dependencies are released.
+const FULL_QA_POWER miner.SectorOnChainInfoFlags = 1 << 1
+
+// SectorIsFullQaPower includes both flagged sectors and legacy sectors at 10x power.
```
ZenGround0, line 16: I'm not sure exactly where and how you are using this function but just from the usage in this file the handling of legacy sectors at full power is unnecessary.  Also you should probably leave TODO here that you are going to use miner29.QAPower from go state types instead of this code once that is released.

lib/curiochain/sector_power.go
```diff
@@ -0,0 +1,45 @@
...
+	"golang.org/x/xerrors"
+
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/go-state-types/big"
+	miner19 "github.com/filecoin-project/go-state-types/builtin/v19/miner"
+
+	"github.com/filecoin-project/lotus/chain/actors/builtin/miner"
+)
+
+// TODO: Use the upstream constant when the Solstice dependencies are released.
```
ZenGround0, line 13: It might make sense to just hold off on merging until that is released, we should really get go state types PR out this week.

market/mk12/mk12.go
```diff
@@ -392,31 +399,6 @@ func (m *MK12) validateDealProposal(ctx context.Context, deal *ProviderDealState
 		return &validationError{error: err}
 	}
 
-	// Verified deal checks
-	if proposal.VerifiedDeal {
```
ZenGround0, line 396: Lovely how many things get cleaned up

market/mk20/contract/CurioDealViewV2.sol
```diff
@@ -0,0 +1,36 @@
...
+
+    // Input argument Curio will send to verification function
+    struct CurioDealView {
+        uint256 dealId;
+        DealState state;
+        uint256 providerActorId;// Miner's actor ID
+        bytes clientId;         // Client address bytes
+        bytes pieceCidV2;       // Piece CID v2 bytes
+        uint256 startEpoch;     // Optional Start Epoch, must be 0 when not used
+        uint256 duration;       // Duration of deal in Filecoin epochs
```
ZenGround0, line 24: The only change to this struct appears to be the removal of allocationID.  Since that is always optional I'm not sure I see the need for a fully new version.  We could just mark as deprecated in comments and handle any nonzero value as error?  Then again idk the usage of this contract that well so not so sure about that.

market/mk20/ddo_v1.go
```diff
@@ -186,24 +186,47 @@ func (d *DDOV1) VerifyMarketDeal(ctx context.Context, db *harmonydb.DB, eth ethc
...
-	mdeal := mk20contract.ICurioDealViewV1CurioDealView{
+	mdeal := mk20contract.ICurioDealViewV2CurioDealView{
 		DealId:          new(big.Int).SetUint64(*d.MarketDealID),
 		State:           mk20contract.DealStatusOpen,
 		ProviderActorId: new(big.Int).SetUint64(localProviderID),
 		ClientId:        localClient,
 		PieceCidV2:      deal.Data.PieceCID.Bytes(),
 		StartEpoch:      startEpoch,
 		Duration:        new(big.Int).SetUint64(uint64(d.Duration)),
-		AllocationId:    alloc,
```
ZenGround0, line 197: Yeah it feels like we could just set this to 0 always now

tasks/expmgr/handlers_extend.go
```diff
@@ -81,14 +83,14 @@ func (e *ExpMgrTask) handleExtend(ctx context.Context, cfg extendPresetConfig) (
...
-	// Use pre-crawled claim data to filter sectors:
+	// Before NV29, use pre-crawled claim data to filter sectors:
 	// - If drop_claims=false: exclude sectors where min_claim_epoch < target_expiration
 	// - If drop_claims=true: include sectors, we'll check max_claim_epoch later
 	var sectors []SectorMeta
 
 	// Use conditional queries based on CC filter and drop_claims setting
 	// Exclude sectors in snap pipeline or with open pieces (not yet finalized)
-	if cfg.CC != nil && !cfg.DropClaims {
+	if cfg.CC != nil && useClaims && !cfg.DropClaims {
```
ZenGround0, line 93: Please file a cleanup issue for removing all verifreg handling after nv29 has finished upgrading.

tasks/expmgr/handlers_extend.go
```diff
@@ -107,7 +109,7 @@ func (e *ExpMgrTask) handleExtend(ctx context.Context, cfg extendPresetConfig) (
 			  AND NOT EXISTS (SELECT 1 FROM open_sector_pieces osp WHERE osp.sp_id = sm.sp_id AND osp.sector_number = sm.sector_num)
 			ORDER BY sm.deadline, sm.partition, sm.sector_num`,
 			cfg.SpID, int64(currEpoch), int64(maxCandidateEpoch), *cfg.CC, int64(targetExpEpoch))
-	} else if cfg.CC != nil && cfg.DropClaims {
+	} else if cfg.CC != nil {
```
ZenGround0, line 112: Im definitely confused as to why we have removed the cfg.DropClaims clause here.  Naively I would have expected cfg.DropClaims && useClaims

I don't care too much about this case as extending verified deals before nv29 is not something the nv upgrade release really needs to get right.

web/api/webrpcporep/sector.go
```diff
@@ -764,17 +763,30 @@ func (a *PoRep) SectorInfo(ctx context.Context, sp string, intid int64) (*Sector
 	}
 
 	if onChainInfo != nil {
-		dw, vp := .0, .0
+		qaPower, err := curiochain.SectorQAPower(onChainInfo)
+		if err != nil {
```
ZenGround0, line 767: Reiterating that we should probably move away from displaying deal weights at all in favor or QAP

tasks/storage-market/mk20.go
```diff
@@ -1032,22 +1037,24 @@ func (d *CurioStorageDealMarket) processMK20DealIngestion(ctx context.Context) {
...
-			log.Errorw("failed to lookup client id", "deal", deal.ID, "error", err)
-			continue
-		}
-
-		clientId, err := address.IDFromAddress(clientIdAddr)
-		if err != nil {
-			log.Errorw("failed to parse client id", "deal", deal.ID, "error", err)
-			continue
+		var client address.Address
+		var clientId uint64
```
ZenGround0, line 1041: Its no good to keep an inaccurate clientId hanging around for a large code block like this. We should move declaration of clientId directly into the if nv < network.Version29 block

ZenGround0 (approved):
CI also indicates some work needed on testing and gen check failures

cmd/curio/market.go
LexLuthr: This was always verified DDO. There was no unverified DDO in MK12.

web/api/webrpcporep/sector.go
LexLuthr: Yes. ~But, We must ship and pre/post compatible code and then remove the branches later to avoid breaking pre-upgrade.~

tasks/storage-market/mk20.go
LexLuthr: Fixed

web/api/sector/sector.go
```diff
@@ -217,7 +227,7 @@ func (c *cfg) getSectors(w http.ResponseWriter, r *http.Request) {
 						pi = append(pi, pieces[k])
 					}
 				}
-				estimate := st.Expiration-st.Activation <= 0 || s.HasSnap
+				estimate := st.Expiration-st.PowerBaseEpoch <= 0
```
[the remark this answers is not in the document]
LexLuthr: HasSnap comes from the update file and stays true after snap is complete. We should show the on-chain QAP until the update lands. Restoring this would keep using estimates for completed snaps too.

tasks/expmgr/handlers_extend.go
LexLuthr: We still need the CC filter post NV29. Adding cfg.DropClaims && useClaims here would skip that filter because useClaims is false. The first branch already handles claims pre NV29.

lib/curiochain/sector_power.go
LexLuthr: Yes, we are not merging till a Lotus rc is out.

market/mk20/contract/CurioDealViewV2.sol
LexLuthr: I basically wanted to not allow anyone to even accidentally use allocations for anything. But there are probably easier way to block it than this.

market/mk20/ddo_v1.go
LexLuthr: Done

lib/curiochain/sector_power.go
LexLuthr: Updated the func to use the newer dep and flag from it. The function still exists for easier long term maintenance.

cmd/sptool/sector.go
LexLuthr: Done

tasks/expmgr/handlers_extend.go
LexLuthr: https://github.com/filecoin-project/curio/issues/1547

lib/curiochain/sector_power.go
```diff
@@ -10,12 +10,9 @@ import (
 	"github.com/filecoin-project/lotus/chain/actors/builtin/miner"
 )
 
-// TODO: Use the upstream constant when the Solstice dependencies are released.
-const FULL_QA_POWER miner.SectorOnChainInfoFlags = 1 << 1
-
 // SectorIsFullQaPower includes both flagged sectors and legacy sectors at 10x power.
 func SectorIsFullQaPower(info *miner.SectorOnChainInfo) bool {
-	if info.Flags&FULL_QA_POWER != 0 {
+	if info.Flags&miner19.FULL_QA_POWER != 0 {
```
ZenGround0: I don't understand this flag shouldn't be here in miner19?

ZenGround0 (changes requested):

lib/curiochain/sector_power.go
LexLuthr: I am switching to Lotus's implementation.

cmd/sptool/sector.go
```diff
@@ -19,6 +19,7 @@ import (
 	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/go-state-types/big"
 	"github.com/filecoin-project/go-state-types/builtin"
+	stminer "github.com/filecoin-project/go-state-types/builtin/v19/miner"
```
ZenGround0, line 23: I understand now we had preemptively created v19 in go state types and imported everywhere

go.mod
```diff
@@ -33,15 +33,15 @@ require (
 	github.com/filecoin-project/go-jsonrpc v0.10.2
 	github.com/filecoin-project/go-keccak v0.1.0
 	github.com/filecoin-project/go-padreader v0.0.1
-	github.com/filecoin-project/go-state-types v0.19.0-rc1
+	github.com/filecoin-project/go-state-types v0.19.0-rc1.0.20260902071721-6cb27cf2e8be // dependency-check-ignore: unknown
```
ZenGround0, line 36: I guess this is a good workaround for unmerged go state types PRs so we can get access to the types we need.  I was losing my mind trying to understand why this compiles with the FULL_QA_POWER flag

cmd/sptool/sector.go
ZenGround0: Thanks @rjan90 for being a few months ahead

cmd/sptool/sector.go
```diff
@@ -1300,3 +1306,195 @@ func siStr(bi types.BigInt) string {
...
+		if err != nil {
+			return err
+		}
+		addrSectors := sectorsMax
+		if n := cctx.Int("max-sectors"); n != 0 {
+			if n < 0 {
+				return xerrors.Errorf("--max-sectors must be positive, got %d", n)
+			}
+			if n > sectorsMax {
+				return xerrors.Errorf("--max-sectors %d exceeds the protocol limit of %d", n, sectorsMax)
```
ZenGround0: Don't error at most log a warning and then leave addrSectors set to sectorsMax.  If the cli param is called "max-sectors" then we can sensibly address fewer.

cmd/sptool/sector.go
```diff
@@ -1300,3 +1306,195 @@ func siStr(bi types.BigInt) string {
...
+		},
+		&cli.StringFlag{
+			Name:  "max-fee",
+			Usage: "maximum FIL to spend on gas per message",
+			Value: "0",
+		},
+		&cli.BoolFlag{
+			Name:  "really-do-it",
+			Usage: "must be specified for the action to take effect",
+		},
```
ZenGround0, line 1329: I think it makes a lot more sense to limit the total number of sectors that get upgraded here with the max-sectors field.  As an SP why do I care how many sectors are packed per message?  I would want this command to figure that out for me.  But since each upgrade involves a business decision of how much capital to lock up I really do care about total sectors addressed.  Right now there is nothing you can pass in to prevent the entire miner actor from getting upgraded and to me that looks like an oversight.

ZenGround0: A dry run flag that calculated how many sectors and how much power increased and pledge lockup will be increased would also be really useful.

ZenGround0 (changes requested):
For the most part I like it, I just have some useability concerns with sector quality upgrade

cmd/sptool/sector.go
LexLuthr: Historically, "really-do-it" flag has been used to differentiate dry run and actual execution in filecoin CLIs. So flagged is named "really-do-it" for the same reason.

I am not sure I get your concern here. Max sector is the deciding factor. There is no flag to specify the number of messages. As for printing the number of messages, it is more for debugging and auditing along with message CIDs themselves.

LexLuthr: I added the more info as requested in second part. I chose to always print the details it is useful to know what was actually paid (estimate).

ZenGround0: I agree on really do it but I want to talk more about max sectors

cmd/sptool/sector.go
```diff
@@ -1381,22 +1381,17 @@ var sectorsUpgradeQualityCmd = &cli.Command{
...
-			}
-			if n > sectorsMax {
-				n = sectorsMax
-			}
-			addrSectors = n
+		if maxSectors > 0 {
+			limit = true
 		}
 
+		batchSize := 12500
```
ZenGround0, line 1393: I don't see a reason to stop reading policy.GetAddressedSectorsMax here

cmd/sptool/sector.go
```diff
@@ -1421,6 +1416,10 @@ var sectorsUpgradeQualityCmd = &cli.Command{
 
 				var upgrade *stminer.UpgradeSectorQuality
 				return active.ForEach(func(sn uint64) error {
+					if total >= maxSectors && limit {
```
ZenGround0: This is going to keep iterating all partitions and deadlines if we hit the limit.  You should add a check in the outer loop functions too to end early.

ZenGround0 (approved):
Thanks, gen-check needs some help too but ready to merge otherwise

cmd/sptool/sector.go
LexLuthr: Actor has no limit, batch size is hardcoded so policy adds nothing to this code.

BigLep:
@LexLuthr : anything holding up merging this?

PR #1550, opened 2026-09-21
fix(pdp-guide): send storage capacity to FSRegister

SgtPooki:
### What changed

The PDP Guide's "Register with FOC" form called `FSRegister` with 3 arguments, but the RPC takes 4 since `capacityTiB int64` was added (`web/api/webrpc/pdp.go:500`), so every submit failed with `wrong param count (method 'CurioWeb.FSRegister'): 3 != 4`. Added a "Storage capacity (TiB)" field to the guide form and pass it through, matching what `pages/pdp/register.mjs` already sends. The input is `type="number" min="1" step="1" required`, so the browser enforces the positive whole number the backend expects.

### How to verify

Bring up a skiff node (calibnet is fine), complete wallet, storage and DNS, then submit the guide's register form. Before: the red param-count error. After: "Registration submitted" and the provider appears in the registry.

No test added: `web/static` has no JS test harness, and the change is a form field plus one argument.


web/static/pages/pdp-guide/guide.mjs
```diff
@@ -592,6 +595,9 @@ customElements.define('pdp-guide', class PDPGuideElement extends LitElement {
                                    @input=${(e) => { this.registerLocation = e.target.value; }}
                                    placeholder="C=US;ST=California;L=San Francisco" required>
                             <p class="hint">Location format: <span class="mono">C=US;ST=California;L=San Francisco</span></p>
+                            <label for="reg-cap">Storage capacity (TiB)</label>
+                            <input id="reg-cap" type="number" min="1" step="1" .value=${this.registerCapacityTiB}
+                                   @input=${(e) => { this.registerCapacityTiB = e.target.value; }} required>
```
SgtPooki, line 600: the other form to submit is also requiring 1TiB, but we should be able to onboard with < 100GiB now right? 

Image

Image

SgtPooki (commented):
self review

snadrus (approved):

web/static/pages/pdp-guide/guide.mjs
snadrus: Yes, please have the step be 0.1

snadrus: We should also change the heading shown to "Have 100 GiB of attached storage" and require 100 in subtext

ZenGround0 (approved):

ZenGround0:
@SgtPooki needs lint

web/static/pages/pdp-guide/guide.mjs
SgtPooki: @snadrus addressed your comments as discussed offline.. now minimum required is 100GiB, and we are registering SPs with `capacityGiB` instead of `capacityTiB` from now on, parsed properly as `capacityGiB=capacityTiB * 1024` on reads. older SPs who had `capacityTiB` stored in their registry will update to `capacityGiB` if they update their PDP offering.

requested re-review

